PR #21, opened 2021-11-25
Mitigate go 1.13/Ubuntu 20.04 "path@version" syntax error

mikini:
From Go 1.16 install subcommand accepts version suffixes and should be used for
executable installation according to [go get deprecation notice](https://go.dev/doc/go-get-install-deprecation)
and [release notes](https://go.dev/doc/go1.16).

Go <1.16 will error out when parsing a version suffix so this adds a check for
Go version and switches the argument and module mode between get and install
accordingly when installing toitpkg.

Shells out with "go version" piped through grep and two additional shell
invocations of bc for integer comparision.

Assumes to have available;
* grep with --perl-regexp for doing look-around (GNU only?)
* echo
* bc

Tested on Ubuntu 20.04 with go 1.13.8.

See also [writeup of module usage across Go versions](https://maelvls.dev/go111module-everywhere/).


kasperl:
Summoning the CLA bot: @cla-bot check.

kasperl:
Re-checking CLA status: @cla-bot check.

Issue #51, opened 2021-12-02
Missing make flash target


lask:
When I build an image for ESP32 with `make esp32`, I get the following output:
```
AS build/esp32/xtensa/expression_with_stack_xtensa_asm.o
AS build/esp32/xtensa/xtensa_intr_asm.o
AR build/esp32/xtensa/libxtensa.a
make[2]: Leaving directory '/home/lau/toitlang/toit/build/esp32/xtensa'
Generating esp32.project.ld
LD build/esp32/toit.elf
esptool.py v3.1-dev
Merged 2 ELF sections
To flash all build output, run 'make flash' or:
python /home/lau/toitware/esp-idf/components/esptool_py/esptool/esptool.py --chip esp32 --port /dev/ttyUSB0 --baud 921600 --before default_reset --after hard_reset write_flash -z --flash_mode dio --flash_freq 40m --flash_size detect 0xd000 /home/lau/toitlang/toit/build/esp32/ota_data_initial.bin 0x1000 /home/lau/toitlang/toit/build/esp32/bootloader/bootloader.bin 0x10000 /home/lau/toitlang/toit/build/esp32/toit.bin 0x8000 /home/lau/toitlang/toit/build/esp32/partitions.bin
make[1]: Leaving directory '/home/lau/toitlang/toit/toolchains/esp32'
```
However, there is no `make flash` target.
```
$ make flash
make: *** No rule to make target 'flash'.  Stop.
```

PR #65, opened 2021-12-05
Add make flash target

justind000:
PR for issue #51. Adds `make flash` target. 

`make flash` can be called with ESP32_ENTRY, ESP32_WIFI_PASSWORD, ESP32_WIFI_SSID as with `make esp32` and additionally adds ESP32_PORT, which is required, to specify the port the ESP32 is connected to. 

To compile new changes and flash to device:
`make flash ESP32_PORT=/dev/ttyS17 ESP32_WIFI_PASSWORD=password ESP32_WIFI_SSID=ssid ESP32_ENTRY=examples/hello.toit`

Assumes `python` is installed, IDF_PATH is valid, particular ESP32 can communicate at 921600 baud, and user has permission to access the device (sudo adduser $USER dialout on most systems).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ BUILD_DATE = $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
 ESP32_ENTRY=examples/hello.toit
 ESP32_WIFI_PASSWORD=
 ESP32_WIFI_SSID=
+ESP32_PORT=
 
 .PHONY: all
 all: tools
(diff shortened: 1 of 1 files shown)


kasperl:
Hi @justind000. Thank you so much for working on this! Our guidelines for contributing are [here](https://github.com/toitlang/toit/blob/master/CONTRIBUTING.md) and they explain the CLA steps.

Makefile
```diff
@@ -30,6 +31,11 @@ tools: check-env toitpkg toitlsp build/host/bin/toitvm build/host/bin/toitc
 .PHONY: esp32
 esp32: check-env build/esp32/toit.bin
 
+.PHONY: flash
+flash: esp32
```
kasperl, line 35: Maybe depend on check-env too and move any printing of IDF_PATH there?

kasperl: Maybe it would be cleaner to depend on build/esp32/toit.bin? We could expand that to all the .bin files and make it clear that they are produced by the same rule.

kasperl (approved):
Looks good to me.

kasperl:
Rechecking CLA status: @cla-bot check.

kasperl:
Let me know how you feel about the review comments, @justind000. I am ready to land as is and then we can always make this more involved in the future.

justind000:
Either way is perfectly fine with me. I would guess the next Makefile change isn't far away regardless.

kasperl:
Landed :)

PR #68, opened 2021-12-05
Fix writing arrays into images.

kasperl:


tools/image.toit
```diff
@@ -1094,7 +1094,7 @@ class ToitInstance extends ToitHeapObject:
 
     o_.fields.size.repeat:
       field := o_.fields[it]
-      field_address := (ToitHeapObject field).write_to_ image
+      field_address := (ToitHeapObject field).write_to image
```
kasperl, line 1097: @floitsch: Not entirely sure about this change, but it feels a bit odd to call the internal method here.

erikcorry:
lgtm

floitsch:
LGTM!

Issue #43, opened 2021-12-01
Add RISC-V 64-bit platform support (SiFive Unmatched + QEMU)


dsobotta:
I've made some good progress on the SiFive Unmatched.  For anyone else looking to contribute, feel free to use [my fork](https://github.com/dsobotta/toit-riscv) as a starting point.  I'll issue a PR back into mainline once things have stabilized a bit.

* [x] build environment
* [x] IDF environment
* [X] IDF compile
* [ ] IDF export 
* [x] Toit generate build files
* [x] Toit compile sources
* [x] Toit generate snapshot
* [x] Toit run examples
* [x] riscv64 seccomp code review
* [x] Add makefile support to cross-compile binaries for riscv64

dsobotta:
Currently blocked by [is_directory: Invalid argument](https://github.com/dsobotta/toit-riscv/issues/3) .  This is reproducible in a QEMU/riscv64 environment following my [build steps](https://github.com/dsobotta/toit-riscv).  Is this something you could take a look at @kasperl ?

kasperl:
Will take a look, @dsobotta!

dsobotta:
Made some good progress on cross-compiling.

I added a few entries to 'Makefile' (and added toolchains/riscv64.cmake) to get cross-compiling of Toit binaries from amd64 host -> riscv64 working.  It successfully generates build/riscv64/bin/{lib/, mbedtls, toitc, toitvm}.  These binaries have been validated on RISC-V hardware, and appear to run the examples with the same results as toitvm/toitc binaries generated directly from a RISC-V host.

There's one snag though--the toitvm_boot.snapshot is not successfully generated as a part of the cross-compile build, as the build scripts attempt to run the riscv64 ELF when compiling it.  Is there a cmake var I can override in toolchains/riscv64.cmake , or will I need to refactor some portion of the build system to break out that step?

Here's the [github diff ](https://github.com/dsobotta/toit-riscv/commit/ff6353ece167cfe617a119775d7a65a16eb45e7a?diff=unified)with the current Makefile changes.

dsobotta:
With the [makefile refactor](https://github.com/toitlang/toit/pull/197) merged, we now have end-to-end support for riscv64/linux targets and cross-compiling from other hosts.  I don't suspect that many will be deploying to ESP32 devices from riscv64 hosts, so closing this out.

PR #74, opened 2021-12-06
Add 64-bit RISC-V/Linux platform support

dsobotta:
Note, this does not include support for embedded RISC-V devices; a full Linux environment is expected.  I will be evaluating support for the [ESP32-C3](https://www.espressif.com/sites/default/files/documentation/esp32-c3-wroom-02_datasheet_en.pdf) in the next few weeks.

See [status tracker](https://github.com/dsobotta/toit-riscv/blob/master/README_RISCV.md) for capabilities.
See [feature request](https://github.com/toitlang/toit/issues/43) for development history


Makefile
```diff
@@ -104,6 +114,9 @@ build/esp32/: check-env
 	mkdir -p $@
 	make -C toolchains/esp32 -s $(shell pwd)/build/esp32/include/sdkconfig.h
 
+build/riscv64/:
```
kasperl, line 117: Merge this with the build/host/ line?

```
build/host/ build/riscv64/:
        mkdir -p $@
```

src/compiler/filesystem_local.cc
```diff
@@ -80,7 +84,7 @@ bool FilesystemLocal::do_is_regular_file(const char* path) {
 }
 
 bool FilesystemLocal::do_is_directory(const char* path) {
-#ifndef TOIT_LINUX
+#ifndef USE_XSTAT64
```
kasperl, line 87: I would probably flip this around and use the positive (here and above):

```
#ifdef USE_XSTAT64
   xstat64
#else 
   ... stat
```

src/compiler/filesystem_local.cc
```diff
@@ -36,7 +36,11 @@ extern int __xstat64(int ver, const char* path, struct stat64* stat_buf);
 
 }
 
-#ifdef BUILD_64
+#if !defined(__riscv) && !defined(TOIT_LINUX)
+  #define USE_XSTAT64 1
+#endif
+
+#ifndef BUILD_64
```
kasperl, line 43: Was this a bug before?

src/sandbox.cc
```diff
@@ -72,14 +84,28 @@ static int COMPILER_SYSCALLS[] = {
...
   SYS_shutdown,
+#ifdef __riscv
+  SYS_ppoll,
+#else
   SYS_poll,
+#endif
   -1
 };
 
 static int MOST_SYSCALLS[] = {
```
kasperl, line 95: This is getting a little bit hard to review. I wonder if it would be better to go for separate lists (rather than all these ifdefs) long term. For now, this is good enough for me.

kasperl (approved):

Makefile
dsobotta: This one I think we should keep separate for all cross-compile builds (also ARM/Linux, which I'm working on at the moment), since we'll soon need to use the host's toitc to generate the vm boot snapshot.

src/compiler/filesystem_local.cc
dsobotta: Sure, I can update that quick.

src/compiler/filesystem_local.cc
dsobotta: Both xstat versions 1 and 3 returned error code -1 on riscv64, but stat works fine. 

riscv64/Linux builds are still Linux, so it seemed like a good idea to use a more descriptive define check.

kasperl: I was thinking more about the change from 

```
ifdef BUILD_64
```

to

```
ifndef BUILD_64
```

src/sandbox.cc
dsobotta: The syscalls should eventually converge as RISC-V/Linux support matures; but in the meantime I agree that this is very ugly.

If you prefer, I can consolidate the #ifndef __riscv 's together, which would clean up a lot of it.  I wasn't sure if that was ideal, or keeping the syscalls in the original order.

kasperl: Let's keep it as it is now and revisit later!

Makefile
kasperl: Sounds good to me.

src/sandbox.cc
dsobotta: Sounds good.

kasperl (approved):
Looks good. Landing!

PR #82, opened 2021-12-07
Add Toitware's ESP-IDF fork as submodule

kasperl:
Based on @mikini's work.


README.md
```diff
@@ -70,28 +70,29 @@ For builds targeting RISC-V, ARM32, or ARM64 hardware, see the [Other platforms
...
-popd
+
 ```
 
-Remember to add it to your ENV as `IDF_PATH`:
+For the build to succeed, you will need to add its path to your ENV as `IDF_PATH`:
 
 ``` sh
-export IDF_PATH=...
+export IDF_PATH=`pwd`/third_party/esp-idf
```
floitsch, line 91: I would make this the default, if IDF_PATH isn't set, but we can do that in a follow-up PR.

floitsch (approved):
LGTM.

README.md
floitsch: The newer posix way seems to be to use `$()` instead of backticks.
```suggestion
export IDF_PATH=$(pwd)/third_party/esp-idf
```

PR #102, opened 2021-12-09
Pack sdk for darwin and linux

jstroem:
Creates *.tgz for darwin and linux and uploads on release

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -2,9 +2,11 @@ name: Build macOS
 
 on:
   push:
-    branches: [master]
+    branches: [master,pack-artifacts]
   pull_request:
     branches: [master]
+  release:
+    types: [published]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


floitsch (approved):
LGTM.

.github/workflows/build_macos.yml
```diff
@@ -22,9 +24,19 @@ jobs:
...
+      - name: Upload release artifacts
+        if: github.event_name == 'release'
+        uses: svenstaro/upload-release-action@v2
         with:
-          name: macos-build
-          path: |
-            build/host/bin/toit*
-            build/host/bin/lib/
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: build/darwin.tgz
```
kasperl, line 40: Should this be macos.tgz?

jstroem: I choose Darwin because that is how golang does but I'm not sure if that's the right way

PR #107, opened 2021-12-09
Use config.json for wifi credentials

andersjohnsen:


Makefile
```diff
@@ -129,7 +130,8 @@ $(SNAPSHOT_DIR)/system_message.snapshot: tools/system_message.toit $(TOITC_BIN)
 	$(TOITC_BIN) -w $@ $<
 
 build/snapshot: $(TOITC_BIN) $(ESP32_ENTRY)
-	$(TOITC_BIN) -w $@ $(ESP32_ENTRY) -Dwifi.ssid=$(ESP32_WIFI_SSID) -Dwifi.password=$(ESP32_WIFI_PASSWORD)
+	echo '{"wifi": {"ssid": "$(ESP32_WIFI_SSID)", "password": "$(ESP32_WIFI_PASSWORD)"}}' > build/config.json
```
jstroem, line 133: we should make a seperate .PHONY `build/config.json` target that does this

lib/wifi.toit
```diff
@@ -0,0 +1,76 @@
...
+  wifi := wifi.Wifi
+  try:
+    wifi.set_ssid ssid password
+    with_timeout WIFI_CONNECT_TIMEOUT_: wifi.connect
+    with_timeout WIFI_DHCP_TIMEOUT_: wifi.get_ip
+    return Wifi wifi
+  finally: | is_exception _ |
+    if is_exception: wifi.close
+
+class Wifi extends net.Interface:
```
kasperl, line 34: Having looked a bit more at this, I think the right thing is to make `connect` return a `net.Interface` and to make the `wifi.Wifi` class private (call it `WifiInterface_`).

Also, I think this probably still belongs in `net/`, so you would import it with `import net.wifi` and use it with `wifi.connect`.

jstroem (approved):

PR #108, opened 2021-12-09
add inject_config.snapshot to the SDK

jstroem:


floitsch (approved):
LGTM.

If all snapshots come from `tools` we should at some point consider a generic Make rule.
But all of these should be created in cmake anyway (to track their dependencies).

Issue #89, opened 2021-12-07
Add ESP32-C3 (32-bit RISC-V) platform support


dsobotta:
Should be possible with some tweaks and extensions to the Toit build pipeline.

dsobotta:
What would it take to utilize cmake or idf.py instead of make?

libtoit is compiling with the changes found here: https://github.com/dsobotta/toit


After compiling libtoit_image.a , we get the following errors:

[dusten@dls-shuttle toit]$ make esp32 ESP32_CHIP=esp32c3
mkdir -p build/esp32c3/
make -C toolchains/esp32c3 -s /home/dusten/git/toit/build/esp32c3/include/sdkconfig.h
Toolchain path: /home/dusten/.espressif/tools/riscv32-esp-elf/esp-2021r1-8.4.0/riscv32-esp-elf/bin/riscv32-esp-elf-gcc
Toolchain version: esp-2021r1
Compiler version: 8.4.0
Python requirements from /home/dusten/git/toit/third_party/esp-idf/requirements.txt are satisfied.
/home/dusten/git/toit/third_party/esp-idf/components/esp_timer/component.mk:21: *** esp_timer is only supported by the Make build system for esp32 chip. For other chips, use the Cmake build system.  Stop.
(cd build/host && cmake ../.. -G Ninja -DCMAKE_BUILD_TYPE=Release)
CMake Deprecation Warning at third_party/esp-idf/components/mbedtls/mbedtls/CMakeLists.txt:1 (cmake_minimum_required):
  The OLD behavior for policy CMP0076 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.


-- Configuring done
-- Generating done
-- Build files have been written to: /home/dusten/git/toit/build/host
(cd build/host && ninja build_toitvm)
ninja: no work to do.
build/host/bin/toitc -w build/snapshot examples/hello.toit -Dwifi.ssid= -Dwifi.password=
(cd build/host && ninja build_toitvm)
ninja: no work to do.
build/host/bin/toitvm tools/snapshot_to_image.toit build/snapshot build/esp32c3/esp32c3.image.s
[...]

PR #109, opened 2021-12-09
support compiling libtoit_image.a for ESP32-C3

dsobotta:
Still need to work on the build system as a whole, but this provides preliminary support for ESP32-C3 throughout Toit source.

While the Toit static libraries can be built with these changes, it is not possible to compile the full ESP32-C3 stack, as even the hello_world example provided by ESP-IDF does not compile.

Manually copying libtoit_image.a and libtoit_vm.a to the toolchains/esp32c3/main directory will allow full compiling and flashing of an ESP32-C3 when using ESP-IDF 4.4.  However this will crash currently, as that method compiles against 2 different versions of ESP-IDF.

I think the next step after merging this will be to upgrade Toit to ESP-IDF 4.4

#89


kasperl:
Thank you for this, @dsobotta! Will take a look on Monday.

Maybe it is time to start thinking about which defines we rely on and unify them even further? I think I'd prefer having TOIT_ defines for everything we depend on in the code and set them up in one place. YMMV.

dsobotta:
> Thank you for this, @dsobotta! Will take a look on Monday.
> 
> Maybe it is time to start thinking about which defines we rely on and unify them even further? I think I'd prefer having TOIT_ defines for everything we depend on in the code and set them up in one place. YMMV.

I agree; it would be a good step in reducing dependence on ESP-IDF too.  Some abstraction around logging would also be nice.

erikcorry:
I meant to get to this today, but didn't, sorry.   Tomorrow!

toolchains/esp32c3/sdkconfig
```diff
@@ -0,0 +1,1403 @@
...
+#
+# ESP System Settings
+#
+# CONFIG_ESP_SYSTEM_PANIC_PRINT_HALT is not set
+CONFIG_ESP_SYSTEM_PANIC_PRINT_REBOOT=y
+# CONFIG_ESP_SYSTEM_PANIC_SILENT_REBOOT is not set
+# CONFIG_ESP_SYSTEM_PANIC_GDBSTUB is not set
+CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE=y
+CONFIG_ESP_SYSTEM_RTC_FAST_MEM_AS_HEAP_DEPCHECK=y
+CONFIG_ESP_SYSTEM_ALLOW_RTC_FAST_MEM_AS_HEAP=y
```
erikcorry, line 560: Interesting choice.  The cmpctmalloc hasn't been tested on RTC memory.  I don't know why it wouldn't work though.

README_OTHERPLATFORMS.md
```diff
@@ -27,22 +27,19 @@ apt install git build-essential bc cmake python3 python3-pip python-is-python3 l
...
+#ESP-IDF
+git submodule update --init --recursive
 ```
 
 ## 4) Compile Toit
 ``` sh
-cd toit
+#Add IDF path to environment
+export IDF_PATH=`pwd`/third_party/esp-idf
+
```
erikcorry, line 42: It would be best to make the links below relative, not pointing at the dsobotta repo, so that they point to the repo they are in.  (Github won't let me comment on lines that are not changed, but it's about 20 lines down from here.)

erikcorry (approved):

erikcorry:
Thanks for this!

I think there will be more work to do before this actually works, but it's great to see support for new hardware appearing.

dsobotta:
Thanks for taking the time to review a massive PR!  It will definitely need some more work; but at least this gets to the point of flashing with an update to ESP-IDF 4.4

PR #120, opened 2021-12-13
Add '.exe' on windows in make-targets

jstroem:


Makefile
```diff
@@ -25,11 +25,18 @@ ESP32_WIFI_SSID=
 ESP32_WIFI_PASSWORD=
 ESP32_PORT=
 
+
+ifeq ($(OS),Windows_NT)
```
kasperl, line 29: Is this always the right test?

Makefile
```diff
@@ -25,11 +25,18 @@ ESP32_WIFI_SSID=
 ESP32_WIFI_PASSWORD=
 ESP32_PORT=
 
+
```
kasperl, line 28: Drop one newline.

kasperl (approved):
Looks good to me.

Makefile
jstroem: The internet says so.

PR #123, opened 2021-12-13
Renamed platform names and added constants in toit

jstroem:
This also fixes `filesystem_local.cc` such that we use `USERPROFILE`  env variable instead of  `HOME` on windows.


kasperl (approved):
Looks good to me.

PR #151, opened 2021-12-15
Simplify building and better error message.

floitsch:


Makefile
```diff
@@ -25,6 +25,9 @@ ESP32_WIFI_SSID=
 ESP32_WIFI_PASSWORD=
 ESP32_PORT=
 
+# Use Toitware fork by default.
```
kasperl, line 28: Should we call this "Toitware's ESP-IDF fork" consistently?

Makefile
```diff
@@ -204,8 +192,15 @@ clean:
...
-	$(error IDF_PATH is not set, if you want to use the Toitware fork execute "export IDF_PATH=`pwd`/third_party/esp-idf" (see README.md))
+ifeq ("$(wildcard $(IDF_PATH)/components/mbedtls/mbedtls/LICENSE)","")
+ifeq ("$(IDF_PATH)", "$(CURDIR)/third_party/esp-idf")
+	$(error mbedtls sources are missing. Did you initialize the submodules?)
+else
+	$(error Invalid IDF_PATH. Missing mbedtls sources.)
+endif
+endif
+ifneq ("$(IDF_PATH)", "$(CURDIR)/third_party/esp-idf")
+	$(info -- Not using Toitware ESP-IDF)
```
kasperl, line 203: Should we call this "Toitware's ESP-IDF fork" consistently?

Makefile
```diff
@@ -183,11 +175,7 @@ toitpkg: $(TOITPKG_BIN)
 
 TOITPKG_VERSION := v0.0.0-20211126161923-c00da039da00
 $(TOITPKG_BIN):
-ifeq ($(GO_USE_INSTALL), 1)
 	GOBIN="$(CURDIR)"/$(dir $@) go install github.com/toitlang/tpkg/cmd/toitpkg@$(TOITPKG_VERSION)
```
kasperl, line 178: Should we check the go version?

floitsch: That's what the old code was trying to do, and it failed on Windows.
If the version isn't new enough there is an error, which can easily be googled. I can try to find a better way to check the version, but it would take time.
It is true, that Go 1.16 is relatively new, though (Feb 2021).

Makefile
floitsch: done.

Makefile
floitsch: done.

PR #154, opened 2021-12-15
Remove ar, directory, file and pipe from tools.

floitsch:
These files are now available in packages/host.


kasperl (approved):
Looks good to me.

PR #159, opened 2021-12-16
Remove legacy HTTP and websocket

andersjohnsen:
Should we check in .packages? Would probably be much easier for the users.


tests/http_test.toit
```diff
@@ -1,249 +0,0 @@
-// Copyright (C) 2018 Toitware ApS. All rights reserved.
```
kasperl, line 1: Do we have a test like this for the new http package?

andersjohnsen: There are examples, yes

kasperl: I worry that we get more and more untested functionality in packages and we will have no way of knowing if we're breaking things. Can we add tests to the packages (in this case the http one) and set it up so we run our package tests through GitHub actions?

andersjohnsen: Sure, but sounds a bit out of scope for this change.

kasperl: I can see that - and I am all for getting this cleaned up for users of the Toit SDK. Will you file an issue for the postponed testing, so we don't forget?

examples/README.md
```diff
@@ -2,6 +2,15 @@
 
 The Toit examples are meant to get you started writing code for an ESP32.
 
+## Packages
+
+Some of the examples use other packages, e.g. HTTP. To run the examples you first need to run
```
kasperl, line 7: Nit: Add ':' after run for consistency with other READMEs.

examples/README.md
```diff
@@ -2,6 +2,15 @@
...
+
+```
```
kasperl, line 9: Drop $ from the instructions and change it to:

``` sh
build/host/sdk/bin/toitpkg pkg install
```

(adding sh and using the built toitpkg)

kasperl (approved):

PR #161, opened 2021-12-16
Split udp test.

floitsch:


tests/udp_local_address_test.toit
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2021 Toitware ApS. All rights reserved.
```
andersjohnsen, line 1: Just call this `interface_address_test.toit`

andersjohnsen (approved):

tests/udp_local_address_test.toit
floitsch: done.

PR #165, opened 2021-12-17
Better nano-zlib implementation.

erikcorry:
It stil only does run-length encoding
for speed and memory savings, but now
it is capable of recognizing repeating
patterns of 2, 3, or 4 bytes as seen in
color image files.  Big improvement in
very simple color PNG files.


src/nano_zlib.cc
```diff
@@ -45,18 +45,128 @@ word ZlibRle::add(const uint8* contents, intptr_t extra) {
...
+        last_bytes_valid_ = 4;
+        unemitted_bytes_ = 0;
+        unemitted_bytes_valid_ = 0;
+      } else if (last_bytes_valid_ >= 2 &&
+          (last_bytes_ & 0xffff) == (unemitted_bytes_ & 0xffff) &&
+          (unemitted_bytes_ >> 16) == (unemitted_bytes_ & 0xffff)) {
+        // We recognized that the four unemitted bytes are two repetitions of
+        // the last two emitted ones.
+        mode_ = REP2;
+        repetitions_ = 4;
```
floitsch, line 74: Feels a bit wrong to call it "repetitions = 4" if it's only two repetitions.

src/nano_zlib.cc
```diff
@@ -45,18 +45,128 @@ word ZlibRle::add(const uint8* contents, intptr_t extra) {
...
+        last_bytes_valid_ = 4;
+        unemitted_bytes_ &= 0xff;
+        unemitted_bytes_valid_ = 1;
+        output_repetitions();  // Will flush completely since repetitions_ == 3.
+      } else if (last_bytes_valid_ >= 4 &&
+          last_bytes_ == unemitted_bytes_) {
+        // We recognized that the four unemitted bytes are the same as
+        // the last four emitted ones.  Start a run with offset 4.
+        mode_ = REP4;
+        repetitions_ = 4;
```
floitsch, line 99: Same: it's just one repetition of 4 bytes. Not 4 of them.

src/nano_zlib.cc
```diff
@@ -45,18 +45,128 @@ word ZlibRle::add(const uint8* contents, intptr_t extra) {
...
+        unemitted_bytes_ = 0;
+        unemitted_bytes_valid_ = 0;
+      } else if (last_bytes_valid_ >= 3 &&
+          (last_bytes_ & 0xffffff) == (unemitted_bytes_ >> 8)) {
+        // We recognized that the first four of the unemitted bytes are
```
floitsch, line 104:
```suggestion
        // We recognized that the first three of the unemitted bytes are
```

src/nano_zlib.cc
```diff
@@ -45,18 +45,128 @@ word ZlibRle::add(const uint8* contents, intptr_t extra) {
...
+        // the same as the last three emitted ones.
+        mode_ = REP3;
+        if ((unemitted_bytes_ & 0xff) == ((last_bytes_ >> 16) & 0xff)) {
```
floitsch, line 107: Why the shift of 16?

src/nano_zlib.cc
```diff
@@ -45,18 +45,128 @@ word ZlibRle::add(const uint8* contents, intptr_t extra) {
...
+          // The last unemitted byte also matches the pattern of three
```
floitsch, line 108: Not clear which pattern we are talking about.

src/nano_zlib.cc
```diff
@@ -45,18 +45,128 @@ word ZlibRle::add(const uint8* contents, intptr_t extra) {
...
+        unemitted_bytes_ |= b;
+      }
     } else {
-      while (repetitions_ != 0) output_repetitions();
-      last_byte_ = b;
-      literal(b);
+      // We have some sort of run going.  As long as it is running we don't
+      // have unemitted bytes.
+      ASSERT(unemitted_bytes_valid_ == 0);
+      int shift = static_cast<int>(mode_ - 1) << 3;
```
floitsch, line 149: Add explicit values to the Mode constants and add a comment explaining that their value is crucial.
Alternatively make a map from mode to shift. (Almost prefer that).

src/nano_zlib.cc
```diff
@@ -45,18 +45,128 @@ word ZlibRle::add(const uint8* contents, intptr_t extra) {
...
+        unemitted_bytes_ <<= 8;
+        unemitted_bytes_valid_++;
+        unemitted_bytes_ |= b;
+      }
     } else {
-      while (repetitions_ != 0) output_repetitions();
-      last_byte_ = b;
-      literal(b);
+      // We have some sort of run going.  As long as it is running we don't
+      // have unemitted bytes.
```
floitsch, line 147: Do I understand this correctly, that the "mode" defines the size of the repeated bytes. Or (in other words) how far back each byte needs to look to find itself again?
If yes, mention that somewhere.

floitsch (approved):
LGTM.

src/nano_zlib.cc
erikcorry: Renamed to bytes_repeated_

src/nano_zlib.cc
erikcorry: This should be clearer now with the comment.  We have recently emitted the pattern abc and we are checking for the unemitted bytes abca.  The shift of 16 gets the a down to the low bits.

PR #172, opened 2021-12-20
Allow cmake to work without extra arguments.

floitsch:
This means that IDEs (like Kdevelop) can take the cmake-file as input
and understand the code.


kasperl (approved):
Looks good to me.

PR #173, opened 2021-12-20
Follow symlinks when trying to find the SDK root.

floitsch:
The toitc executable might be symlinked, so we need to follow it to find
the lib directory.

Also point the SDK to the actual SDK root (which should contain the 'lib' directory).


kasperl (approved):
Looks good to me.

PR #175, opened 2021-12-21
Add support for key value storage in flash

lask:


lib/device_impl.toit
```diff
@@ -4,6 +4,7 @@
 
 import drivers.cellular
 import device
+import encoding.json
```
kasperl, line 7: Is ubjson better?

lib/device_impl.toit
```diff
@@ -64,3 +73,42 @@ class Gnss_:
 
 get_mac_address_:
   #primitive.esp32.get_mac_address
+
+class Volume_:
+  name_ ::= ?
```
kasperl, line 78: Add types for these fields?

lib/device_impl.toit
```diff
@@ -64,3 +73,42 @@ class Gnss_:
...
+  read_only_ ::= ?
+
+  constructor.init_ .name_ .read_only_:
+
+  from name:
+    return KeyValue_ this name
+
+class KeyValue_:
+  group_ ::= ?
```
kasperl, line 87: Add type?

src/resources/primitive_kv_flash_posix.cc
```diff
@@ -0,0 +1,125 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// The license can be found in the file `LICENSE` in the top level
+// directory of this repository.
+
+#include "../top.h"
+
+#if defined(TOIT_LINUX) || defined(TOIT_BSD)
```
kasperl, line 18: How about Windows?

src/resources/primitive_kv_flash_posix.cc
```diff
@@ -0,0 +1,125 @@
...
+}
+
+PRIMITIVE(delete) {
+  ARGS(PersistentResourceGroup, resource_group, String, key);
+  USE(resource_group);
+
+  if (key->length() > static_cast<int>(MAX_KEY_LENGTH_)) INVALID_ARGUMENT;
+  String::Bytes bytes(key);
+  std::string str(char_cast(bytes.address()), bytes.length());
+  AllowThrowingNew unix_only;
```
kasperl, line 103: If this only runs on Windows, we should probably change the `unix_only` names in this file.

src/resources/primitive_kv_flash_esp32.cc
```diff
@@ -0,0 +1,151 @@
...
+
+  array->resize(length);
+
+  return array;
+}
+
+PRIMITIVE(write_bytes) {
+  ARGS(PersistentResourceGroup, resource_group, cstring, key, ByteArray, value);
+  if (!is_valid_key(key, process)) INVALID_ARGUMENT;
+  // The NVS code does not check for malloc failure.  See
```
kasperl, line 109: File issue on github.com/toitlang/toit and use that instead.

src/resources/primitive_kv_flash_esp32.cc
```diff
@@ -0,0 +1,151 @@
...
+
+  ByteArray::Bytes bytes(array);
+  err = nvs_get_blob(resource_group->handle(), key, bytes.address(), &length);
+  if (err == ESP_ERR_NVS_NOT_FOUND) {
+    return process->program()->null_object();
+  } else if (err != ESP_OK) {
+    return Primitive::os_error(err, process);
+  }
+
+  array->resize(length);
```
kasperl, line 101: Add a comment that explains how the length can change from the allocation of the byte array.

src/resources/primitive_kv_flash_esp32.cc
```diff
@@ -0,0 +1,151 @@
...
+
+  nvs_handle handle() { return _handle; }
+
+ private:
+  nvs_handle _handle;
+};
+
+bool is_valid_key(const char* key, Process* process) {
+  if (key[0] == '\0' || strlen(key) > MAX_KEY_LENGTH_) return false;
+  if (!process->is_privileged() && key[0] == '_') return false;
```
kasperl, line 47: I think this check should be present in the POSIX/Windows version too. Maybe add an `is_valid_key` helper there too?

src/resources/primitive_kv_flash_esp32.cc
```diff
@@ -0,0 +1,151 @@
...
+#if defined(TOIT_FREERTOS)
+
+#include <nvs_flash.h>
+
+#include "../resource.h"
+#include "../objects_inline.h"
+#include "../process.h"
+
+namespace toit {
+const size_t MAX_KEY_LENGTH_ = 15;
```
kasperl, line 27: Add newline after namespace?

kasperl (approved):

lib/device_impl.toit
lask: Probably. I will update it to ubjson. It should be no problem to change the format later because the partition should be erased across updates.

lib/device_impl.toit
lask: This is a resource proxy, so there is no good type for it.

lib/device_impl.toit
lask: Added.

src/resources/primitive_kv_flash_posix.cc
lask: Good point, maybe I should rename the file. `_desktop.toit`?

src/resources/primitive_kv_flash_posix.cc
lask: Probably not. I have revived the old NVS implementation, so I honestly don't know if this is still necessary.

src/resources/primitive_kv_flash_esp32.cc
lask: This issue relates to the implementation of NVS having a throwing `new`. However, as pointed out by @erikcorry, this problem may be fixed in the version of ESP IDF we depend on now. Should we create an issue to investigate or just wait and see?

src/resources/primitive_kv_flash_esp32.cc
lask: This is a revival of our old NVS implementation. I cannot find any justification for the resize, so I have removed it.

src/resources/primitive_kv_flash_esp32.cc
kasperl: Let's create an issue to investigate.

lib/device_impl.toit
```diff
@@ -42,20 +46,23 @@ class FlashStore_ implements device.Store:
...
   get key/string -> any:
-    throw "NOT IMPLEMENTED"
+    bytes := kv_store_.bytes key
+    if not bytes: return null
+
+    return ubjson.decode bytes
 
   delete key/string:
-    throw "NOT IMPLEMENTED"
+    return kv_store_.delete key
```
kasperl, line 55: I'd just drop the return. Doesn't look like delete is supposed to return anything (-> none).

kasperl (approved):
Still looks good to me.

src/resources/primitive_kv_flash_esp32.cc
lask: https://github.com/toitlang/toit/issues/179

PR #182, opened 2021-12-22
Move HTTP example

floitsch:
Otherwise simple examples (like `hello.toit`) won't work without a `pkg install`.


kasperl (commented):
Looks good to me.

PR #184, opened 2021-12-22
Fix LSP server for unsaved files on Windows.

### LINKED ISSUE #183: LSP server doesn't work for unsaved files on Windows
If a file has never been saved, then the LSP server on Windows doesn't work.

floitsch:
A virtual file is starting with '///'. When converting from compiler
paths to local paths, the language server would convert the slashes to
backslashes thus giving it a new name.

Fixes #183.

### DIFF
--- a/tools/toitlsp/lsp/compiler/path/path.go
+++ b/tools/toitlsp/lsp/compiler/path/path.go
@@ -21,6 +21,14 @@ import (
 	"strings"
 )
 
+const (
+	virtualFileMarker = "///"
+)
+
+func isVirtualPath(path string) bool {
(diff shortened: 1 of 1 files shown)


tools/toitlsp/lsp/compiler/path/path.go
```diff
@@ -21,6 +21,14 @@ import (
 	"strings"
 )
 
+const (
```
floitsch, line 24: This is currently duplicated from uri/convert.go, but Go doesn't allow cyclic imports, so removing the duplication would require a refactoring.
I might do it in a follow-up PR.

kasperl (approved):
Looks good to me.

PR #189, opened 2021-12-23
Add lsp tests and Toit LSP server.

floitsch:


tests/lsp/fail.cmake
```diff
@@ -0,0 +1,96 @@
+set(TOIT_FAILING_TESTS
```
floitsch, line 1: In this PR all tests are failing.
Follow-up PR is on its way.

tests/fail.cmake
```diff
@@ -0,0 +1,2 @@
+set(TOIT_FAILING_TESTS
```
kasperl, line 1: Should we have copyright headers in these files too? tests/CMakeLists.txt has such a header.

kasperl (approved):
Looks good to me.

tests/fail.cmake
floitsch: done.

PR #192, opened 2021-12-23
Add project_root_flag tests.

floitsch:
They don't seem to run at the moment. Needs investigation.

### DIFF
--- a/tests/project_root_flag/foo/src/foo.toit
+++ b/tests/project_root_flag/foo/src/foo.toit
@@ -1,0 +1,3 @@
+// Copyright (C) 2020 Toitware ApS. All rights reserved.
+
+identify: return "foo"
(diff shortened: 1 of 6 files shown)


tests/project_root_flag/foo/src/foo.toit
```diff
@@ -0,0 +1,3 @@
+// Copyright (C) 2020 Toitware ApS. All rights reserved.
```
kasperl, line 1: We should try to make sure all the tests have the same copyright header that includes the licensing information.

kasperl (approved):
Looks good to me.

tests/project_root_flag/foo/src/foo.toit
floitsch: Completely agree.
I will change the copyrights once the migration is complete.

PR #207, opened 2021-12-27
Improve gitversion.

floitsch:


src/CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ add_library(
   )
 
 set_source_files_properties(interpreter_run.cc PROPERTIES COMPILE_FLAGS "-O3 ${TOIT_INTERPRETER_FLAGS} $ENV{LOCAL_INTERPRETER_CXXFLAGS}")
-set_source_files_properties(utils.cc PROPERTIES COMPILE_FLAGS "-DTOIT_MODEL=\"\\\"${TOIT_MODEL}\\\"\" -DVM_GIT_INFO=\"\\\"${VM_GIT_INFO}\\\"\" -DVM_GIT_VERSION=\"\\\"${VM_GIT_VERSION}\\\"\"")
+set_source_files_properties(utils.cc PROPERTIES COMPILE_FLAGS "-DTOIT_MODEL=\"\\\"${TOIT_MODEL}\\\"\" -DVM_GIT_INFO=\"\\\"${VM_GIT_INFO}\\\"\" -DVM_GIT_VERSION=\"\\\"${TOIT_GIT_VERSION}\\\"\"")
```
kasperl, line 51: Where do we get the VM_GIT_INFO from?

kasperl (approved):
Looks good to me.

src/CMakeLists.txt
floitsch: It's completely unused ATM.
I already started removing it, but since it was even accessible from the Toit libraries I decided to keep that for a different PR. Also wasn't sure if there wasn't some use-case for it somewhere that I had missed.

PR #210, opened 2021-12-28
Always generate the SDK snapshots.

kasperl:


Makefile
```diff
@@ -115,6 +115,21 @@ build/host/:
 $(VERSION_FILE):
 	echo $(GIT_VERSION) > $@
 
+.PHONY: snapshots
```
floitsch, line 118: I think all of these should go into the cmake file (also because it has dependency tracking), but that's for a different PR.

floitsch (approved):
LGTM.

PR #211, opened 2021-12-28
Don't allow underscore in branch names.

floitsch:


floitsch:
TBR, as some tests rely on it.

PR #222, opened 2021-12-29
Add ctests.

floitsch:


floitsch:
No change except for copyright header. Otherwise directly copied from internal repo.

floitsch:
Needed to disable automatic generation of the ctest files with `build_tools`. (Was failing on Windows where the tests weren't compiling.)
Conceptually it feels also cleaner not to compile them with `build_tools`.

tests/ctest/ar_generator.cc
```diff
@@ -0,0 +1,83 @@
+// Copyright (C) 2020 Toitware ApS. All rights reserved.
```
kasperl, line 1: License for these files should be 0BSD?

kasperl (approved):
Looks good to me.

tests/ctest/ar_generator.cc
floitsch: I will change all the licenses in a later PR.

PR #198, opened 2021-12-23
Remove obsolete OTA initialization code

lask:
https://github.com/toitlang/toit/commit/746df36f67d563249612d012ab03aa50c9c1664f removed the OTA partitions from the partition table. However, the OTA initialization code was not removed. This PR removes the OTA initialization code which solves https://github.com/toitlang/toit/issues/186.

PR #224, opened 2021-12-29
Dynamically check for OTA support.

kasperl:
An alternative to https://github.com/toitlang/toit/pull/198 that leaves in the OTA support in case someone needs it.


floitsch (approved):
LGTM.

PR #227, opened 2021-12-29
Add minus_s tests.

floitsch:


tests/minus_s/run.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+# Copyright (C) 2021 Toitware ApS. All rights reserved.
```
kasperl, line 2: License?

kasperl (approved):
Looks good to me.

tests/minus_s/run.sh
floitsch: In a follow-up PR.

PR #228, opened 2021-12-29
Add profiler tests.

floitsch:


tests/profiler/basic_input.toit
```diff
@@ -0,0 +1,24 @@
+// Copyright (C) 2020 Toitware ApS. All rights reserved.
```
kasperl, line 1: License (maybe we should do a run through all files separately).

kasperl (approved):
Looks good to me.

PR #226, opened 2021-12-29
Add health tests.

floitsch:

PR #232, opened 2021-12-29
Shard tests.

floitsch:


floitsch:
We will need to adapt the `ctest` call in https://github.com/toitlang/toit/pull/226 to do the same sharding.

.github/workflows/ci.yml
```diff
@@ -124,22 +147,28 @@ jobs:
...
-          tar -czf build/${{ matrix.artifact }} -C ./build/host -h --exclude toit/bin/mbedtls toit
+          tar -czf build/${{ steps.constants.outputs.artifact }} -C ./build/host -h --exclude toit/bin/mbedtls toit
+
       - name: Upload artifacts
+        if: matrix.shard == 1
         uses: actions/upload-artifact@v2
         with:
-          name: ${{ matrix.os }}-build
-          path: build/${{ matrix.artifact }}
+          name: ${{ runner.os }}-build
```
floitsch, line 161: This changes the name of the artifacts from:
- `linux-build` to `Linux-build`
- `macos-build` to `macOS-build`
- `windows-build` to `Windows-build`

If the artifact names are important (for example fro Jaguar), then we could change it (using the `constants` step), or we would need to adapt it.

kasperl: I don't think we use these from Jaguar at all.

kasperl (approved):
Looks good to me.

PR #247, opened 2022-01-03
Trigger dispatch event

enyo:
This will trigger a dispatch event to the toitlang.org repo to rebuild the website when a new release has been created or an existing one was edited.


floitsch (approved):
LGTM.

PR #256, opened 2022-01-04
Fix health.

floitsch:


floitsch:
TBR.

tests/health/examples__ntp__ntp.toit.gold
```diff
@@ -0,0 +1,9 @@
+<...>/examples/ntp/package.lock:5:10: error: Package 'github.com/toitlang/pkg-ntp-1.0.0' not found
```
kasperl, line 1: Is this the expected output?

floitsch: On the buildbot it is because we don't fetch packages there.
We should probably start thinking about running the package manager before running the health-check. Things will get more interesting once our own Toit tools start relying on packages. At that point we need to run the pkg-tool even earlier.

PR #257, opened 2022-01-04
Reduce use of as_cstr.

erikcorry:
as_cstr relies on the null terminator for on-string
strings and doesn't work for slices.  This change
reduces the use of as_cstr.  Still some uses in
fork and exec.

### DIFF
--- a/src/primitive.h
+++ b/src/primitive.h
@@ -683,12 +683,47 @@ namespace toit {
 
 #define _A_T_cstring(N, name)                                           \
   Object* _raw_##name = __args[-(N)];                                   \
-  const char* name;                                                     \
+  char* _nonconst_##name = null;                                        \
+  if (_raw_##name != process->program()->null_object()) {               \
+    Blob _blob_##name;                                                  \
+    if (!_raw_##name->byte_content(process->program(), &_blob_##name, STRINGS_ONLY)) WRONG_TYPE; \
+    _nonconst_##name = unvoid_cast<char*>(calloc(_blob_##name.length() + 1, 1)); \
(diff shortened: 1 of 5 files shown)


src/primitive.h
```diff
@@ -683,12 +683,47 @@ namespace toit {
...
 #define _A_T_cstring(N, name)                                           \
   Object* _raw_##name = __args[-(N)];                                   \
-  const char* name;                                                     \
+  char* _nonconst_##name = null;                                        \
+  if (_raw_##name != process->program()->null_object()) {               \
+    Blob _blob_##name;                                                  \
+    if (!_raw_##name->byte_content(process->program(), &_blob_##name, STRINGS_ONLY)) WRONG_TYPE; \
+    _nonconst_##name = unvoid_cast<char*>(calloc(_blob_##name.length() + 1, 1)); \
+    if (!_nonconst_##name) MALLOC_FAILED;                               \
+    memcpy(_nonconst_##name, _blob_##name.address(), _blob_##name.length()); \
```
floitsch, line 692: This always copies the string, even if the internal string was already null-terminated (and isn't a slice or so).
Are you preparing for the future when this won't be the case anymore?

floitsch (approved):
LGTM.

An alternative would be to only use one part of the mbed-tls API consistently. We could just require that the certificates are always byte-arrays or strings and never use the other functionality. That would remove the `blob_or_string_with_terminating_null` type.

src/primitive.h
erikcorry: Yes.
Most of the places it is used are for small strings where I don't think it matters.  The exception is the certs where they have a non-trivial size, and where I still use them in-place until we have better memory management.

PR #258, opened 2022-01-05
Use cmake glob to find all toit files.

floitsch:


tests/health/CMakeLists.txt
```diff
@@ -13,23 +13,15 @@
 # The license can be found in the file `LICENSE` in the top level
 # directory of this repository.
 
-# Using GLOB is not recommended to find all files in a repository.
```
floitsch, line 16: The reason GLOB is discouraged is that the build tool (ninja) wouldn't know when to rebuild the source list.
However, generating the source-files ourselves with a `find` doesn't make things better. In fact it makes it worse.

kasperl (approved):
Looks good to me.

PR #259, opened 2022-01-05
Switch to cmake for running gold tests.

floitsch:


tests/negative/gold/as2_test.gold
```diff
@@ -1,3 +1,5 @@
 As check failed: a B is not a A.
   0: main                      tests/negative/as2_test.toit:8:5
   1: __entry__.<lambda>        <sdk>/core/entry.toit:48:20
+
```
kasperl, line 4: What's up with all the extra newlines?

kasperl (approved):
Looks good to me.

tests/negative/gold/as2_test.gold
floitsch: Previously we used `echo` and other bash commands to capture the output and store it in variables.
However, the bash trims trailing new lines, and the gold files thus didn't show them.

This is the actual output of the run:
```
ᐅ build/host/sdk/bin/toit.run tests/negative/as2_test.toit 
As check failed: a B is not a A.
  0: main                      tests/negative/as2_test.toit:8:5
  1: __entry__.<lambda>        <sdk>/core/entry.toit:48:20


ᐅ
```
I should probably look into the compiler and remove them. Filed https://github.com/toitlang/toit/issues/260

Issue #252, opened 2022-01-03
Error message through LSP is poorly formatted


kasperl:
In VS Code, I get errors like this:

```
Argument mismatch for 'bar'Too few arguments provided
```

Not sure where the problem stems from. Maybe it is a newline handling thing?

lask:
[The newline is definitely added in the compiler](https://github.com/toitlang/toit/blob/9acf6e2a949df699b1a2366523dbd526df6922e7/src/compiler/no_such_method.cc#L170). VS Code does ignore newlines in a number of places, so that is probably the problem.

floitsch:
Yes.
Erik asked me to add the newlines to make the error messages look better on the console.
I forgot to check how they were looking in vscode.
The easiest is probably to just replace newlines with ". " (dot, spaces) When sending them to the LSP client.
Alternatively I could also just revert the newline error message changes.

lask:
It looks like VS Code doesn't ignore newlines. I tested it with the VS Code sample LSP server found here https://github.com/microsoft/vscode-extension-samples/tree/main/lsp-sample by adding newlines to the error messages. This worked fine, so it would seem that it is possible to display error messages with newlines in VS Code.

floitsch:
Could easily be that the LSP protocol requires \r\n.

PR #261, opened 2022-01-05
Add newlines for lines read from error message in LSP

lask:
This fixes https://github.com/toitlang/toit/issues/252.
image

### DIFF
--- a/tools/toitlsp/lsp/compiler/parser.go
+++ b/tools/toitlsp/lsp/compiler/parser.go
@@ -122,10 +122,9 @@ func (p *parser) AnalyzeOutput(r io.Reader) (*AnalyzeResult, error) {
 				}
 				if !first {
 					msg += "\n"
+					first = false
 				}
 				msg += line
-
-				first = false
 			}
 
 			if !withPosition {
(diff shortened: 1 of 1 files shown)


tools/toitlsp/lsp/compiler/parser.go
```diff
@@ -119,7 +120,12 @@ func (p *parser) AnalyzeOutput(r io.Reader) (*AnalyzeResult, error) {
 				if line == "*******************" {
 					break
 				}
+				if !first {
+					msg += "\n"
```
jstroem, line 124: move `first = false` into this branch

jstroem (approved):
Thanks - good catch

tools/toitlsp/lsp/compiler/parser.go
```diff
@@ -111,6 +111,7 @@ func (p *parser) AnalyzeOutput(r io.Reader) (*AnalyzeResult, error) {
 			}
 
 			var msg string
+			first := true
```
floitsch, line 114: Maybe switch to `isFirst` but fine with `first` as well.

floitsch (approved):
LGTM.
thanks.

PR #262, opened 2022-01-05
Avoid unaligned reads in the interpreter

kasperl:


kasperl:
This does seem to come with a small slow-down. We should probably use the unaligned raw reads if the platform supports it, but for now this unblocks running on a few ARM targets that take offence when doing the unaligned reads.

erikcorry (approved):

erikcorry:
LGTM, but we probably need to be smarter for Xtensa, where gcc makes a dog's breakfast of this. https://godbolt.org/z/rT9xPMYj7

PR #266, opened 2022-01-06
Fix way of how environment is passed to command.

floitsch:
Under windows we can't just prefix the environment options.


kasperl (approved):
Looks good to me.

PR #272, opened 2022-01-07
Add agnostic pins

lask:
The cellular drivers use pins that are agnostic to their wiring. That is, they are agnostic to whether they are active high or active low. An implementation of this abstraction is not part of the standard library as it is now.

I am not sure about the naming. Suggestions are welcome.


floitsch:
Note that Erik added the `InvertedPin` to handle this use-case for displays: https://github.com/toitlang/toit/blob/150572a202fde92397796c861ca81107a5c6d15a/lib/gpio/pin.toit#L180

lask:
Cool! I will try to use that for the cellular drivers.

kasperl:
So we may not need this change?

PR #273, opened 2022-01-07
Start the sentence with the 'if'.

floitsch:
Humans have a tendency to stop reading after the first half of a
sentence. Make it clearer that the esp32 tools are only needed when
trying to build an image.


kasperl (approved):
Looks good to me.

PR #276, opened 2022-01-07
Add 32-bit platform.

floitsch:
Mark max_heap_size_test as failing on 32-bit machines.

### DIFF
--- a/toolchains/host32.cmake
+++ b/toolchains/host32.cmake
@@ -1,0 +1,38 @@
+# Copyright (C) 2019 Toitware ApS.
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; version
+# 2.1 only.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


toolchains/host32.cmake
```diff
@@ -0,0 +1,38 @@
...
+
+set(CMAKE_CXX_FLAGS_DEBUG "-Og -ggdb3 -rdynamic -fdiagnostics-color $ENV{LOCAL_CXXFLAGS}" CACHE STRING "c++ Debug flags")
+set(CMAKE_CXX_FLAGS_RELEASE "-Os $ENV{LOCAL_CXXFLAGS}" CACHE STRING "c++ Release flags")
+set(CMAKE_CXX_FLAGS_ASAN "-O1 -fsanitize=address -fno-omit-frame-pointer -g" CACHE STRING "c++ Asan flags")
+set(CMAKE_CXX_FLAGS_PROF "-Os -DPROF -pg" CACHE STRING "c++ Prof flags")
+
+set(CMAKE_SYSTEM_LIBRARY_PATH /lib32 /usr/lib32)
+set(FIND_LIBRARY_USE_LIB64_PATHS OFF)
+
+set(GOOS "linux")
```
kasperl, line 35: Is this tied to Linux? If so, we should probably call the file linux32.cmake.

floitsch: Good point. I think I can remove this line.

PR #277, opened 2022-01-07
Change submodule URL to public repository.

floitsch:
The relative path for git submodules is only really convenient when the
repository is private.


floitsch:
TBR.

kasperl (commented):
Looks good to me.

PR #282, opened 2022-01-10
Remove support for first-segment prefixes.

floitsch:
This removes the deprecation warning and simply uses the last segment.


kasperl (approved):
Looks good to me.

PR #283, opened 2022-01-10
Remove metrics library.

### LINKED ISSUE #238: Metrics library is empty and should be removed
Except for some constants there is nothing in the metrics library.
We should remove it.

floitsch:
Fixes #238.


kasperl (approved):
Looks good to me.

PR #287, opened 2022-01-11
Defragment handshake packets at Toit level.

erikcorry:
MbedTLS can't reassemble packets during TLS
hanshake, but some sites insist on breaking them
up.  We reassemble them in Toit to work around this.


CMakeLists.txt
```diff
@@ -106,7 +106,7 @@ endif()
 # that the allocation from MbedTLS is < 4k, 4033bytes to be precise.  We can
 # also change the input buffer length, but this requires that all communication
 # partners support the TLS protocol extension to tell them about this.
-set(MBEDTLS_C_FLAGS "-DMBEDTLS_SSL_IN_CONTENT_LEN=4608 -DMBEDTLS_SSL_OUT_CONTENT_LEN=3700 -DMBEDTLS_PLATFORM_MEMORY=1")
+set(MBEDTLS_C_FLAGS "-DMBEDTLS_SSL_IN_CONTENT_LEN=6000 -DMBEDTLS_SSL_OUT_CONTENT_LEN=6000 -DMBEDTLS_PLATFORM_MEMORY=1")
```
kasperl, line 109: Is this change still necessary? If so, you should definitely update the comment above.

lib/tls/session.toit
```diff
@@ -189,14 +192,115 @@ class Session:
...
+        while text_end < 100 and text_end < reader_.buffered and is_ascii_ (reader_.byte text_end):
+          text_end++
+        throw "Unknown TLS record type: $content_type - server replied '$(reader_.read_string text_end)'"
+      reader_.ensure 4  // 4 byte handshake message header.
+      message_size = (reader_.byte 1) << 16
+      message_size += (reader_.byte 2) << 8
+      message_size += (reader_.byte 3)
+      message_size += 4  // Encoded size does not include the 4 byte handshake header.
+
+    if message_size >= 0x4000: throw "TLS handshake message too large to defragment"
```
floitsch, line 273: Add comment that this is linked to the constant in the CMakeList? (If it is).

lib/tls/session.toit
```diff
@@ -189,14 +192,115 @@ class Session:
...
+    // Make an artificial record that was not on the wire.
+    record := ByteArray message_size + 5  // Include space for header.
+    record.replace 0 header
+    // Overwrite record size of header.
+    BIG_ENDIAN.put_uint16 record 3 message_size
+    remaining_in_record := BIG_ENDIAN.uint16 header 3
+    while message_size > 0:
```
floitsch, line 280: Feels wrong to call it `message_size` when it is changed. -> `remaining_message_size` ? (not really happy about the name, so if you have a better one...)

floitsch (approved):
LGTM.

We need to make sure the changes to the toolchains are correctly ported when we update the internal version.

erikcorry:
I improved the comments to reflect my better understanding of the issues.

PR #288, opened 2022-01-11
Fix unaligned word reads in image writing

kasperl:


kasperl:
TBR'ed by @erikcorry.

PR #290, opened 2022-01-11
Support FixedPoint numbers as input to JSON

erikcorry:


floitsch (commented):
I think that's the wrong way.
We can't just add a new entry for every possible class to the json library. I think the FixedPoint class should be in a package (as should maybe be the json library). In that case you wouldn't even be able to reference the class without adding a new dependency.

We need to think of how we want to support user-types (including `FixedPoint`).
One way is with `to_json` but I don't really like that either.
This leaves either:
- a block that does whatever is necessary
- a converter class (basically a more structured version of the block).
(or both).

erikcorry:
Deprecated in favour of https://github.com/toitlang/toit/pull/293

PR #291, opened 2022-01-11
Compile ESP32 sources in parallel.

floitsch:


Makefile
```diff
@@ -122,11 +124,20 @@ BIN_DIR = build/host/sdk/bin
 TOITVM_BIN = $(BIN_DIR)/toit.run$(EXE_SUFFIX)
 TOITC_BIN = $(BIN_DIR)/toit.compile$(EXE_SUFFIX)
 
+ifeq ($(DETECTED_OS), Linux)
+	NUM_CPU := $(shell nproc)
+else ifeq ($(DETECTED_OS), Darwin)
+	NUM_CPU := $(shell sysctl -n hw.ncpu)
+else
+	# Just assume two cores.
+	NUM_CPU := 2
```
erikcorry, line 133: I think you could be more generous here.

erikcorry (approved):

Makefile
floitsch: The buildbots only have 2 cores...

PR #295, opened 2022-01-12
Add per-process DNS cache

erikcorry:


lib/net/modules/dns.toit
```diff
@@ -174,15 +187,26 @@ class DnsQuery_:
       position += 2
       if clas != CLASS: throw (DnsException "Unexpected response class: $clas")
 
-      position += 4  // Skip TTL field.
+      ttl := BIG_ENDIAN.int32 response position
+      position += 4
+
+      // We won't cache more than a day, even if the TTL is very high.  (In
+      // practice TTLs over one hour are rare.)
+      ttl = min ttl 3600 * 24
```
floitsch, line 195:
```suggestion
      ttl = min ttl (3600 * 24)
```

lib/net/modules/dns.toit
```diff
@@ -209,3 +233,35 @@ class DnsQuery_:
         position_block.call position + 2
         return
     position_block.call position + 1
+
+/// Limit the size of the cache to avoid using too much memory.
```
floitsch, line 237:
```suggestion
/// Limits the size of the cache to avoid using too much memory.
```

lib/net/modules/dns.toit
```diff
@@ -209,3 +233,35 @@ class DnsQuery_:
...
+
+  // Set the limit a bit lower now - we want to remove at least one third of
+  // the entries when we trim the cache since it's an expensive operation.
+  if CACHE_.size < MAX_TRIMMED_CACHE_SIZE_: return
+
+  // Remove every second entry.
+  toggle := true
+  CACHE_.filter --in_place: | key value |
+    toggle = not toggle
+    toggle
```
floitsch, line 255: Would it make sense to remove either entries with low TTL, or to keep track of when items were last used?
In theory removing an entry and adding it again in the cache-lookup would make it such that the unused ones are in front.

lib/net/modules/dns.toit
```diff
@@ -209,3 +233,35 @@ class DnsQuery_:
...
+
+class CacheEntry:
+  server / string          // Unparsed server name like "8.8.8.8"
+  end / int                // Time in µs, compatible with Time.monotonic_us
```
floitsch, line 259:
```suggestion
  server / string          // Unparsed server name like "8.8.8.8".
  end / int                // Time in µs, compatible with Time.monotonic_us.
```

tests/dns_test.toit
```diff
@@ -19,3 +25,14 @@ main:
...
+
+cache_test:
+  // Prime cache.
+  dns_lookup "www.apple.com"
+
+  // We should easily be able to hit the cache in less than 5ms (a round trip
+  // lookup takes about 15ms).
+  start := Time.now
+  dns_lookup "www.apple.com"
+  duration := start.to Time.now
```
floitsch, line 37:
```suggestion
  duration := Duration.of: dns_lookup "www.apple.com"
```

floitsch (approved):
LGTM.

lib/net/modules/dns.toit
erikcorry: I don't want to remove the ones with low TTL, because it means some long-lived ones that are not used often can clog the cache.

It would probably be a good idea to remove the least recently used, but for now I want to keep code size down, and I think 99.9% of the time we won't hit this code at all - 30 entries is actually a lot for one process.

PR #297, opened 2022-01-12
Use pkgCommand as root command in toit.pkg.

floitsch:


tools/tpkg/main.go
```diff
@@ -66,6 +83,10 @@ func main() {
 			fmt.Fprintln(os.Stderr, e)
 		}
 	}
-	rootCmd.AddCommand(pkgCmd)
-	rootCmd.Execute()
+	deprecatedPkg := copyAsDeprecated(pkgCmd)
+
+	pkgCmd.Use = "toit.pkg"
+	pkgCmd.Short = "The toit package manager"
```
kasperl, line 89: toit -> Toit

tools/tpkg/main.go
```diff
@@ -22,27 +22,44 @@ import (
...
-)
-
 func getTrimmedEnv(key string) string {
 	return strings.TrimSpace(os.Getenv(key))
 }
 
 // Should be set with '-X' flag when linking.
 var sdkVersion = ""
 
+func copyAsDeprecated(cmd *cobra.Command) *cobra.Command {
```
kasperl, line 39: Does this mean anything for 'jag pkg'?

floitsch: No.
Jag uses the tpkg library the same way we do here.
Only difference is that Jag uses the pkgCommand as a subcommand and not as root command.

tools/tpkg/main.go
floitsch: done.

jstroem (approved):

PR #303, opened 2022-01-14
Simplify the RPC mechanism used for process -> system communcations.

kasperl:


tools/rpc.toit
```diff
@@ -13,133 +13,140 @@
...
+    id/int := decoded[0]
+    name/int := decoded[1]
+
+    send_exception_reply := (:| exception |
+      process_send_bytes_ pid type (ubjson.encode [ id, true, exception ])
+      return
+    )
+
+    procedures_.get name --if_present=: | procedure |
+      if queue_.add (RpcRequest_ pid gid id decoded[2] procedure): return
```
floitsch, line 35:
```suggestion
      request := RpcRequest_ pid gid id decoded[2] procedure
      if queue_.add request: return
```

tools/rpc.toit
```diff
@@ -13,133 +13,140 @@
...
+    decoded := ubjson.decode message
+    id/int := decoded[0]
+    name/int := decoded[1]
+
+    send_exception_reply := (:| exception |
+      process_send_bytes_ pid type (ubjson.encode [ id, true, exception ])
+      return
+    )
+
+    procedures_.get name --if_present=: | procedure |
```
floitsch, line 34: I think it's more natural to read this way:
```suggestion
    procedure := procedures_.get name
    if procedure:
```

tools/rpc.toit
```diff
@@ -13,133 +13,140 @@
...
 
-  constructor .task_cache_:
+    if decoded.is_empty:
+      send_exception_reply.call "Missing call context"
+    context ::= decoded[0]
+    if context is not int:
+      // TODO(kasper): This is a weird exception to pass back.
+      send_exception_reply.call "Closed descriptor $context"
+
+    handlers_.get name --if_present=: | handler |
```
floitsch, line 45: Same. I think it's more natural this way:
```suggestion
    handler := handlers_.get name
    if handler:
```

floitsch (commented):
DBC.
Definitely looks simpler (although I didn't look too much at the old code).

tools/rpc.toit
```diff
@@ -13,133 +13,141 @@
...
+  handlers_/Map ::= {:}
+  queue_/RpcRequestQueue_ ::= RpcRequestQueue_
+
+  on_message type gid pid message -> none:
+    assert: type == SYSTEM_RPC_CHANNEL_LEGACY_
+    decoded := ubjson.decode message
+    id/int := decoded[0]
+    name/int := decoded[1]
+
+    send_exception_reply := (:| exception |
```
erikcorry, line 29: missing space before |

tools/rpc.toit
```diff
@@ -13,133 +13,141 @@
...
+      reply := is_exception
+          ? [ id, true, exception.value, exception.trace ]
+          : [ id, false, result ]
+      process_send_bytes_ pid SYSTEM_RPC_CHANNEL_LEGACY_ (ubjson.encode reply)
+      return  // Stops any unwinding.
+
+monitor RpcRequestQueue_:
+  static MAX_TASKS ::= 4
+  static MAX_REQUESTS ::= 16
+  static IDLE_TIME_MS ::= 10_000
```
erikcorry, line 102: I think this is too long.  Compare with the time to do a GC caused by the new task being created, which is a few ms.  We want the extra tasks to be gone by the time the next memory high water arrives.

erikcorry (approved):

PR #309, opened 2022-01-14
Refactor LSP code.

floitsch:
Purpose of this refactoring:
- move LSP code into its own folder.
- make all compiler<->LSP-server communication go through "protocols".


src/compiler/compiler.cc
```diff
@@ -447,21 +448,16 @@ void Compiler::lsp_snapshot(const char* source_path,
   Flags::no_fork = true; // No need to fork the compiler when running in LSP mode.
   SnapshotBundle bundle = compile(source_path, configuration);
   if (!bundle.is_valid()) {
-    printf("FAIL\n");
+    configuration.lsp->snapshot()->fail();
     return;
   }
-  // The SnapshotBundle constructor copies all data.
```
floitsch, line 453: Common theme: the `printf`s have been moved into a "protocol" file.

src/compiler/compiler.cc
```diff
@@ -797,23 +791,18 @@ ast::Unit* Pipeline::parse(Source* source) {
   return parser.parse_unit();
 }
 
-void Pipeline::do_with_lsp_selection_handler(const std::function<void (LspSelectionHandler*)>& callback) {
-  callback(null);
+void Pipeline::setup_lsp_selection_handler() {
```
floitsch, line 794: Now that there is an `Lsp` class, it's easier to just set the handler there, than to use this `do_with_lsp_selection_handler`.
The `Pipeline::run` now calls `setup_lsp_selection_handler` before it does a `resolve`.

src/compiler/diagnostic.cc
```diff
@@ -264,41 +264,26 @@ void CompilationDiagnostics::emit(Severity severity,
   reset_colors(stdout);
 }
 
-static const char* severity_to_lsp_severity(Diagnostics::Severity severity) {
```
floitsch, line 267: Most of this code has been moved into the `protocol.cc` file.

src/compiler/lsp/completion.h
```diff
@@ -81,46 +83,20 @@ class CompletionHandler : public LspSelectionHandler {
...
+                                   const PackageLock& package_lock,
+                                   LspProtocol* protocol);
+  static void import_path(Symbol prefix,
+                          const char* path,
+                          Filesystem* fs,
+                          LspProtocol* protocol);
 
  private:
-  // From https://microsoft.github.io/language-server-protocol/specification#textDocument_completion
-  enum CompletionKind {
```
floitsch, line 89: Has been moved into its own file.

src/compiler/lsp/lsp.h
```diff
@@ -0,0 +1,106 @@
+// Copyright (C) 2019 Toitware ApS.
```
floitsch, line 1: Mostly fresh code.

It introduces an `Lsp` class as facade for all Lsp operations.

src/compiler/lsp/protocol.cc
```diff
@@ -0,0 +1,158 @@
+// Copyright (C) 2022 Toitware ApS.
```
floitsch, line 1: New file, but all of the code here has just been moved here (except some trivial forwarding functions).

src/compiler/lsp/protocol.h
```diff
@@ -0,0 +1,130 @@
+// Copyright (C) 2019 Toitware ApS.
```
floitsch, line 1: New.
We introduce an abstraction for talking to the Lsp Server.
Previously, code would just write to stdout.
Now we go through the protocol. Eventually we want to be able to change the protocol in one place.

kasperl (approved):
Looks good to me.

PR #310, opened 2022-01-15
Add SipHash

erikcorry:
This is one thing that would be needed for WireGuard.
Many of the others are available in MbedTLS and could be exposed.
The other missing one is Blake2s.
https://www.wireguard.com/protocol/


lib/crypto/siphash.toit
```diff
@@ -0,0 +1,57 @@
...
+*/
+
+/**
+Calculates the SipHash of the given $data.
+
+The $data must be a string or a byte array.
+
+The $key should be a 16 element byte array.
+
+The $output_length should be 8 or 16 bytes.
```
floitsch, line 20: ?
If not, when can I use what?
```suggestion
The $output_length must be 8 or 16 bytes.
```

lib/crypto/siphash.toit
```diff
@@ -0,0 +1,57 @@
...
+
+See https://en.wikipedia.org/wiki/SipHash
+*/
+
+/**
+Calculates the SipHash of the given $data.
+
+The $data must be a string or a byte array.
+
+The $key should be a 16 element byte array.
```
floitsch, line 18: ?
If not, when can I use something else?
```suggestion
The $key must be a 16 element byte array.
```

floitsch (approved):
LGTM.

PR #312, opened 2022-01-16
Compile under MSYS2

erikcorry:
MSYS2 is a somewhat Posixy environment on Windows with the tools
you are used to.  I found that the following packages gave a system
that could build Toit.

pacman -S make cmake ninja python3
pacman -S mingw-w64-x86_64-toolchain
pacman -S mingw-w64-x86_64-gcc
pacman -S mingw-w64-x86_64-dlfcn

Note that this is a Native Windows gcc so the end result is
Windows-native executables, not something that relies on MSYS2.
The only reason for msys2 is to make the developer's life easier
with command line tools like bash, git, vim, diffutils, patch, etc.


src/CMakeLists.txt
```diff
@@ -85,6 +85,12 @@ else()
   set(TOIT_NETWORK_LIBS mbedtls)
 endif()
 
+if (CMAKEISDUMB${CMAKE_SYSTEM_NAME} MATCHES "CMAKEISDUMBMSYS")
```
erikcorry, line 88: Even though "MSYS" is in quotes, Cmake will expand it because it's also a variable (apparently).  Putting a text in front prevents this.

floitsch: It looks like you can just change the CMP:
https://cmake.org/cmake/help/latest/policy/CMP0054.html#policy:CMP0054

`cmake_policy(SET CMP0054 NEW)`

floitsch (approved):
LGTM.

PR #313, opened 2022-01-17
Implement mkdtemp primitive on Windows

erikcorry:


src/primitive_file_win.cc
```diff
@@ -281,7 +326,56 @@ PRIMITIVE(mkdir) {
...
 PRIMITIVE(mkdtemp) {
-  UNIMPLEMENTED_PRIMITIVE;
+  ARGS(cstring, prefix);
+  DWORD ret;
+
+  bool in_standard_tmp_dir = false;
+  if (strncmp(prefix, "/tmp/", 5) == 0) {
+    in_standard_tmp_dir = true;
+    prefix += 5;
+  }
```
floitsch, line 336: I don't think you need to check for that.
The prefix should just be a prefix for the tmp directory.
For example, some of the directories created with `lsp_repro-` prefix:
```
lsp_repro-0uHt7f
lsp_repro-1fl2A8
lsp_repro-1I2HYr
lsp_repro-1sPJA6
lsp_repro-1vUdC9
lsp_repro-4jTjLL
lsp_repro-4v3TB7
lsp_repro-5GbI7q
lsp_repro-66nfUl
lsp_repro-6gdWvL
```

It's actually called "template" for the C function.
If we allow a full path in the C version, then we should probably fix that instead.

floitsch (approved):
LGTM.

Nice progress.

src/primitive_file_win.cc
erikcorry: The ar test uses a full path as the prefix.

floitsch: oops.
Should change the test...

erikcorry: After some offline discussion we agreed it was enough to document the current behaviour.

PR #319, opened 2022-01-19
Fix RPC error reply message type.

kasperl:


kasperl:
TBR'ed by @erikcorry.

PR #320, opened 2022-01-19
Inline small things

kasperl:


src/messaging.cc
```diff
@@ -26,9 +26,11 @@ enum MessageTag {
   TAG_NULL,
   TAG_TRUE,
   TAG_FALSE,
-  TAG_STRING,
   TAG_ARRAY,
+  TAG_STRING,
+  TAG_STRING_INLINE,
```
erikcorry, line 31: Add comment that encode_copy requires this tag order.

src/messaging.cc
```diff
@@ -278,15 +316,23 @@ Object* MessageDecoder::decode() {
...
-Object* MessageDecoder::decode_string() {
+Object* MessageDecoder::decode_string(bool inlined) {
   int length = read_cardinal();
-  uint8* data = read_pointer();
-  String* result = _process->object_heap()->allocate_external_string(length, data, true);
+  String* result = null;
+  if (inlined) {
+    ASSERT(length <= String::max_internal_size());
+    Error* error = null;
+    result = _process->allocate_string(reinterpret_cast<char*>(&_buffer[_cursor]), length, &error);
```
erikcorry, line 327: Comment that we can ignore the error because we are below the max internal string size, so we know it's an internal allcoation error.

erikcorry (approved):

PR #321, opened 2022-01-19
Add deprecation notices to cellular drivers

lask:


floitsch (approved):
LGTM.
Is there no health gold file that is triggering now?

I would maybe only add the deprecation warning to the note important methods. You should get at least one warning but you don't need to get one for every call you make.

PR #323, opened 2022-01-19
Keep track of unprocessed RPC requests.

kasperl:


tools/rpc.toit
```diff
@@ -117,16 +118,23 @@ monitor RpcRequestQueue_:
...
         // The task code runs outside the monitor, so the monitor
         // is unlocked when the requests are being processed but
         // locked when the requests are being dequeued.
         try:
-          while next := remove_first: next.process
+          while next := remove_first:
+            try:
+              next.process
+            finally:
+              unprocessed_--
```
kasperl, line 137: This probably doesn't have to be in a finally-block because `next.process` should never unwind, but being a little bit defensive feels right.

erikcorry (approved):

floitsch (approved):
LGTM.

PR #327, opened 2022-01-20
Removing debug information.

floitsch:
We haven't seen timeouts on the new buildbots and are not actively
working on this.
We can always add more debug info back.


kasperl (approved):
Looks good to me.

PR #329, opened 2022-01-20
Issue305 adc state

un01s:
Still not clear about the inner workings.

### DIFF
--- a/src/resources/adc_esp32.cc
+++ b/src/resources/adc_esp32.cc
@@ -64,18 +64,6 @@ static adc_atten_t get_atten(int mv) {
   if (mv <= 2200) return ADC_ATTEN_DB_6;
   return ADC_ATTEN_DB_11;
 }
-/*
-class AdcState : public SimpleResource {
- public:
-  AdcState(adc_unit_t unit, int chan)
-    : unit(unit)
-    , chan(chan) {}
-
-  adc_unit_t unit;
-  int chan;
-  esp_adc_cal_characteristics_t calibration;
-};
-*/
 
 AdcState::AdcState(SimpleResourceGroup* group) : SimpleResource(group) {
   // what to do with this?
--- a/src/resources/adc_esp32.cc
+++ b/src/resources/adc_esp32.cc
@@ -17,6 +17,8 @@
 
 #ifdef TOIT_FREERTOS
 
+#include "adc_esp32.h"
+
 #include <driver/gpio.h>
 #include <driver/adc.h>
 #include <esp_adc_cal.h>
@@ -62,9 +64,10 @@ static adc_atten_t get_atten(int mv) {
   if (mv <= 2200) return ADC_ATTEN_DB_6;
   return ADC_ATTEN_DB_11;
 }
-
-struct ADCState {
-  ADCState(adc_unit_t unit, int chan)
+/*
+class AdcState : public SimpleResource {
+ public:
+  AdcState(adc_unit_t unit, int chan)
     : unit(unit)
     , chan(chan) {}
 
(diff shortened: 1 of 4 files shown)


src/resources/adc_esp32.cc
```diff
@@ -63,20 +65,20 @@ static adc_atten_t get_atten(int mv) {
   return ADC_ATTEN_DB_11;
 }
 
-struct ADCState {
-  ADCState(adc_unit_t unit, int chan)
-    : unit(unit)
-    , chan(chan) {}
+AdcState::AdcState(SimpleResourceGroup* group) : SimpleResource(group) {
+  // what to do with this?
+}
```
erikcorry, line 70: Unless I'm missing something you can combine the group, the unit and the chan into one constructor with three arguments.  This is preferred, because then you can't forget to call init.

src/resources/adc_esp32.cc
```diff
@@ -104,27 +106,26 @@ PRIMITIVE(init) {
     }
   }
 
-  ADCState* state = _new ADCState(unit, chan);
+  AdcState* state = _new AdcState(group);
```
erikcorry, line 109: This code is OK, but it's actually written the wrong way around, before you even got to it.

The thing is that allocations off the GCed heap and on the GCed heap can fail, and the reaction to both is to return from the primitive with an error.  The primitive will be retried after a GC.

The state is allocated with new so it must be deleted with delete before we return an error.  The proxy is allocated on the GCed heap, so you don't need to free it, the GC will clean it up.  Therefore it would be better to allocate the proxy first, then create the AdcState.  That way you can drop the "delete state" because the state hasn't been 'new'ed yet when the proxy fails to allocate.

It's not a huge thing, it's just neater that way around.

src/resources/adc_esp32.cc
```diff
@@ -104,27 +106,26 @@ PRIMITIVE(init) {
...
   if (proxy == null) {
     delete state;
     ALLOCATION_FAILED;
   }
+  state->init(unit, chan);
 
   const int DEFAULT_VREF = 1100;
   esp_adc_cal_characterize(unit, atten, ADC_WIDTH_BIT_12, DEFAULT_VREF, &state->calibration);
 
+  proxy->set_external_address(state);
```
erikcorry, line 122: This uses template magic to also set the tag on the AdcState so that we can check the type on entry to `get` primitve :-)

src/resources/adc_esp32.cc
```diff
@@ -104,27 +106,26 @@ PRIMITIVE(init) {
...
   return proxy;
 }
 
 PRIMITIVE(get) {
-  ARGS(ByteArray, raw_state, int, samples);
-
-  ADCState* state = reinterpret_cast<ADCState*>(raw_state->as_external());
-  // Check if it's already closed.
-  if (state == null) INVALID_ARGUMENT;
+  ARGS(AdcState, state, int, samples);
```
erikcorry, line 129: I think you still need this so you get an exception instead of a null pointer crash if someone closes it twice.

src/resources/adc_esp32.h
```diff
@@ -0,0 +1,41 @@
...
+#include "../resource.h"
+
+namespace toit {
+
+class AdcState : public SimpleResource {
+ public:
+  TAG(AdcState);
+  AdcState(SimpleResourceGroup* group);
+
+  void init(adc_unit_t unit, int chan);
```
erikcorry, line 31: See above with combining init and the constructor.

erikcorry (commented):
Thanks for this.  I have a couple of nit-picks, and also some random comments that you don't need to do anything about!

src/resources/adc_esp32.cc
erikcorry: No I was wrong here, it's already in the macro.

kasperl:
@cla-bot check

erikcorry (approved):

erikcorry:
Thanks for the contribution.

I landed it for you over at #446

PR #330, opened 2022-01-20
Introduce a lsp-fs protocol abstraction.

floitsch:
No new functionality.
Introduces an abstraction layer that separates the file-server communication from the file-server protocol.


kasperl (approved):
Looks good to me.

PR #331, opened 2022-01-20
Add abstraction layer for LSP-compiler communication.

floitsch:
Makes all LSP communication go through a writer interface (`LspWriter`).


src/compiler/lsp/protocol.h
```diff
@@ -48,13 +49,57 @@ struct LspRange {
 int utf16_offset_in_line(Source::Location location);
 LspRange range_to_lsp_range(Source::Range range, SourceManager* manager);
 
+struct LspWriter {
+  virtual ~LspWriter() {}
+  virtual void printf(const char* format, va_list& arguments) = 0;
```
kasperl, line 54: I don't think it is idiomatic to take references to va_lists? They are usually just passed by value, no?

kasperl (approved):
Looks good to me.

src/compiler/lsp/protocol.h
floitsch: Not completely sure.
Currently everything is using `va_list&`, so this use here is consistent.
I can change to `va_list` in a separate PR. Fundamentally, I prefer the `&`, as it makes clear that using the data could change the caller's `va_list`. From what I gather, this is platform dependent. Some copy the data, some give a pointer.

PR #332, opened 2022-01-20
Add defer functionality.

floitsch:


kasperl (approved):
Looks good to me.

src/compiler/compiler.cc
```diff
@@ -328,26 +328,37 @@ void Compiler::language_server(const Compiler::Configuration& compiler_config) {
...
+    fs_protocol = _new LspFsProtocol(connection);
+    fs = _new FilesystemLsp(fs_protocol);
   }
 
-  LspWriterStdout stdout_writer;
-  LspProtocol lsp_protocol(&stdout_writer);
-  Lsp lsp(&lsp_protocol);
+  // We generally don't explicitly keep track of memory, but here we might need
+  // to release resources.
+  Defer del_fs { [&]() {
```
erikcorry, line 344: I think the () is optional in lambda syntax.  Could be a later version of C++ though.

erikcorry (approved):

src/compiler/compiler.cc
floitsch: good to know. Will switch to it in a later PR.

PR #333, opened 2022-01-20
Refactor Toit version of LSP server.

floitsch:
Separates the FileServer protocol from the communication.


kasperl (approved):
Looks good to me.

PR #337, opened 2022-01-20
Allow reading from ReaderWriter when there is data.

### LINKED ISSUE #336: ReaderWriter doesn't allow reads while the buffer isn't full
The read function doesn't work unless the buffer is full, or the writer closes the pipe.

floitsch:
Fixes #336.

### DIFF
--- a/lib/reader_writer.toit
+++ b/lib/reader_writer.toit
@@ -58,19 +58,16 @@ class ReaderWriterReader_ implements CloseableReader:
 // than privacy to indicate which methods are synchronized.
 monitor ReaderWriterHelper_:
   buffer_size_/int ::= ?
-  buffer_/ByteArray := ?
-  outgoing_ := null
+  buffer_/ByteArray? := null
+  // Buffer to reuse instead of allocating a new one.
+  reuse_/ByteArray? := null
   fullness_ := 0
   writer_closed_ := false
   reader_closed_ := false
 
   constructor .buffer_size_:
-    buffer_ = ByteArray buffer_size_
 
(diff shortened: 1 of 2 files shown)


lib/reader_writer.toit
```diff
@@ -80,27 +77,38 @@ monitor ReaderWriterHelper_:
...
   read:
-    await: outgoing_ or writer_closed_
-    if outgoing_:
-      result := outgoing_
-      outgoing_ = null
+    await: buffer_ or writer_closed_
+    result := ?
+    if buffer_:
+      if fullness_ != buffer_.size:
+        result = buffer_[..fullness_]
```
erikcorry, line 106: I think this violates the expectation that read will return fresh byte arrays that can be held onto.  Subsequent calls to read will return arrays that alias each other.

floitsch: refactored.

lib/reader_writer.toit
```diff
@@ -80,27 +76,25 @@ monitor ReaderWriterHelper_:
...
 
   read:
-    await: outgoing_ or writer_closed_
-    if outgoing_:
-      result := outgoing_
-      outgoing_ = null
+    await: fullness_ != 0 or writer_closed_
+    result := ?
+    if fullness_ != 0:
+      result = buffer_[..fullness_].copy
```
erikcorry, line 96: I think copy takes from and to, so you are not creating a slice object here for no reason.

floitsch: done.

PR #347, opened 2022-01-24
Add support for slices and CoW.

kasperl:


floitsch (approved):
LGTM.
In theory it would make sense to only make a copy if the CoW byte-array already has a copy. Otherwise sending a slice of the original data would be more efficient.

floitsch:
Could make sense to add a test for the CoW byte-array when it has been modified too.

PR #348, opened 2022-01-24
Use auto-frequency detection.

floitsch:


floitsch:
I couldn't get any tool to run, so I just changed the values in the file.

I got the '0' from here: https://github.com/lab11/esp32-apps/blob/master/ble_multiple_connect/sdkconfig.defaults

erikcorry (approved):

PR #349, opened 2022-01-24
Refactor host package.

floitsch:
All three packages are in the process of getting their own repositories.


kasperl (approved):
Looks good to me.

PR #350, opened 2022-01-25
Dispose resource states properly.

kasperl:


src/primitive_events.cc
```diff
@@ -53,4 +53,15 @@ PRIMITIVE(register_object_notifier) {
   return process->program()->null_object();
 }
 
+PRIMITIVE(unregister_object_notifier) {
+  ARGS(ByteArray, group_proxy, ByteArray, resource_proxy);
```
erikcorry, line 57: You can just declare these as ResourceGroup and Resource.

erikcorry (approved):
Nice.

src/primitive_events.cc
kasperl: Not if they are closed.

PR #351, opened 2022-01-25
Fix GC logs by freeing blocks before reporting memory after GC.

kasperl:


src/heap.cc
```diff
@@ -502,78 +502,81 @@ int ObjectHeap::scavenge() {
...
-    if (heap->_finalizer_notifier != null) {
-      heap->_finalizer_notifier->notify();
-    }
-    return true; // Remove node from list.
-  });
 
-  // Process the finalizers in the to space.
-  while (!objects.eos()) ss.process_to_objects(objects);
-  ASSERT(objects.eos());
+  { ScavengeState ss(this);
```
kasperl, line 506: The change is just making sure that the ScavengeState destructor gets called before we compute the amount of memory available after the GC.

erikcorry (approved):

PR #356, opened 2022-01-25
Stop extending tls.Session

kasperl:


erikcorry (approved):
Probably not backwards compatible, but hopefully nobody was typing their tls socket as a session.

PR #358, opened 2022-01-26
Remove object snapshots, only program snapshots

erikcorry:
I think for something like this we want to use Kasper's new C++ code that
is used for RPC.  It's part of separating out the object heap and the
program heap so we can change the GC.


kasperl (approved):
Looks good to me.

src/primitive_core.cc
```diff
@@ -445,6 +445,7 @@ PRIMITIVE(read_int_little_endian) {
 
 PRIMITIVE(args) {
   char** argv = process->args();
+#ifndef TOIT_FREERTOS
```
floitsch, line 448: This feels like a completely unrelated change.
If we don't want to have compile-time arguments then we should remove them in a different PR.

src/snapshot.cc
```diff
@@ -33,7 +33,7 @@ enum class SnapshotTypeTag {
   OBJECT_TAG = 0,
   IN_TABLE_TAG,
   BACK_REFERENCE_TAG,
-  PROGRAM_HEAP_REFERENCE_TAG,
+  UNUSED_TAG,
```
floitsch, line 36: I would prefer if it was removed, but since snapshots are disappearing soon anyway...

tools/toit.run.toit
```diff
@@ -17,7 +17,6 @@
 // only: To make stack traces readable.
 
 import ar show *
-import debug.rpc show RPC_ECHO
 import .debug_message
```
floitsch, line 20:
```suggestion
```

lib/debug/debug.toit
```diff
@@ -5,10 +5,6 @@
 import .rpc
 import rpc
```
floitsch, line 6: Same for the 'echo.toit'
```suggestion
```

floitsch (approved):
LGTM.

src/objects.h
```diff
@@ -460,6 +460,7 @@ class ByteArray : public HeapObject {
     *extra_bytes = raw_length;
   }
 
+#ifndef TOIT_FREERTOS
```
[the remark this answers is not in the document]
erikcorry: Removed them instead.

src/primitive_core.cc
erikcorry: Change removed.

erikcorry (commented):
All other changes implemented.

PR #363, opened 2022-01-27
Avoid leaking external memory for TLS sessions

kasperl:


src/heap.h
```diff
@@ -281,7 +281,9 @@ class ObjectHeap final : public Heap {
   ByteArray* allocate_proxy(int length, uint8* memory, bool dispose = false) {
     return allocate_external_byte_array(length, memory, dispose, false);
   }
-  ByteArray* allocate_proxy() { return allocate_proxy(0, null); }
```
erikcorry, line 284: This would be so much more readable if it took an enum instead of a bool, so you could write Heap::DISPOSE as the argument.

src/resources/tls.cc
```diff
@@ -680,14 +680,14 @@ PRIMITIVE(set_session) {
     return tls_error(null, process, result);
   }
 
+  // Set the session and remember to always free the fake session
```
erikcorry, line 683: Is it fake, or does mbedtls_ssl_set_session just take a copy?

erikcorry (approved):

src/resources/tls.cc
kasperl: I just used the existing terminology from the deleted comment in line 688.

src/heap.h
kasperl: Yeah, that would be nicer. We should fix that.

PR #364, opened 2022-01-27
Fix 64-bit external string length handling

kasperl:


floitsch (approved):
LGTM.
I would have expected you to change the sentinel to -1, but this works too :)

PR #367, opened 2022-01-27
Enable user RTC memory

lask:


lib/esp32.toit
```diff
@@ -100,3 +100,19 @@ image_config_ -> ByteArray:
 
 set_real_time_clock_ seconds/int ns/int -> none:
   #primitive.core.set_real_time_clock
+
+/**
+Sets the RTC user data to the given $data starting from the given $offset.
+
+The given $data must be within bounds. That is 0 <= $offset <= $offset + data.size < 2048.
+*/
+set_rtc_data data offset/int=0 -> none:
```
floitsch, line 109: Depending on how nice we want to make the RTC api, we could provide a singleton with `operator[]` ..., but as an underlying API this is probably good enough.

lib/esp32.toit
```diff
@@ -100,3 +100,19 @@ image_config_ -> ByteArray:
 
 set_real_time_clock_ seconds/int ns/int -> none:
   #primitive.core.set_real_time_clock
+
+/**
+Sets the RTC user data to the given $data starting from the given $offset.
```
floitsch, line 105: Add some explanation what RTC is.

lib/esp32.toit
```diff
@@ -100,3 +100,19 @@ image_config_ -> ByteArray:
...
+
+The given $data must be within bounds. That is 0 <= $offset <= $offset + data.size < 2048.
```
floitsch, line 107: I would add a constant `RTC_SIZE` and then reference it here:
```suggestion
The given $data must be within bounds. That is 0 <= $offset <= $offset + data.size < $RTC_SIZE.
```

src/rtc_memory_esp32.h
```diff
@@ -24,6 +24,8 @@ namespace toit {
 // The RTC memory holds state that is preserved across reboots.
 class RtcMemory {
  public:
+  static const int RTC_USER_DATA_SIZE = 4096;
```
floitsch, line 27: In the comment for the Toit functions it says 2K.

floitsch (approved):
LGTM.

src/rtc_memory_esp32.h
lask: You asked about the amount of RTC memory we have available. It seems like we have a total of 8K available. We use a bit for boot count and related things, so we could in principle bump this further.

It should be no problem to increase it (maybe we shouldn't have a default value for the length in the toit-level primitive calls if we want to change it).

lib/esp32.toit
lask: Called it `RTC_MEMORY_SIZE`.

lib/esp32.toit
lask: Added advanced section with a bit of explanation.

lib/esp32.toit
lask: We could definitely do that, but we should probably add a primitive specifically for that to reduce the overhead.

lask: Maybe the correct thing to do is to make a `ByteArray` with an external allocation that points to the RTC memory.

lask:
I added an alternative implementation that uses external byte arrays: https://github.com/toitlang/toit/pull/368

lask:
Abandoning this PR in favor of #368.

PR #368, opened 2022-01-28
Allow access to RTC memory 

lask:


lask:
This is an alternative to https://github.com/toitlang/toit/pull/367

floitsch (approved):
LGTM.

Please check that the byte-array can be used for primitive operations, like:
- send through WiFi
- write to a file
- use `replace` (which uses optimized code).
-

lask:
Wifi tested with
```
main:
  network := net.open
  client := http.Client network
  host := "postman-echo.com"
  esp32.rtc_user_bytes.fill 5
  response := client.post esp32.rtc_user_bytes --host=host --path="/post"

  while data := response.body.read:
    print "Read $data.to_string_non_throwing"
```

lask:
`replace` tested with
```
import esp32

main:
  rtc_bytes := esp32.rtc_user_bytes
  rtc_bytes = rtc_bytes[..10]
  print rtc_bytes

  other_bytes := ByteArray 10: it
  rtc_bytes.replace 0 other_bytes
  print rtc_bytes

  other_bytes = ByteArray 10: 255 - it
  other_bytes.replace 0 rtc_bytes
  print rtc_bytes

  rtc_bytes.fill 0

  esp32.deep_sleep
    Duration --s=3
```

lask:
The RTC memory is only available on ESP32 where we do not have a file system, so I don't think it makes sense to test it with writing to a file.

PR #372, opened 2022-01-28
Add RTC backed byte array with checksum

lask:


lask:
Moving this to a package.

PR #377, opened 2022-01-28
Use resource framework to free up open dirs

erikcorry:


src/primitive_file_posix.cc
```diff
@@ -145,35 +145,36 @@ PRIMITIVE(open) {
   return Smi::from(fd);
 }
 
-class Directory {
+class Directory : public SimpleResource {
  public:
   TAG(Directory);
-  DIR* dir;
+  Directory(SimpleResourceGroup* group, DIR* dir) : SimpleResource(group), _dir(dir) { }
+  ~Directory() { closedir(_dir); }
```
kasperl, line 152: What happens if _dir is null?

packages/host/src/directory.toit
```diff
@@ -95,17 +95,17 @@ class DirectoryStream:
...
+      byte_array := readdir_ dir_
       if not byte_array: return null
       str := byte_array.to_string
       if str == "." or str == "..": continue
       return str
 
   close -> none:
-    fd := fd_
-    closedir_ fd
+    dir := dir_
```
kasperl, line 105: What's up with this dance? Should we set dir_ to null here?

kasperl (approved):
Looks good to me.

src/primitive_file_posix.cc
erikcorry: That can't happen.

packages/host/src/directory.toit
erikcorry: That depends on whether we want to silently allow multiple close calls.  I think in general that's just nicer.  If we leave it as is the underlying Directory is deleted, but the proxy lives on with a null pointer in it.  That means we get an ALREADY_CLOSED exception if we call close twice.

PR #382, opened 2022-01-28
Download host, tar and ar package

floitsch:
Since these are now packages don't keep a copy around.


tests/package.lock
```diff
@@ -1,11 +1,17 @@
...
-    path: ../packages/tar
-  ar:
-    path: ../packages/ar
+  pkg-ar:
+    url: github.com/toitlang/pkg-ar
+    version: 1.0.0
+    hash: 175bcdfd59b2597c0825ef8765a20498a7e6a2db
+  pkg-host:
+    url: github.com/toitlang/pkg-host
+    version: 1.0.0
```
floitsch, line 12: This is now the wrong as the host-pkg has the wrong primitive interface.
Not completely sure how we want to fix this.

kasperl (approved):
Looks good to me.

PR #385, opened 2022-01-28
Run external package tests on the buildbot.

floitsch:
The external tests run with the normal build-tests, but they are allowed to be negative.
We collect their result and forward it to a new job. This new job is not marked as required and may thus go red without blocking a commit.


.github/workflows/ci.yml
```diff
@@ -155,6 +157,43 @@ jobs:
             ninja check
           fi
 
+      # We use a cache to send the external test result to the external job.
+      - name: Retrieve/Store external test result
```
kasperl, line 161: Retrieve/store ... results

tests/health/fail.cmake
```diff
@@ -18,6 +18,6 @@ set(TOIT_FAILING_TESTS
 
 if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows" OR "${CMAKE_SYSTEM_NAME}" STREQUAL "MSYS")
   list(APPEND TOIT_FAILING_TESTS
-    tests/lock_file/multi_test/target/sub/src/sub.toit
+    health-tests/lock_file/multi_test/target/sub/src/sub.toit
```
kasperl, line 21: What's the easiest of reading this? Maybe split the prefix (the stuff controlled by the test infrastructure) and the stuff controlled by the external package, so it is easier to find the right stuff.

$(HEALTH_PREFIX)/sub/src/sub.toit

kasperl (approved):
Looks good to me.

tests/health/fail.cmake
floitsch: Added a HEALTH_TEST_PREFIX in https://github.com/toitlang/toit/pull/383

.github/workflows/ci.yml

PR #387, opened 2022-01-28
Add message to static assert

kasperl:


floitsch (commented):
LGTM.
Classic mistake ;)

PR #390, opened 2022-01-29
Make directory change backwards compatible

erikcorry:


packages/host/src/directory.toit
```diff
@@ -106,9 +106,13 @@ class DirectoryStream:
       closedir_ dir_
       dir_ = null
 
-opendir_ resource_group name:
+// No longer in use.
```
floitsch, line 109:
```suggestion
// For backwards compatibility.
// Deprecated as of 2022-01.
```

floitsch (approved):
LGTM.

PR #392, opened 2022-01-30
Separate heaps

erikcorry:
This is almost all just separating and duplicating code.

Probably both heaps can be simplified a lot after this.

The page size is still shared between the program and object heaps.
A next step could be to make the program heap detection range
based on non-embedded so that the page sizes are not tied.


src/compiler/CMakeLists.txt
```diff
@@ -41,6 +41,7 @@ target_link_libraries(
   toit.compile
   toit_core
   toit_compiler
+  toit_core
```
kasperl, line 44: This is already in line 42?

src/primitive_core.cc
```diff
@@ -2225,13 +2225,11 @@ PRIMITIVE(literal_index) {
   ARGS(Object, o);
   auto null_object = process->program()->null_object();
   if (!o->is_heap_object()) return null_object;
-  auto heap_object = HeapObject::cast(o);
-  if (!Heap::in_read_only_program_heap(heap_object, process->object_heap())) return null_object;
```
kasperl, line 2229: Was it necessary to fix this or did you just decide to remove the fast case? I assume the fast (fail) case isn't used by toitp.

kasperl (approved):
Looks good to me.

src/primitive_core.cc
erikcorry: We have circular dependencies between .a files - not sure why we didn't have this issue before, but this makes it link.

erikcorry: Maybe it wasn't necessary.

PR #397, opened 2022-01-31
Improve git-version for release branches.

floitsch:


kasperl (commented):
Looks good to me, but we have at least two other copies of gitversion that we need to deal with before this has any impact.

kasperl:
Hmm. Once you've created a release branch, we also need the master branch to bump its minor version....

floitsch:
Master branch is currently handled by line 89+.
But that's due to a bug. Will fix in a follow-up PR.

floitsch:
actually might as well fix it in this one.

floitsch:
Should now work as intended for master/main/HEAD.

kasperl:
So are you saying that if builds from the master branch are v1.6.0-pre.18 and we create a branch called release-v0.16, the first commit on the branch will keep the v1.6.0-pre.18 name and the first commit on master after that will be called v1.7.0-pre.0? That's how I think it should be.

floitsch:
Hmm. Maybe not. It might need some kind of tag on master/main for that first.
I can add a check to see if a branch exists.

kasperl (approved):
Looks good to me.

PR #400, opened 2022-02-01
Factor out the DNS-over-UDP code

erikcorry:
This mostly just moved the DNS-over-UDP code to its own lib file.

The lookup function takes a block which creates the right sort of
UDP socket if needed.

The idea of this is to make it easier to use the DNS-over-UDP code
for cellular drivers.


lib/dns_over_udp.toit
```diff
@@ -0,0 +1,277 @@
...
+  dots := 0
+  name.do:
+    if it == '.':
+      dots++
+    else:
+      if not '0' <= it <= '9': return false
+  return dots == 3
+
+/**
+Look up a domain name.
```
lask, line 33:
```suggestion
Looks up a domain name.
```

lib/dns_over_udp.toit
```diff
@@ -0,0 +1,277 @@
...
+    --timeout/Duration=DNS_DEFAULT_TIMEOUT:
+  if ip_string_ host:
+    return net.IpAddress.parse host
+  q := DnsQuery_ host
+  return q.get --server=server --timeout=timeout get_udp_socket
+
+class DnsQuery_:
+  static A     ::= 1  // A address.
+  static CNAME ::= 5  // Canonical name.
+  static CLASS ::= 1  // The Internet class.
```
lask, line 59: IMO make the names so the comments are unnecessary by including all the necessary information, e.g. ` A_ADDRESS`.

lib/dns_over_udp.toit
```diff
@@ -0,0 +1,277 @@
...
+      --timeout/Duration=DNS_DEFAULT_TIMEOUT
+      [get_udp_socket]:
+    if HOSTS_.contains name:
+      return net.IpAddress.parse HOSTS_[name]
+
+    hit := find_in_cache_ server
+    if hit: return hit
+
+    query := create_query_
+    socket := get_udp_socket.call
```
lask, line 88: Maybe add a type to have a dynamic check that it is the correct socket.
```suggestion
    socket/udp.Socket := get_udp_socket.call
```
You would also need to have the import:
```
import net.udp
```
Finally, we should probably mention it in the toit doc.

lask: I looked at the previous implementation and it should create a socket of the correct type.

lib/dns_over_udp.toit
```diff
@@ -0,0 +1,277 @@
...
+
+        retry_timeout := DNS_RETRY_TIMEOUT
+
+        // Resend the query with exponential backoff until the outer timeout
+        // expires.
+        while true:
+          socket.write query
+
+          answer := null
+          exception := catch:
```
lask, line 104:
```suggestion
          exception := catch --trace=(: it != "DEADLINE_EXCEEDED"):
```

lib/dns_over_udp.toit
```diff
@@ -0,0 +1,277 @@
...
+            with_timeout retry_timeout:
+              answer = socket.receive
+          if exception and exception != "DEADLINE_EXCEEDED": throw exception
+
+          if answer:
+            return decode_response_ answer.data server
```
lask, line 110: One-liner like above?

lask (commented):
I think this is a great candidate for a package, but I am also fine with it as a library.

It is less inconvenient for user when we update the interface of a a package than when we update  the interface of a library.

erikcorry:
This is probably not the approach we will go with.

PR #401, opened 2022-02-01
Add metadata (heap organization) files from Dartino.

erikcorry:


src/third_party/dartino/AUTHORS
```diff
@@ -0,0 +1,8 @@
+# Below is a list of people and organizations that have contributed
+# to the Dartino project. Names should be added to the list like so:
+#
+#   Name/Organization <email address>
```
kasperl, line 4: Maybe change this?

PR #402, opened 2022-02-01
Toit naming convention

erikcorry:


kasperl (approved):
Looks good to me. Consider #pragma once.

PR #405, opened 2022-02-02
Handle timeouts in RPC calls

kasperl:


lib/rpc/broker.toit
```diff
@@ -40,8 +41,12 @@ class RpcBroker implements SystemMessageHandler_:
     procedures_.remove name
 
   on_message type gid pid message/List -> none:
-    assert: type == SYSTEM_RPC_REQUEST_
     id/int := message[0]
+    if type == SYSTEM_RPC_CANCEL_:
+      queue_.cancel id
```
lask, line 46: What prevents different tasks from accidentally cancelling each others requests? If they have picked the same ID for a request, then if one cancels, then both are cancelled.

lib/rpc/broker.toit
```diff
@@ -134,3 +155,34 @@ monitor RpcRequestQueue_:
...
+        unprocessed_--
+        result++
+      previous = current
+      current = next
+    // Then we cancel any requests that are in progress by canceling the
+    // associated processing task.
+    processing_requests_.size.repeat:
+      request/RpcRequest_? := processing_requests_[it]
+      if request and request.id == id:
+        processing_tasks_[it].cancel
```
lask, line 186: Is there any change that we leave requests unprocessed? Normally tasks stay alive until there is no more work to do, but if all worker tasks are cancelled then we may have unprocessed request left in the pool that won't be processed until a new worker task is created.

lib/rpc/broker.toit
kasperl: Yeah, you're right. We need to check the process id too.

lib/rpc/broker.toit
kasperl: No, I don't think so, but that's because of a bug: Canceled processing tasks stick around, but don't do much processing after that. I think I've worked around that issue.

kasperl: That is: Canceled processing tasks are now removed from the processing pool and new tasks are spawned on demand.

lib/rpc/broker.toit
```diff
@@ -98,39 +114,81 @@ monitor RpcRequestQueue_:
...
             try:
+              processing_requests_[task_index] = next
               next.process
             finally:
               // This doesn't have to be in a finally-block because the call
               // to 'next.process' never unwinds, but being a little bit
               // defensive feels right.
+              processing_requests_[task_index] = null
               unprocessed_--
         finally:
```
lask, line 191: I suppose it is the same for this `finally`; `process` will return from its `finally` which stops any on going unwinding, so the `finally` is strictly not necessary? I at least used this line of thinking to convince myself that a critical block wasn't necessary to prevent a cancel from stopping `ensure_processing_task_` prematurely.

kasperl: The finally here is more for if `remove_first` or any other stuff outside the inner try-block should error out.

lask (approved):
lgtm

PR #406, opened 2022-02-02
Roll to esp-idf with better memory accounting.

erikcorry:
Also has a fix to hopefully compile for RISCV.


kasperl (approved):
Looks good to me.

PR #408, opened 2022-02-02
BLE server

mikkeldamsgaard:
This implements a BLE server for toit.

Special care has been taken to keep backwards compatibility and to keep the number of new primitives to a minimum.

The main idea is to setup the ble config in a separate resource group and when the Device is initialised then setup the nimble host stack.

To separate client connections and connections made to other server, a small rearrangement of the original implementation had to be changed, so that the connection and disconnection events are send on the GATT resource.

Added an example of how to use the lib.


kasperl:
@cla-bot check.

examples/ble/heartrate.toit
```diff
@@ -0,0 +1,53 @@
+// The standard ble peripheral demo
```
kasperl, line 1: Add the same copyright and license information as used in the other examples.

examples/ble/heartrate.toit
```diff
@@ -0,0 +1,53 @@
...
+  heartrate_receive := service.add_write_only_characteristic
+    Uuid #[0x63, 0x4b, 0x3c, 0x6e, 0x1c, 0x41, 0x40, 0x85,
+           0xa9, 0x7c, 0xdd, 0x68, 0x7f, 0xa1, 0xe5, 0x0d]
+
+  device := Device.default config
+
+  advertiser := device.advertise
+  advertiser.set_data
+    AdvertisementData
+           --name="Toit heartrate demo"
```
kasperl, line 30: 4 space indent relative to AdvertisementData.

examples/ble/heartrate.toit
```diff
@@ -0,0 +1,53 @@
...
+    simulated_heartrate := 60
+    while true:
+      sleep --ms=500
+      heartrate_send.value= #[0x06, simulated_heartrate]
+      simulated_heartrate += 1
+      if simulated_heartrate == 130: simulated_heartrate = 60
+
+  while true:
+    print
+      "Heart rate app received data $(heartrate_receive.value)"
```
kasperl, line 53: Add a newline at the end.

lib/ble.toit
```diff
@@ -93,13 +100,15 @@ class Advertiser:
   The data is advertised for the given $duration and once every $interval.
     If the $duration is null, then the data is advertised indefinitely.
 
+  The advertise will include the given connection_mode, use one of the BLE_CONNECTION_MODE_* constants
```
kasperl, line 103: Break the long line and indent the continuation with 2 spaces like done in line 101.

lib/ble.toit
```diff
@@ -93,13 +100,15 @@ class Advertiser:
...
+
   Use $set_data to set the data.
 
   Only one advertiser can advertise at any given time.
   */
-  start --duration/Duration?=null --interval/Duration=DEFAULT_INTERVAL:
+  start --duration/Duration?=null --interval/Duration=DEFAULT_INTERVAL --connection_mode/int=BLE_CONNECT_MODE_NONE:
```
kasperl, line 109: Consider putting the long parameters on individual lines. Maybe add a return type in the process.

```
start -> none
    --interval/Duration=DEFAULT_INTERVAL
    --connection_mode/int=BLE_CONNECT_MODE_NONE:
  ...
```

lib/ble.toit
```diff
@@ -237,12 +246,12 @@ class Client:
...
-    ble_connect_ device.resource_group_ address.raw_
-    state := device.resource_state_.wait_for_state CONNECTED_EVENT_ | CONNECTED_EVENT_
-    if state & CONNECT_FAILED_EVENT_ != 0:
-      throw "BLE connection failed"
     gatt_ = ble_get_gatt_ device.resource_group_
     resource_state_ = monitor.ResourceState_ device.resource_group_ gatt_
+    ble_connect_ device.resource_group_ address.raw_ gatt_
+    state := resource_state_.wait_for_state CONNECTED_EVENT_ | CONNECT_FAILED_EVENT_
+    if state & CONNECT_FAILED_EVENT_ != 0:
+      throw "BLE connection failed"
```
kasperl, line 254: Should we dispose the resource state and close the resource group here to clean up nicely?

lib/ble.toit
```diff
@@ -257,23 +266,159 @@ class Client:
...
+    services.add service
+    return Service resource_group_ uuid
+
+
+/** 
+Defines a BLE service with characteristics 
+*/
+class Service:
+  /**
+  The uuid of the service
```
kasperl, line 287: UUID? Terminate comment with .

lib/ble.toit
```diff
@@ -257,23 +266,159 @@ class Client:
...
+  
+  For 16 and 32 bit UUIDs, form the BLE variant with the top-level uuid function
```
kasperl, line 289: Terminate comment with .

lib/ble.toit
```diff
@@ -257,23 +266,159 @@ class Client:
...
+  */
+  uuid/uuid_pkg.Uuid
+  
+  resource_group_/any
+  resource_/any
+
+  characteristics_/Map := {:}
+
+
```
kasperl, line 298: Drop one newline.

lib/ble.toit
```diff
@@ -257,23 +266,159 @@ class Client:
...
+  add_write_only_characteristic uuid -> WriteOnlyCharacteristic:
+    return WriteOnlyCharacteristic this uuid
+
+  add_read_write_characteristic uuid --value=#[]-> ReadWriteCharacteristic:
+    return ReadWriteCharacteristic this uuid value
+
+  add_notification_characteristic uuid -> NotificationCharacteristic:
+    return NotificationCharacteristic this uuid
+
```
kasperl, line 313: Drop one newline.

lib/ble.toit
```diff
@@ -257,23 +266,159 @@ class Client:
...
+BLE_CHR_TYPE_READ_WRITE ::= 3
+BLE_CHR_TYPE_NOTIFICATION ::= 4
+
+BLE_WAIT_RECV ::= 1 << 0
+BLE_WAIT_ACCESSED ::= 1 << 1
+BLE_WAIT_SUBSCRIBED ::= 1 << 2
+
+abstract class Characteristic:
+  /**
+  The uuid of the characteristic
```
kasperl, line 330: UUID? Terminate comment with .

lib/ble.toit
```diff
@@ -257,23 +266,159 @@ class Client:
...
+  */
+  uuid/uuid_pkg.Uuid
+
+  state_/ResourceState_
+
+  constructor service/Service resource .uuid:
+    state_ = ResourceState_ service.resource_group_ resource
+  
+/** 
+A characteristic that can only be read by clients
```
kasperl, line 340: Terminate comment with .

lib/ble.toit
```diff
@@ -257,23 +266,159 @@ class Client:
...
+
+  add_notification_characteristic uuid -> NotificationCharacteristic:
+    return NotificationCharacteristic this uuid
+
+
+  get_characteristic uuid -> Characteristic?:
+    return characteristics_[uuid]
+
+
+BLE_CHR_TYPE_READ_ONLY ::= 1
```
kasperl, line 319: Are these constants supposed to be public or are they only used internally? If they are private, give them an underscore suffix.

lib/ble.toit
```diff
@@ -257,23 +266,159 @@ class Client:
...
+    super service resource uuid
+
+  value -> ByteArray?:
+    state_.wait_for_state BLE_WAIT_RECV 
+    data := ble_get_characteristics_value_ state_.resource
+    state_.clear_state BLE_WAIT_RECV
+    return data;
+
+/**
+A characteristic that allows both read and write by the client
```
kasperl, line 371: Terminate comment with .

lib/ble.toit
```diff
@@ -257,23 +266,159 @@ class Client:
...
+    value_ = value
+
+  value -> ByteArray: return value_
+
+  value= value/ByteArray -> none:
+    ble_set_characteristics_value_ state_.resource value
+    value_ = value
+
+/**
+A characteristic that can only be written to by clients
```
kasperl, line 357: Terminate comment with .

lib/ble.toit
```diff
@@ -257,23 +266,159 @@ class Client:
...
+    data := ble_get_characteristics_value_ state_.resource
+    state_.clear_state BLE_WAIT_RECV
+    return data;
+
+  value= value/ByteArray -> none:
+    ble_set_characteristics_value_ state_.resource value
+
+
+/** 
+A characteristic that the client can subscribe to changes on
```
kasperl, line 389: Complete sentence and terminate with .

lib/ble.toit
```diff
@@ -302,10 +447,13 @@ class Device:
   Initializes an advertiser for the local device.
 
   The returned advertiser is not started.
+
+  If the device has services defined, these are automatically added to the advertised data
```
kasperl, line 451: Terminate comment with .

lib/ble.toit
```diff
@@ -257,23 +266,159 @@ class Client:
...
+    server_config_resource_group := server_config!=null?server_config.resource_group_:null
+    resource_group_ = ble_init_ server_config_resource_group
     add_finalizer this:: this.close
     try:
       gap := ble_gap_ resource_group_
       resource_state := monitor.ResourceState_ resource_group_ gap
-      resource_state.wait_for_state STARTED_EVENT_
+      state := resource_state.wait_for_state STARTED_EVENT_
       resource_state_ = resource_state
+
```
kasperl, line 421: Intentional edit?

lib/ble.toit
```diff
@@ -302,10 +447,13 @@ class Device:
   Initializes an advertiser for the local device.
 
   The returned advertiser is not started.
+
+  If the device has services defined, these are automatically added to the advertised data
   */
   advertise -> Advertiser:
     return Advertiser this
 
+
```
kasperl, line 456: Unintentional edit?

lib/ble.toit
```diff
@@ -343,13 +491,24 @@ class Device:
     finally:
       ble_scan_stop_ resource_group_
 
-STARTED_EVENT_        ::= 1 << 0
-COMPLETED_EVENT_      ::= 1 << 1
-DISCOVERY_EVENT_      ::= 1 << 2
-CONNECTED_EVENT_      ::= 1 << 3
-CONNECT_FAILED_EVENT_ ::= 1 << 4
+  client_connected:
```
kasperl, line 494: The methods name here are bit weird. Maybe `wait_for_client_connected` would be better? At least I find the example hard to understand where you use this method. Are you telling the system that the client has connected or are you waiting for something?

lib/ble.toit
```diff
@@ -343,13 +491,24 @@ class Device:
...
+    resource_state_.wait_for_state CONNECTED_EVENT_
+    resource_state_.clear_state CONNECTED_EVENT_
+    return
```
kasperl, line 497: No need for the return here.

lib/ble.toit
```diff
@@ -343,13 +491,24 @@ class Device:
...
+
+  client_disconnected:
```
kasperl, line 499: Add `-> none` return types for these methods.

src/event_sources/ble_esp32.cc
```diff
@@ -235,6 +235,40 @@ void BLEEventSource::on_gatt_attribute_event(uint16_t conn_handle,
...
 }
 
+int BLEEventSource::on_gatt_server_characteristic(uint16_t conn_handle, uint16_t attr_handle,
+                                                  struct ble_gatt_access_ctxt *ctxt, void *arg) {
+  auto characteristic = reinterpret_cast<BLEServerCharacteristicResource*>(arg);
+  return instance()->on_gatt_server_characteristic_event(ctxt, characteristic);
+}
+
+int BLEEventSource::on_gatt_server_characteristic_event(ble_gatt_access_ctxt *ctxt,
+                                                         BLEServerCharacteristicResource *characteristic) {
```
kasperl, line 245: Indentation is off by one.

src/event_sources/ble_esp32.h
```diff
@@ -51,9 +53,9 @@ class GAPResource : public BLEResource {
 class GATTResource : public BLEResource {
  public:
   TAG(GATTResource);
-  GATTResource(ResourceGroup* group, uint16 handle)
+  GATTResource(ResourceGroup* group)
       : BLEResource(group, GATT)
-      , _handle(handle)
+      , _handle(kInvalidHandle)
```
kasperl, line 58: Put this at the declaration of `_handle` instead?

src/event_sources/ble_esp32.h
```diff
@@ -102,6 +106,92 @@ class GATTResource : public BLEResource {
...
+class BLEServerServiceResource;
+typedef LinkedList<BLEServerServiceResource> BLEServerServiceList;
+
+class BLEServerCharacteristicResource;
+typedef LinkedList<BLEServerCharacteristicResource> BLEServerCharacteristicList;
+
+class BLEServerCharacteristicResource: public Resource, public BLEServerCharacteristicList::Element{
+public:
+  TAG(BLEServerCharacteristicResource);
+  explicit BLEServerCharacteristicResource(ResourceGroup *resource_group, BLEServerServiceResource* service,
```
kasperl, line 119: No need for explicit.

src/event_sources/ble_esp32.h
```diff
@@ -102,6 +106,92 @@ class GATTResource : public BLEResource {
...
+  bool is_indicate_enabled() const { return _indicate; }
+  uint16 conn_handle() const { return _conn_handle; }
+  void set_mbuf_received(os_mbuf *mbuf);
+  os_mbuf *mbuf_received() { Locker locker(_mutex); return _mbuf_received; }
+
+  void set_subscription_status(bool indicate, bool notify, uint16 conn_handle) {
+    _indicate = indicate;
+    _notify = notify;
+    _conn_handle = conn_handle;
+  };
```
kasperl, line 150: Drop ;

src/event_sources/ble_esp32.h
```diff
@@ -102,6 +106,92 @@ class GATTResource : public BLEResource {
...
+
+private:
+  BLEServerServiceResource* _service;
+  ble_uuid_any_t _uuid;
+  int _type;
+  struct os_mbuf *_mbuf_to_send;
+  uint16 _nimble_value_handle;
+  os_mbuf *_mbuf_received;
```
kasperl, line 158: `os_mbuf* _mbuf_received;`

src/event_sources/ble_esp32.h
```diff
@@ -102,6 +106,92 @@ class GATTResource : public BLEResource {
...
+  bool _indicate;
+  bool _notify;
+  uint16 _conn_handle;
+  Mutex *_mutex;
```
kasperl, line 162: `Mutex* _mutex` (this is general and there are more cases)

src/event_sources/ble_esp32.h
```diff
@@ -126,7 +216,11 @@ class BLEEventSource : public LazyEventSource, public Thread {
                                void* arg);
   static void on_started();
 
- protected:
+
```
kasperl, line 219: Drop one newline.

src/event_sources/ble_esp32.h
```diff
@@ -126,7 +216,11 @@ class BLEEventSource : public LazyEventSource, public Thread {
...
+  static int on_gatt_server_characteristic(uint16_t conn_handle, uint16_t attr_handle,
+                                           struct ble_gatt_access_ctxt *ctxt, void *arg);
+
+protected:
```
kasperl, line 223: Indent protected by one space.

src/event_sources/ble_esp32.h
```diff
@@ -151,12 +245,17 @@ class BLEEventSource : public LazyEventSource, public Thread {
...
+                                          BLEServerCharacteristicResource *characteristic);
   void on_started_event();
 
+
   ConditionVariable* _resources_changed = null;
   bool _running = false;
   bool _should_run = false;
   bool _stop = false;
+  static ble_gatt_access_fn *on_access;
+
```
kasperl, line 258: Drop newline.

src/resources/ble_esp32.cc
```diff
@@ -139,53 +156,105 @@ class BLEResourceGroup : public ResourceGroup {
...
+
+  BLEDiscovery *next() {
     Locker locker(_mutex);
     return _discoveries.first();
   }
 
   bool remove_next() {
     Locker locker(_mutex);
-    BLEDiscovery* next = _discoveries.first();
+    BLEDiscovery *next = _discoveries.first();
```
kasperl, line 182: No thanks :)

src/resources/ble_esp32.cc
```diff
@@ -139,53 +156,105 @@ class BLEResourceGroup : public ResourceGroup {
...
     if (!next) return false;
     _discoveries.remove_first();
     delete next;
     return true;
   }
 
- private:
+private:
```
kasperl, line 189: Keep indentation.

src/resources/ble_esp32.cc
```diff
@@ -139,53 +156,105 @@ class BLEResourceGroup : public ResourceGroup {
...
-  bool _scanning = false;
+  GAPResource *_gap = null;
+  BLEServerConfigGroup *_server_config;
 };
 
-uint32_t BLEResourceGroup::on_event(Resource* resource, word data, uint32_t state) {
-  struct ble_gap_event* event = reinterpret_cast<struct ble_gap_event*>(data);
+
+class BLEServerConfigGroup : public ResourceGroup, public Thread {
+public:
```
kasperl, line 199: We use one space indent for all of these. Maybe run through the changes and adapt to that style?

src/resources/ble_esp32.cc
```diff
@@ -139,53 +156,105 @@ class BLEResourceGroup : public ResourceGroup {
...
+  TAG(BLEServerConfigGroup);
+
+  explicit BLEServerConfigGroup(Process *process, EventSource* event_source) :
+      ResourceGroup(process, event_source), Thread("toit.ble"),
+      _gatt_services(null) {}
+
+  ~BLEServerConfigGroup() override {
```
kasperl, line 206: Can this be protected? We need to through `tear_down`, so I don't think this should be exposed.

src/resources/ble_esp32.cc
```diff
@@ -139,53 +156,105 @@ class BLEResourceGroup : public ResourceGroup {
...
+    for (BLEServerServiceResource *service: _services) {
+      delete service;
+    }
+
+    free_gatt_service_structure();
+  }
+
+  void tear_down() override {
+    free_gatt_service_structure();
```
kasperl, line 215: We will always call the destructor too, right? So why do you have this override?

src/resources/ble_esp32.cc
```diff
@@ -253,6 +337,90 @@ static void blecent_on_sync(void) {
...
+int BLEResourceGroup::init_server() {
+  if (_server_config != null) {
+    nimble_port_init();
+
+    ble_svc_gap_init();
+    ble_svc_gatt_init();
+
+    // Build the service structure
+    int service_cnt = 0;
+    for (__attribute__((unused)) BLEServerServiceResource *t: _server_config->services()) service_cnt++;
```
kasperl, line 353: I'd rather not have attributes in here. Maybe just use the `USE` macro in the loop body?

src/resources/ble_esp32.cc
```diff
@@ -253,6 +337,90 @@ static void blecent_on_sync(void) {
...
+}
+
+int BLEResourceGroup::init_server() {
+  if (_server_config != null) {
+    nimble_port_init();
+
+    ble_svc_gap_init();
+    ble_svc_gatt_init();
+
+    // Build the service structure
```
kasperl, line 351: Terminate comment with .

kasperl (commented):
Very impressive piece of work. I'm not done reviewing this yet as it is a huge change with lots of resources, primitives, etc. 

Here is my first round of somewhat superficial comments.

src/resources/ble_esp32.cc
mikkeldamsgaard: I noticed that not many attributes in the source, I missed this when I cleaned it up. I will fix

src/resources/ble_esp32.cc
mikkeldamsgaard: Removed the overridden tear_down, as in fact all its work is done in the destructor

lib/ble.toit
mikkeldamsgaard: I could not decide when I wrote them, if I wanted these exposed and in contrast reduce the number of add_*_characteristic  methods. I did in the end turn the constants in effect internal and will update with the undescore

lib/ble.toit
mikkeldamsgaard: Actually, this is no longer true, as that lead to some rather convoluted code in the Advertiser class and the intention from the client is not always clear. So I removed the functionality, but not the comment.

lib/ble.toit
mikkeldamsgaard: Yes. At some point I had a lot of extra code here to try catch some state that happened not to be important. So git kind of never reverted back to the original.

lib/ble.toit
mikkeldamsgaard: yes

lib/ble.toit
mikkeldamsgaard: changed the method names to the suggested wait_for_ variants.

src/event_sources/ble_esp32.h
mikkeldamsgaard: I have tries to change all occurences of type *var to type* var and <type *> to <type*>. To make it consistant across the whole cc file, i also changed some of the existing occurences

src/resources/ble_esp32.cc
mikkeldamsgaard: Ups.

lib/ble.toit
mikkeldamsgaard: Yes. Please advice how that should be done.

kasperl:
This is now at the top of my priority list, @mikkeldamsgaard. I will review and provide feedback soon, so we can continue to make progress on this!

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
     state := resource_state_.wait_for_state STARTED_EVENT_
     resource_state_.clear_state STARTED_EVENT_
 
+class ServerConfig:
+  resource_group_ := null
+  services/List := []
+
+  constructor:
+    resource_group_ = ble_server_config_init_
```
kasperl, line 278: Should this be disposed when ServerConfig goes away? It looks like a leak because the resource group will not go away before the current process dies, unless we add code to clean it up. It looks like we will need to keep it around as long as there are registered services, so maybe we should let the Service object hold a pointer to the ServerConfig it came from instead of just holding on to the resource_group. That way, we can add finalization behaviour to ServerConfig that cleans this up.

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+*/
+class Service:
+  /**
+  The UUID of the service.
+  
+  For 16 and 32 bit UUIDs, form the BLE variant with the top-level uuid function.
+  */
+  uuid/uuid_pkg.Uuid
+  
+  resource_group_/any
```
kasperl, line 297: The type of these should probably be `ByteArray?` because they are proxies for C++ objects.

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+  resource_/any
+
+  characteristics_/Map := {:}
```
kasperl, line 300: `::=`

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+
+  constructor .resource_group_ .uuid:
+    resource_ = ble_add_server_service_ resource_group_ uuid.to_byte_array
```
kasperl, line 303: Looks like we will be leaking the resource as long as the resource group is around. In this case, that might not be so bad because it sticks to the ServerConfig that doesn't allow unregistering.

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+    return WriteOnlyCharacteristic this uuid
+
+  add_read_write_characteristic uuid --value=#[]-> ReadWriteCharacteristic:
+    return ReadWriteCharacteristic this uuid value
+
+  add_notification_characteristic uuid -> NotificationCharacteristic:
+    return NotificationCharacteristic this uuid
+
+  get_characteristic uuid -> Characteristic?:
+    return characteristics_[uuid]
```
kasperl, line 318: Is the intention that this will return null if the characteristic isn't found? You should use `characteristics_.get uuid` instead then, because the `[]` will throw if the `uuid` isn't in the map.

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+  resource_group_/any
+  resource_/any
+
+  characteristics_/Map := {:}
+
+  constructor .resource_group_ .uuid:
+    resource_ = ble_add_server_service_ resource_group_ uuid.to_byte_array
+
+  add_read_only_characteristic uuid --value=#[]-> ReadOnlyCharacteristic:
+    return ReadOnlyCharacteristic this uuid value
```
kasperl, line 306: Should these be put in the `characteristics_` map? Not doing that looks like a bug to me.

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+  constructor service/Service resource .uuid:
+    state_ = ResourceState_ service.resource_group_ resource
+  
+/** 
+A characteristic that can only be read by clients.
+*/
+class ReadOnlyCharacteristic extends Characteristic:
+  value_/ByteArray := #[]
+
+  constructor service/Service uuid/uuid_pkg.Uuid value:
```
kasperl, line 347: Add a type for `value`.

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
     state := resource_state_.wait_for_state STARTED_EVENT_
     resource_state_.clear_state STARTED_EVENT_
 
+class ServerConfig:
```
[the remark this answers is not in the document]
kasperl: Also, I would consider calling this ServerConfiguration just because all the other names in there are so long and spelled out.

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+abstract class Characteristic:
+  /**
+  The UUID of the characteristic.
+  */
+  uuid/uuid_pkg.Uuid
+
+  state_/ResourceState_
+
+  constructor service/Service resource .uuid:
+    state_ = ResourceState_ service.resource_group_ resource
```
kasperl, line 339: Is this where you would want to register the `Characteristic` in the `service.characteristics_` map?

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+
+/**
+A characteristic that can only be written to by clients.
+*/
+class WriteOnlyCharacteristic extends Characteristic:
+  constructor service/Service uuid/uuid_pkg.Uuid:
+    resource := ble_add_server_characteristic_ service.resource_ uuid.to_byte_array BLE_CHR_TYPE_WRITE_ONLY_ null
+    super service resource uuid
+
+  value -> ByteArray?:
```
kasperl, line 366: It looks odd that you can only read from the `WriteOnlyCharacteristics`.

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+*/
+class ReadOnlyCharacteristic extends Characteristic:
+  value_/ByteArray := #[]
+
+  constructor service/Service uuid/uuid_pkg.Uuid value:  
+    resource := ble_add_server_characteristic_ service.resource_ uuid.to_byte_array BLE_CHR_TYPE_READ_ONLY_ value
+    super service resource uuid
+    value_ = value
+
+  value -> ByteArray: return value_
```
kasperl, line 352: Not sure why `ReadOnlyCharacteristics` have methods for reading and writing.

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+class ReadWriteCharacteristic extends Characteristic:
+  constructor service/Service uuid/uuid_pkg.Uuid value:  
+    resource := ble_add_server_characteristic_ service.resource_ uuid.to_byte_array BLE_CHR_TYPE_READ_WRITE_ value
+    super service resource uuid
+
+  value -> ByteArray?:
+    state_.wait_for_state BLE_WAIT_RECV_
+    data := ble_get_characteristics_value_ state_.resource
+    state_.clear_state BLE_WAIT_RECV_
+    return data;
```
kasperl, line 384: Remove ;

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+
+/**
+A characteristic that allows both read and write by the client.
+*/
+class ReadWriteCharacteristic extends Characteristic:
+  constructor service/Service uuid/uuid_pkg.Uuid value:  
+    resource := ble_add_server_characteristic_ service.resource_ uuid.to_byte_array BLE_CHR_TYPE_READ_WRITE_ value
+    super service resource uuid
+
+  value -> ByteArray?:
```
kasperl, line 380: Can this code be shared somehow?

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+    state_.wait_for_state BLE_WAIT_RECV_
+    data := ble_get_characteristics_value_ state_.resource
+    state_.clear_state BLE_WAIT_RECV_
+    return data;
+
+/**
+A characteristic that allows both read and write by the client.
+*/
+class ReadWriteCharacteristic extends Characteristic:
+  constructor service/Service uuid/uuid_pkg.Uuid value:
```
kasperl, line 376: Add a type for `value`.

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+class WriteOnlyCharacteristic extends Characteristic:
+  constructor service/Service uuid/uuid_pkg.Uuid:
+    resource := ble_add_server_characteristic_ service.resource_ uuid.to_byte_array BLE_CHR_TYPE_WRITE_ONLY_ null
+    super service resource uuid
+
+  value -> ByteArray?:
+    state_.wait_for_state BLE_WAIT_RECV_
+    data := ble_get_characteristics_value_ state_.resource
+    state_.clear_state BLE_WAIT_RECV_
+    return data;
```
kasperl, line 370: Remove ;

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
 class Device:
   resource_group_ := ?
   resource_state_/monitor.ResourceState_? := null
 
-  constructor.default:
-    resource_group_ = ble_init_
+  server_config/ServerConfig?
+
+  constructor.default .server_config/ServerConfig?=null:
+    server_config_resource_group := server_config!=null?server_config.resource_group_:null
```
kasperl, line 415: Add some spacing here.

```
  server_config_resource_group := server_config != null
      ? server_config.resource_group_
      : null
```

lib/ble.toit
```diff
@@ -257,23 +270,157 @@ class Client:
...
+
+If services is not empty, sets up the services for the advertiser. 
 */
 class Device:
   resource_group_ := ?
   resource_state_/monitor.ResourceState_? := null
 
-  constructor.default:
-    resource_group_ = ble_init_
+  server_config/ServerConfig?
```
kasperl, line 412: Maybe this could just be called `server`? It looks like you're expecting folks to write code that manipulates this after the device has been created (like `device.server_config.add_xxx`), so I guess this is intentionally public?

src/event_sources/ble_esp32.cc
```diff
@@ -235,6 +235,40 @@ void BLEEventSource::on_gatt_attribute_event(uint16_t conn_handle,
...
+                                                  struct ble_gatt_access_ctxt* ctxt, void* arg) {
+  auto characteristic = reinterpret_cast<BLEServerCharacteristicResource*>(arg);
+  return instance()->on_gatt_server_characteristic_event(ctxt, characteristic);
+}
+
+int BLEEventSource::on_gatt_server_characteristic_event(ble_gatt_access_ctxt* ctxt,
+                                                        BLEServerCharacteristicResource* characteristic) {
+  int err = 0;
+  switch (ctxt->op) {
+    case  BLE_GATT_ACCESS_OP_READ_CHR:
```
kasperl, line 248: Drop one of the spaces after case.

src/event_sources/ble_esp32.cc
```diff
@@ -235,6 +235,40 @@ void BLEEventSource::on_gatt_attribute_event(uint16_t conn_handle,
...
+        err = os_mbuf_appendfrom(ctxt->om, characteristic->mbuf_to_send(), 0, characteristic->mbuf_to_send()->om_len);
+      }
+      break;
+    case BLE_GATT_ACCESS_OP_WRITE_CHR:
+      characteristic->set_mbuf_received(ctxt->om);
+      ctxt->om = null;
+      break;
+    case BLE_GATT_ACCESS_OP_READ_DSC:
+    case BLE_GATT_ACCESS_OP_WRITE_DSC:
+      // ignore
```
kasperl, line 259: Maybe improve the comment here. We ignore this and don't dispatch it.

src/event_sources/ble_esp32.cc
```diff
@@ -235,6 +235,40 @@ void BLEEventSource::on_gatt_attribute_event(uint16_t conn_handle,
...
+      return 0;
+    default:
+      // unknown
```
kasperl, line 262: Improve comment. What do we do with unknown stuff? It looks like we go on and dispatch, without taking any further action here.

src/event_sources/ble_esp32.cc
```diff
@@ -235,6 +235,40 @@ void BLEEventSource::on_gatt_attribute_event(uint16_t conn_handle,
...
+      break;
+  }
+
+  Locker locker(mutex());
```
kasperl, line 266: Is it intentional that you dispatch even when there is an error? Maybe. Add a comment to explain why.

src/event_sources/ble_esp32.h
```diff
@@ -95,13 +98,99 @@ class GATTResource : public BLEResource {
...
+    _indicate = indicate;
+    _notify = notify;
+    _conn_handle = conn_handle;
+  }
+
+ private:
+  BLEServerServiceResource* _service;
+  ble_uuid_any_t _uuid;
+  int _type;
+  struct os_mbuf* _mbuf_to_send;
```
kasperl, line 155: Drop struct?

src/event_sources/ble_esp32.h
```diff
@@ -95,13 +98,99 @@ class GATTResource : public BLEResource {
...
 };
 
+
+class BLEServerServiceResource;
+typedef LinkedList<BLEServerServiceResource> BLEServerServiceList;
+
+class BLEServerCharacteristicResource;
+typedef LinkedList<BLEServerCharacteristicResource> BLEServerCharacteristicList;
+
+class BLEServerCharacteristicResource: public Resource, public BLEServerCharacteristicList::Element{
```
kasperl, line 115: Add space between Element and {.

src/event_sources/ble_esp32.h
```diff
@@ -95,13 +98,99 @@ class GATTResource : public BLEResource {
...
+ public:
+  TAG(BLEServerCharacteristicResource);
+  BLEServerCharacteristicResource(ResourceGroup* resource_group, BLEServerServiceResource* service,
```
kasperl, line 118: This appears leaky. What happens when the resource group is torn down and deletes this resource? Do you just leak the mbufs, etc.?

src/event_sources/ble_esp32.h
```diff
@@ -95,13 +98,99 @@ class GATTResource : public BLEResource {
...
+      _mbuf_received(null),
+      _indicate(false),
+      _notify(false),
+      _conn_handle(0),
+      _mutex(OS::allocate_mutex(3, "")) {}
+
+  ble_uuid_any_t uuid() const { return _uuid; }
+  ble_uuid_t* uuid_p() { return &_uuid.u; }
+  int type() const { return _type; }
+  struct os_mbuf* mbuf_to_send() { Locker locker(_mutex); return _mbuf_to_send; };
```
kasperl, line 135: Why is this struct necessary?

src/event_sources/ble_esp32.h
```diff
@@ -95,13 +98,99 @@ class GATTResource : public BLEResource {
...
+      _mbuf_to_send(value),
+      _nimble_value_handle(0),
+      _mbuf_received(null),
+      _indicate(false),
+      _notify(false),
+      _conn_handle(0),
+      _mutex(OS::allocate_mutex(3, "")) {}
+
+  ble_uuid_any_t uuid() const { return _uuid; }
+  ble_uuid_t* uuid_p() { return &_uuid.u; }
```
kasperl, line 133: What does the _p suffix indicate?

src/event_sources/ble_esp32.h
kasperl: I would take the methods that are accessed from other threads and group them. Looks like it is the mbuf operations. I wouldn't have any uses of Locker in inline method definitions so move the bodies to the .cc file.

src/event_sources/ble_esp32.h
```diff
@@ -95,13 +98,99 @@ class GATTResource : public BLEResource {
...
+  }
+
+  ble_uuid_any_t uuid() const { return _uuid; }
+  ble_uuid_t* uuid_p() { return &_uuid.u; }
+  BLEServerCharacteristicList characteristics() const { return _characteristics; }
+
+ private:
+  BLEServerCharacteristicList _characteristics;
+  ble_uuid_any_t _uuid;
+
```
kasperl, line 189: Drop newline.

src/event_sources/ble_esp32.h
```diff
@@ -95,13 +98,99 @@ class GATTResource : public BLEResource {
...
+  BLEServerServiceResource* _service;
+  ble_uuid_any_t _uuid;
+  int _type;
+  struct os_mbuf* _mbuf_to_send;
+  uint16 _nimble_value_handle;
+  os_mbuf* _mbuf_received;
+  bool _indicate;
+  bool _notify;
+  uint16 _conn_handle;
+  Mutex* _mutex;
```
kasperl, line 161: Do they each need a mutex? Sounds expensive.

src/resources/ble_esp32.cc
```diff
@@ -76,7 +90,7 @@ class BLEDiscovery : public DiscoveriesFIFO::Element {
     free(_data);
   }
 
-  bool init(struct ble_gap_disc_desc& disc) {
+  bool init(struct ble_gap_disc_desc &disc) {
```
kasperl, line 93: Move & back again for consistency.

src/resources/ble_esp32.cc
```diff
@@ -93,8 +107,11 @@ class BLEDiscovery : public DiscoveriesFIFO::Element {
   }
 
   ble_addr_t addr() { return _addr; }
+
```
kasperl, line 110: Unintentional edits?

src/resources/ble_esp32.cc
```diff
@@ -104,13 +121,17 @@ class BLEDiscovery : public DiscoveriesFIFO::Element {
...
 class BLEResourceGroup : public ResourceGroup {
  public:
   TAG(BLEResourceGroup);
+
   BLEResourceGroup(Process* process, BLEEventSource* event_source, int id)
-      : ResourceGroup(process, event_source)
-      , _id(id)
-      , _mutex(OS::allocate_mutex(3, "")) {
+      : ResourceGroup(process, event_source),
```
kasperl, line 131: Go back to the old formatting for consistency (: and ,).

src/resources/ble_esp32.cc
```diff
@@ -174,16 +176,80 @@ class BLEResourceGroup : public ResourceGroup {
...
+  uint32_t on_event(Resource* resource, word data, uint32_t state) override;
+
+  void entry() override;
+
+  void set_subscription_status(uint16 attr_handle, uint16 conn_handle, bool indicate, bool notify);
+
+  void set_gatt_service_structure(ble_gatt_svc_def* gatt_services) {
+    if (_gatt_services != null) free_gatt_service_structure();
+    _gatt_services = gatt_services;
+  };
```
kasperl, line 233: Remove ;

src/resources/ble_esp32.cc
```diff
@@ -218,33 +284,48 @@ uint32_t BLEResourceGroup::on_event(Resource* resource, word data, uint32_t stat
...
+      break;
   }
 
   return state;
 }
 
-static void blecent_on_sync(void) {
-    /* Make sure we have proper identity address set (public preferred) */
+static void ble_on_sync(void) {
+  /* Make sure we have proper identity address set (public preferred) */
```
kasperl, line 328: Use //.

src/resources/ble_esp32.cc
```diff
@@ -469,27 +689,30 @@ PRIMITIVE(scan_next) {
...
 
-  if (group->is_advertising()) ALREADY_EXISTS;
+  if (BLEResourceGroup::is_advertising()) ALREADY_EXISTS;
 
   int32 duration_ms = duration_us < 0 ? BLE_HS_FOREVER : duration_us / 1000;
 
   struct ble_gap_adv_params adv_params = { 0 };
-  // No support for connections yet.
-  adv_params.conn_mode = BLE_GAP_CONN_MODE_NON;
+  if (group->server_config() != null) {
```
kasperl, line 710: This looks weird. You check the server_config and do the same thing if it is null or not.

src/resources/ble_esp32.cc
```diff
@@ -647,10 +851,27 @@ PRIMITIVE(request_data) {
...
+  if (gatt->error() == BLE_HS_ENOENT) {
+    OUT_OF_RANGE;
+  } else if (gatt->error() != 0) {
+    INVALID_ARGUMENT;
+  }
+
+  const struct os_mbuf* mbuf = gatt->mbuf();
+  if (!mbuf) return process->program()->null_object();
+  Object* ret_val = convert_mbuf_to_heap_object(process, mbuf);
+  gatt->set_mbuf(null);
```
kasperl, line 870: Do you want to do this even if the allocation of the heap object failed? That looks wrong.

src/resources/ble_esp32.cc
```diff
@@ -697,6 +918,112 @@ PRIMITIVE(request_attribute) {
...
+  proxy->set_external_address(group);
+  return proxy;
+}
+
+PRIMITIVE(add_server_service) {
+  ARGS(BLEServerConfigGroup, group, Blob, uuid_blob);
+  ByteArray* proxy = process->object_heap()->allocate_proxy();
+  if (!proxy) ALLOCATION_FAILED;
+
+  ble_uuid_any_t uuid = uuid_from_blob(uuid_blob);
```
kasperl, line 949: This doesn't allocate in the malloc heap, right?

src/resources/ble_esp32.cc
```diff
@@ -697,6 +918,112 @@ PRIMITIVE(request_attribute) {
...
+}
+
+PRIMITIVE(add_server_characteristic) {
+  ARGS(BLEServerServiceResource, service, Blob, uuid_blob, int, type, Object, value);
+  ByteArray* proxy = process->object_heap()->allocate_proxy();
+  if (!proxy) ALLOCATION_FAILED;
+
+  ble_uuid_any_t uuid = uuid_from_blob(uuid_blob);
+
+  struct os_mbuf* om =null;
```
kasperl, line 965: Space before null.

src/resources/ble_esp32.cc
```diff
@@ -697,6 +918,112 @@ PRIMITIVE(request_attribute) {
...
+  if (!object_to_mbuf(process, value, &om)) WRONG_TYPE;
+
+  BLEServerCharacteristicResource* characteristic = service->add_characteristic(uuid, type, om);
+  if (!characteristic) MALLOC_FAILED;
```
kasperl, line 969: Are you leaking the mbuf here? I think so.

src/resources/ble_esp32.cc
```diff
@@ -697,6 +918,112 @@ PRIMITIVE(request_attribute) {
...
+    if (resource->is_notify_enabled()) {
+      err = ble_gattc_notify_custom(resource->conn_handle(), resource->nimble_value_handle(), om);
+    }
+
+    if (err == 0 && resource->is_indicate_enabled()) {
+      err = ble_gattc_indicate_custom(resource->conn_handle(), resource->nimble_value_handle(), om);
+    }
+
+    if (err != ESP_OK) {
+      return Primitive::os_error(err, process);
```
kasperl, line 1003: Are you leaking the mbuf here? I think so.

src/resources/ble_esp32.cc
```diff
@@ -697,6 +918,112 @@ PRIMITIVE(request_attribute) {
...
+}
+
+PRIMITIVE(notify_characteristics_value) {
+  ARGS(BLEServerCharacteristicResource, resource, Object, value);
+
+  if (resource->is_notify_enabled() || resource->is_indicate_enabled()) {
+    struct os_mbuf* om;
+    if (!object_to_mbuf(process, value, &om)) WRONG_TYPE;
+
+    int err = 0;
```
kasperl, line 993: Use ESP_OK?

src/resources/ble_esp32.cc
```diff
@@ -697,6 +918,112 @@ PRIMITIVE(request_attribute) {
...
+    if (resource->is_notify_enabled()) {
+      err = ble_gattc_notify_custom(resource->conn_handle(), resource->nimble_value_handle(), om);
+    }
+
+    if (err == 0 && resource->is_indicate_enabled()) {
```
kasperl, line 998: Use ESP_OK?

src/resources/ble_esp32.cc
```diff
@@ -697,6 +918,112 @@ PRIMITIVE(request_attribute) {
...
+      return Primitive::os_error(err, process);
+    }
+  }
+
+  if (resource->is_notify_enabled())
+    return process->program()->true_object();
+  else
+    return process->program()->false_object();
+
+  return process->program()->null_object();
```
kasperl, line 1012: Drop this unreachable code.

src/resources/ble_esp32.cc
```diff
@@ -697,6 +918,112 @@ PRIMITIVE(request_attribute) {
...
+      err = ble_gattc_indicate_custom(resource->conn_handle(), resource->nimble_value_handle(), om);
+    }
+
+    if (err != ESP_OK) {
+      return Primitive::os_error(err, process);
+    }
+  }
+
+  if (resource->is_notify_enabled())
+    return process->program()->true_object();
```
kasperl, line 1008: Always use { } for multiline if constructs.

src/resources/ble_esp32.cc
```diff
@@ -697,6 +918,112 @@ PRIMITIVE(request_attribute) {
...
+    if (err == 0 && resource->is_indicate_enabled()) {
+      err = ble_gattc_indicate_custom(resource->conn_handle(), resource->nimble_value_handle(), om);
+    }
+
+    if (err != ESP_OK) {
+      return Primitive::os_error(err, process);
+    }
+  }
+
+  if (resource->is_notify_enabled())
```
kasperl, line 1007: Replace this with:

```
return BOOL(resource->is_notify_enabled());
```

src/resources/ble_esp32.cc
```diff
@@ -697,6 +918,112 @@ PRIMITIVE(request_attribute) {
...
+
+PRIMITIVE(get_characteristics_value) {
+  ARGS(BLEServerCharacteristicResource, resource);
+
+  os_mbuf* mbuf = resource->mbuf_received();
+  if (mbuf == null) return process->program()->null_object();
+
+  Object* ret_val = convert_mbuf_to_heap_object(process, mbuf);
+
+  resource->set_mbuf_received(null);
```
kasperl, line 1023: What happens if the allocation in convert_mbuf_to_heap_object fails? I don't think you want to clear the mbuf_received in that case.

kasperl (commented):
More comments. I think there a bunch of resource leaks and allocation failure scenarios that aren't handled properly, but I think we're getting close!

src/resources/ble_esp32.cc
mikkeldamsgaard: It is a bit more complicated. The nimble api does not use the ESP error codes, but its own set. So even though ESO_OK is 0, it would not really be correct to use that constant here. The nimble stack does not have a corresponding BLE_OK, so i kept it as 0.

src/resources/ble_esp32.cc
mikkeldamsgaard: Why not? If the allocation fails, the memory is running out. if the mbuf is not cleared, it will potentially grow with further calls pressing the memory even more.

lib/ble.toit
mikkeldamsgaard: I know. But the naming is coming from the BLE specs, a write only characteristic is a characteristic that clients can only write to. So on the server side, it can correspondingly only be read from.

lib/ble.toit
mikkeldamsgaard: The naming of ReadOnly is for a client to only read from the characteristic. To set the value that the client reads, the user of the library will set the value. To read what the client will read, the user of the library can use the read function.

lib/ble.toit
mikkeldamsgaard: Yes. Services will be kept around as long as the ServerConfiguration is around. There is no way to deregister a single service in nimble AFAIK.

src/resources/ble_esp32.cc
mikkeldamsgaard: As an allocation failure potentially can retry the primitive, the code has been updated to not clear the mbuf.

kasperl:
Let me know when you feel like this is ready for another (final?) review, @mikkeldamsgaard!

mikkeldamsgaard:
> 

I do not have any pending changes, so you can go ahead

kasperl:
There are still a bunch of unresolved comments for ble_esp32.cc?

lib/ble.toit
mikkeldamsgaard: Removed the field. The server config is freezed on ble create time

src/resources/ble_esp32.cc
mikkeldamsgaard: I do not understand the comment.

mikkeldamsgaard:
> There are still a bunch of unresolved comments for ble_esp32.cc?

Found the comments. Pushed more changes

src/event_sources/ble_esp32.cc
```diff
@@ -246,6 +280,39 @@ void BLEEventSource::on_started() {
...
+    _mbuf_received = mbuf;
+  } else if (mbuf == null) {
+    os_mbuf_free_chain(_mbuf_received);
+    _mbuf_received = null;
+  } else {
+    os_mbuf_concat(_mbuf_received, mbuf);
+  }
+}
+
+os_mbuf *BLEServerCharacteristicResource::mbuf_received() {
```
kasperl, line 295: `os_mbuf *` -> `os_mbuf*`

examples/ble/heart_rate.toit
```diff
@@ -0,0 +1,56 @@
+// Copyright (C) 2021 Toitware ApS.
+// Use of this source code is governed by a Zero-Clause BSD license that can
+// be found in the examples/LICENSE file.
+
+// The standard BLE peripheral demo for simulating a heart rate monitor.
+
+import ble show *
+import uuid show Uuid
+
+GATT_IO_UUID  ::= 0x1825
```
kasperl, line 10: Drop one of the spaces.

lib/ble.toit
```diff
@@ -257,23 +270,174 @@ class Client:
...
+  constructor:
+    resource_group_ = ble_server_configuration_init_
+    add_finalizer this:: ble_server_configuration_dispose_ resource_group_
+
+  add_service uuid -> Service:
+    service := Service this uuid
+    services.add service
+    return service
+
+  finalization:
```
kasperl, line 289: Remove this again?

src/event_sources/ble_esp32.cc
```diff
@@ -246,6 +280,39 @@ void BLEEventSource::on_started() {
...
+  return _mbuf_received;
+}
+
+void BLEServerCharacteristicResource::set_mbuf_to_send(os_mbuf* mbuf) {
+  Locker locker(_mutex);
+  if (_mbuf_to_send != null) os_mbuf_free(_mbuf_to_send);
+  _mbuf_to_send = mbuf;
+}
+
+os_mbuf *BLEServerCharacteristicResource::mbuf_to_send() {
```
kasperl, line 306: `os_mbuf *` -> `os_mbuf*`

src/event_sources/ble_esp32.h
```diff
@@ -51,9 +51,8 @@ class GAPResource : public BLEResource {
 class GATTResource : public BLEResource {
  public:
   TAG(GATTResource);
-  GATTResource(ResourceGroup* group, uint16 handle)
+  GATTResource(ResourceGroup* group)
```
kasperl, line 54: Make explicit.

src/resources/ble_esp32.cc
```diff
@@ -174,16 +170,86 @@ class BLEResourceGroup : public ResourceGroup {
     return true;
   }
 
+ protected:
+  ~BLEResourceGroup() override {
```
kasperl, line 174: I would probably move this out of the class definition.

src/event_sources/ble_esp32.h
```diff
@@ -95,13 +96,101 @@ class GATTResource : public BLEResource {
...
+typedef LinkedList<BLEServerServiceResource> BLEServerServiceList;
+
+class BLEServerCharacteristicResource;
+typedef LinkedList<BLEServerCharacteristicResource> BLEServerCharacteristicList;
+
+class BLEServerCharacteristicResource: public Resource, public BLEServerCharacteristicList::Element {
+ public:
+  TAG(BLEServerCharacteristicResource);
+  BLEServerCharacteristicResource(ResourceGroup* resource_group, BLEServerServiceResource* service,
+                                  ble_uuid_any_t uuid, int type, os_mbuf* value, Mutex* mutex):
```
kasperl, line 117: Reformat the initialization list to look like the others in the code base.

src/resources/ble_esp32.cc
```diff
@@ -253,6 +334,96 @@ static void blecent_on_sync(void) {
...
+    }
+
+    rc = ble_gatts_add_svcs(gatt_services);
+    if (rc != 0) {
+      _server_config->tear_down();
+      return rc;
+    }
+
+    ble_hs_cfg.reset_cb = ble_on_reset;
+    // Start the host thread
```
kasperl, line 417: Terminate comment with .

src/event_sources/ble_esp32.cc
```diff
@@ -235,6 +235,40 @@ void BLEEventSource::on_gatt_attribute_event(uint16_t conn_handle,
...
+        err = os_mbuf_appendfrom(ctxt->om, characteristic->mbuf_to_send(), 0, characteristic->mbuf_to_send()->om_len);
+      }
+      break;
+    case BLE_GATT_ACCESS_OP_WRITE_CHR:
+      characteristic->set_mbuf_received(ctxt->om);
+      ctxt->om = null;
+      break;
+    case BLE_GATT_ACCESS_OP_READ_DSC:
+    case BLE_GATT_ACCESS_OP_WRITE_DSC:
+      // Not currently implemented. Ignoring and not dispatching
```
kasperl, line 259: Terminate comment with .

src/event_sources/ble_esp32.cc
```diff
@@ -235,6 +235,40 @@ void BLEEventSource::on_gatt_attribute_event(uint16_t conn_handle,
...
+      return 0;
+    default:
+      // Unhandled event, no dispatching
```
kasperl, line 262: Terminate comment with .

kasperl (approved):
Looks good to me. Great work!

kasperl:
Closing this PR. All commits landed as part of #522.

PR #411, opened 2022-02-03
Remove C-level DNS resolution.

erikcorry:
Unfortunately it doesn't appear that esp-idf is able to
compile out DNS support.


kasperl (approved):
Looks good to me.

PR #412, opened 2022-02-03
Experiment: Can we get rid of Mac LSP hangs?

erikcorry:
Don't try to read more from sockets and pipes than
we are told is available.  But do try to read one
byte when we are told 0 are available to trigger the
EWOULDBLOCK error.


floitsch (approved):
LGTM.

PR #416, opened 2022-02-04
Fix bug in RPC cancelation

kasperl:


lib/rpc/broker.toit
```diff
@@ -155,7 +168,8 @@ monitor RpcRequestQueue_:
         if not next:
           last_ = previous
         unprocessed_--
-      previous = current
+      else:
```
kasperl, line 171: This is the bug now exposed by tests.

lask (approved):
Good catch!

PR #418, opened 2022-02-04
Faster == on ByteArrays

erikcorry:


lib/core/collections.toit
```diff
@@ -1262,8 +1262,10 @@ abstract class ByteArrayBase_ implements ByteArray:
     if other is not ByteArray: return false
     s := size
     if s != other.size: return false
-    s.repeat: if this[it] != other[it]: return false
-    return true
+    return compare_ other
+
+  compare_ other -> bool:
```
kasperl, line 1267: Add a type for other.

kasperl: This works for strings too, no?

kasperl (approved):
Looks good to me, but can we use this for strings?

kasperl (approved):
Looks even better to me.

PR #421, opened 2022-02-06
Documentation on what a flash-store value can be.

### LINKED ISSUE #420: Saving the map-object in the flash store
A small bug related to saving the map-object in the flash store. If you use string as a key, then there are no problems, but if int, then crash occurs:

```
Class 'int' does not have any method 'size'.
  0: Encoder.encode_string_inner_ <sdk>/encoding/ubjson.toit:94:21
  1: Encoder.encode_map_.<block> <sdk>/encoding/ubjson.toit:63:7
  2: HashedInsertionOrderedCollection_.hash_do_ <sdk>/core/collections.toit:2098:3
  3: Map.do                    <sdk>/core/collections.toit:2628:10
  4: Encoder.encode_map_       <sdk>/encoding/ubjson.toit:62:9
  5: Encoder.encode_           <sdk>/encoding/ubjson.toit:46:25
  6: Encoder.encode            <sdk>/encoding/ubjson.toit:34:7
  7: encode                    <sdk>/encoding/ubjson.toit:14:5
  8: serialize                 <sdk>/serialization.toit:35:16
  9: FlashStore.set            <sdk>/device.toit:85:48
 10: saveData                  /home/micrcx/toit_apps/bytes/hsmfb_transport.toit:50:9
 11: receiveMessage.<block>    /home/micrcx/toit_apps/bytes/hsmfb_transport.toit:145:5
 12: subscribe_.<block>        <sdk>/pubsub.toit:373:15
 13: subscribe_                <sdk>/pubsub.toit:364:1
 14: subscribe                 <sdk>/pubsub.toit:358:3
 15: receiveMessage            /home/micrcx/toit_apps/bytes/hsmfb_transport.toit:125:9
 16: main                      /home/micrcx/toit_apps/bytes/hsmfb_transport.toit:199:3
 17: __entry__.<lambda>        <sdk>/core/entry.toit:48:20

```
I.e. code:

```
map := {:}
map["1"] = "Ok"
store := device.FlashStore
store.set key object
```
performs without problems, and the code:

```
map := {:}
map[1] = "Bug"
store := device.FlashStore
store.set key object
```
causes a crash.
In the sources of the library, I found that only text keys are supported during serialization, so perhaps the bug is not critical. It can be easily bypassed, for example:
```
key := 32
map["$key"] = "Ok",

```
but for those who do not know this, the bug can be an unpleasant surprise.

floitsch:
Values are encoded as UBJSON.

Fixes #420.

### DIFF
--- a/lib/device.toit
+++ b/lib/device.toit
@@ -70,6 +70,11 @@ class FlashStore implements Store:
   Inserts the given $key-$value pair in the store.
 
   If the $key already exists in the store, then the value is overwritten.
+
+  The $value is encoded as UBJSON. As such it supports:
+  - literals: numbers, booleans, strings, null.
+  - lists.
+  - maps. The keys must be strings, and the values must be valid JSON objects.
   */
   set key/string value/any:
     impl.FlashStore_.instance.set key value
--- a/lib/device.toit
+++ b/lib/device.toit
@@ -74,7 +74,7 @@ class FlashStore implements Store:
   The $value is encoded as UBJSON. As such it supports:
   - literals: numbers, booleans, strings, null.
   - lists.
-  - maps. The keys must be strings, and the values must be valid JSON objects.
+  - maps. The keys must be strings, and the values must be valid UBJSON objects.
   */
   set key/string value/any:
     impl.FlashStore_.instance.set key value


lib/device.toit
```diff
@@ -70,6 +70,11 @@ class FlashStore implements Store:
   Inserts the given $key-$value pair in the store.
 
   If the $key already exists in the store, then the value is overwritten.
+
+  The $value is encoded as UBJSON. As such it supports:
+  - literals: numbers, booleans, strings, null.
+  - lists.
+  - maps. The keys must be strings, and the values must be valid JSON objects.
```
erikcorry, line 77:
```suggestion
  - maps. The keys must be strings, and the values must be valid UBJSON objects.
```

erikcorry (approved):

PR #423, opened 2022-02-07
No resize

erikcorry:


src/objects.h
```diff
@@ -483,7 +483,11 @@ class ByteArray : public HeapObject {
      return static_cast<ByteArray*>(byte_array);
   }
 
-  void resize(int new_length);
+  // Only for external byte arrays that were malloced.  Does not change the
```
kasperl, line 486: Would it be hard to change the accounting? I fear that if we don't, we will end up we will end up with really messed up accounting numbers and this will definitely lead to assertion failures and unexpected behavior.

kasperl (approved):

erikcorry:
I fixed the accounting to be consistent with itself.  If the resize fails it will be inconsistent with reality, but that's probably not so bad.  After all, nobody can really know how much the malloc is padding the allocations, so the accounting was approximate anyway.

PR #426, opened 2022-02-08
Add Pystone and scale DeltaBlue down to run on MicroPython

kasperl:


third_party/benchmarks/python/pystone.py
```diff
@@ -0,0 +1,247 @@
...
+TRUE = 1
+FALSE = 0
+
+def main(iterations):
+    startTime = time_ms()
+    for i in range(iterations):
+      pystones(LOOPS)
+    endTime = time_ms()
+    ms = (endTime - startTime) / iterations
+    print("Pystone (%s) - time per iteration: %i ms" % (sys.implementation.name, ms))
```
erikcorry, line 71: If it's a cut down version we should not that right in the print statement.  So easy to forget a few years down the line and wonder why Python is so fast.

erikcorry (approved):

third_party/benchmarks/python/pystone.py
floitsch: Was this comment addressed?

PR #431, opened 2022-02-09
Add support for external messaging

kasperl:


src/messaging.cc
```diff
@@ -386,6 +394,23 @@ Object* MessageDecoder::decode_byte_array(bool inlined) {
   return result;
 }
 
+bool MessageDecoder::decode_byte_array_external(void** data, int* length) {
+  int tag = read_uint8();
+  int encoded_length = read_cardinal();
+  if (tag == TAG_BYTE_ARRAY) {
+    *data = read_pointer();
+  } else if (tag == TAG_BYTE_ARRAY_INLINE) {
+    void* copy = malloc(encoded_length);
```
kasperl, line 403: I need to fix this before I commit.

erikcorry (approved):
LGTM, but I think we need some solution for malloc failures and GC.

src/messaging.cc
kasperl: Fixed.

kasperl:
Tried to improve the out-of-memory behavior. PTAL, @erikcorry.

src/messaging.h
```diff
@@ -205,4 +209,39 @@ class MessageDecoder {
...
+  // Send a message to a specific receiver.
+  bool send(int receiver, int type, void* data, int length);
+
+  // Support for handling failed allocations. Return true from the callback
+  // if you have cleaned up and want to retry the allocation. Returning false
+  // causes the message to be discarded.
+  virtual bool on_failed_allocation(int length) { return false; }
+
+  // Try collecting garbage. If asked to try hard, the system will preempt running
+  // processes and get the to stop before garbage collecting their heaps.
```
erikcorry, line 236:
```suggestion
  // processes and get them to stop before garbage collecting their heaps.
```

src/scheduler.h
```diff
@@ -85,6 +85,9 @@ class Scheduler {
   // Run a new program. Returns the process ID of the root process.
   int run_program(Program* program, char** args, ProcessGroup* group, Block* initial_block);
 
+  // Run a new external program. Returns the process.
```
erikcorry, line 88: External makes it sound a bit like we are fork-execing something.  Should we call it "native thread" instead?

erikcorry (approved):

src/scheduler.h
kasperl: Yeah. Maybe we should tweak the names a bit.

PR #432, opened 2022-02-09
Improve wifi open and close

kasperl:


lib/net/wifi.toit
```diff
@@ -15,26 +16,55 @@ import ..esp32
...
+      // connection. We always set it to a value when leaving this path
+      // and we always clear out the latch when we're done synchronizing.
+      wifi_connecting_ = monitor.Latch
+      wifi := wifi.Wifi
+      try:
+        wifi.set_ssid ssid password
+        with_timeout WIFI_CONNECT_TIMEOUT_: wifi.connect
+        with_timeout WIFI_DHCP_TIMEOUT_: wifi.get_ip
+        // Success: Register the WiFi connection, tell anyone who is waiting
+        // for it that the connetion is ready to be used (no exception), and
```
erikcorry, line 47:
```suggestion
        // for it that the connection is ready to be used (no exception), and
```

erikcorry (approved):

lib/net/wifi.toit
floitsch:
```suggestion
        // for it that the connection is ready to be used (no exception), and
```

lib/net/wifi.toit
```diff
@@ -15,26 +16,55 @@ import ..esp32
 WIFI_CONNECT_TIMEOUT_  ::= Duration --s=10
 WIFI_DHCP_TIMEOUT_     ::= Duration --s=16
 
+wifi_/wifi.Wifi? := null
+wifi_connecting_/monitor.Latch? := null
```
floitsch, line 20: Add a comment explaining what values are on this latch.
Maybe?
```suggestion
/**
Latch used to coordinate creation of the connection.
If the `get` is `null` the WiFi connection is established and can be used.
If it is not null, then it contains the error that was encountered during creation.
*/
wifi_connecting_/monitor.Latch? := null
```

lib/net/wifi.toit
```diff
@@ -15,26 +16,55 @@ import ..esp32
...
+        wifi_connecting_.set null
+      finally: | is_exception exception |
+        if is_exception:
+          wifi.close
+          wifi_connecting_.set exception.value
+        wifi_connecting_ = null
+
+  interface := WifiInterface_
+  if ssid == wifi_.ssid_ and password == wifi_.password_: return interface
+  interface.close
```
floitsch, line 59: Why even allocate the interface if we are not going to use it?

Would be simpler to just:
```
if ssid == wifi_.ssid_ and password == wifi_.password_: return WifiInterface_
```

floitsch (commented):
LGTM.

lib/net/wifi.toit
kasperl: The wifi_ isn't closed.

PR #435, opened 2022-02-09
Add url.decode to match url.encode

erikcorry:
Also adds 3 functions to the global namespace of utils.toit - I could make them private?


floitsch (approved):
LGTM.

PR #437, opened 2022-02-09
Handle Windows newlines.

floitsch:
Normalize line endings for source files when checking out.

### DIFF
--- a/src/compiler/scanner.cc
+++ b/src/compiler/scanner.cc
@@ -506,13 +506,12 @@ Token::Kind Scanner::scan_newline(int peek) {
   int indentation;
 
   do {
-    ASSERT(peek == 10 || peek == 13);
-    if (peek == 10) {
+    ASSERT(peek == '\n' || peek == '\r');
+    if (peek == '\r') {
       peek = advance();
       if (peek == 13) peek = advance();
     } else {
       peek = advance();
-      if (peek == 10) peek = advance();
     }
 
     // Compute indentation level of next line.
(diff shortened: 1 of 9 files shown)


src/compiler/scanner.cc
```diff
@@ -506,13 +506,12 @@ Token::Kind Scanner::scan_newline(int peek) {
   int indentation;
 
   do {
-    ASSERT(peek == 10 || peek == 13);
-    if (peek == 10) {
+    ASSERT(peek == '\n' || peek == '\r');
+    if (peek == '\r') {
       peek = advance();
       if (peek == 13) peek = advance();
```
erikcorry, line 512:
```suggestion
      if (peek == '\r') peek = advance();
```

src/compiler/toitdoc_parser.cc
```diff
@@ -951,28 +955,37 @@ int ToitdocParser::peek() {
   auto text = _toitdoc_source->text();
   ASSERT(_index <= _toitdoc_source->size());
```
erikcorry, line 956: Maybe also assert that the text is null terminated?
I assum that's why the index is allowed to be equal to the size?

src/compiler/toitdoc_parser.cc
```diff
@@ -951,28 +955,37 @@ int ToitdocParser::peek() {
...
+    //   '\n' (only ' ' or '\0').
     // Callers thus don't need to worry about '\n', but can simply check for
     //   whitespace by looking for spaces.
     if (is_single_line) return '\0';
     if (_next_index != -1) {
       // We already computed the indentation once and know that we aren't at a dedent.
       return ' ';
     }
-    _next_index = _index + 1;
+    if (c == '\r' && text[_index + 1] == '\n') {
```
erikcorry, line 968: I guess we are assuming null termination here and below
to avoid out of bounds read?

erikcorry (approved):

src/compiler/toitdoc_parser.cc
floitsch: Yes. Added a comment and assert.

src/compiler/toitdoc_parser.cc
floitsch: yes. added a comment.

PR #440, opened 2022-02-10
Don't use address rounding to find blocks.

erikcorry:
This change removes the process pointer from blocks,
and stops using the blocks to detect program pointers
during GC (incl. stack traversal).

Rounding down an object address to find the block is
no longer done.


src/interpreter_run.cc
```diff
@@ -18,12 +18,13 @@
 #include <math.h>
 
 #include "flags.h"
-#include "objects_inline.h"
 #include "printing.h"
 #include "process.h"
 #include "scheduler.h"
 #include "vm.h"
 
+#include "objects_inline.h"
```
floitsch, line 26: Why the move?
If it's important add a comment.

src/objects.cc
```diff
@@ -252,11 +253,8 @@ void Array::roots_do(RootCallback* cb) {
...
 }
 
-void Stack::roots_do(RootCallback* cb) {
+void Stack::roots_do(Program* program, RootCallback* cb) {
   int top = this->top();
-  Process* owner = this->owner();
-  ASSERT(owner != null);
-  Program* program = owner->program();
   // Skip over pointers into the bytecodes.
   void* bytecodes_from = program->bytecodes.data();
```
floitsch, line 259: Might be cleaner to also use `process->on_program_heap` instead.

floitsch (approved):
LGTM.

PR #447, opened 2022-02-14
Remove unused miniz

erikcorry:
I may be missing something, but I can't see where we use this.


kasperl:
It's dead, Jim.

kasperl (approved):
Looks good to me.

PR #455, opened 2022-02-15
More integration of the Dartino GC

erikcorry:


floitsch (approved):
LGTM.
Why not remove them?

src/third_party/dartino/object_memory.cc
```diff
@@ -84,10 +80,10 @@ HeapObject *Space::object_at_offset(word offset) {
 }
 
 void Space::adjust_allocation_budget(uword used_outside_space) {
-  uword used = used() + used_outside_space;
+  uword used_bytes = used() + used_outside_space;
```
floitsch, line 83: can/should we change the `used` function too?

floitsch (approved):
LGTM.

PR #468, opened 2022-02-18
A tool to analyze stacktraces.

erikcorry:
Pipe in Backtrace: lines from the serial port crash.
Requires the correct toit.elf file.


floitsch (approved):
LGTM.

PR #493, opened 2022-02-22
Remove Dartino code related to schema change

erikcorry:
Changing object layout during program run is not something we do.


floitsch (approved):
LGTM.

PR #494, opened 2022-02-22
Add abstract class HeapObjectVisitor from Dartino

erikcorry:
I already lower-cased the method names in this class that is from object.h in Dartino.


floitsch (approved):
LGTM.

PR #495, opened 2022-02-22
Add pseudo-objects that are put on the heap for GC.

erikcorry:
This adds support for some objects that are not
real heap objects, but are placed on the heap to
make it possible to iterate all objects in memory order.

Also renames call sites to roots_do which is the Toit name for
visit_pointers.

### DIFF
--- a/src/third_party/dartino/mark_sweep.h
+++ b/src/third_party/dartino/mark_sweep.h
@@ -81,9 +81,9 @@ class FreeList {
   void AddChunk(uword free_start, uword free_size) {
     FreeListChunk* result = FreeListChunk::create_at(free_start, free_size);
     if (!result) {
-      // If the chunk is too small to be turned into an actual
-      // free list chunk we turn it into fillers to be coalesced
-      // with other free chunks later.
+      // Since the chunk was too small to be turned into an actual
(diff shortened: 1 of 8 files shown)


src/third_party/dartino/mark_sweep.h
```diff
@@ -79,30 +79,26 @@ class FreeList {
...
-    // with other free chunks later.
-    if (free_size < FreeListChunk::SIZE) {
-      ASSERT(free_size <= 2 * WORD_SIZE);
-      Object** free_address = reinterpret_cast<Object**>(free_start);
-      for (uword i = 0; i * WORD_SIZE < free_size; i++) {
-        free_address[i] = StaticClassStructures::one_word_filler_class();
-      }
+    FreeListChunk* result = FreeListChunk::create_at(free_start, free_size);
+    if (!result) {
+      // If the chunk is too small to be turned into an actual
```
floitsch, line 84: Is the comment still accurate?

floitsch (approved):
LGTM.

src/third_party/dartino/mark_sweep.h
erikcorry: Pretty much, but I edited it for clarity.

PR #497, opened 2022-02-22
Make GnssLocation compatible with toit-gnss-location:v1.0.2

kasperl:


lib/device.toit
```diff
@@ -106,5 +106,69 @@ class Gnss extends impl.Gnss_:
...
+  location -> GnssLocation?:
     return super
+
+
+/**
+GNSS location consisting of coordinates and accuracy measurements.
+*/
+class GnssLocation:
+  latitude/float
+  longitude/float
```
lask, line 118: A quick eye-ball comparison with `GnssLocation` in toit-gnss-location shows that they are not entirely the same. For instance, the package class contains an `==` operator and a `hash_code` method. Do we want to have those in this version?

I agree that the serialization is compatible which may be sufficient.

lask (approved):

lib/device.toit
kasperl: For now, I'm just trying to get 1.6 to make sense of the data it gets when it receives serialized data from the kernel. This class here is lifted out of lib/cellular, so I wanted to not introduce too many changes to the API. Maybe we should replace this with the real deal (inheriting from Location, ==, hash_code, etc.) as a follow up.

PR #498, opened 2022-02-23
Rename FreeListChunk to FreeListRegion.

erikcorry:
We have too many things called chunks in our system.
The Dartino GC adds two more.  Reduce that to one.


floitsch (approved):
LGTM.

PR #501, opened 2022-02-23
Fix clear-health target.

floitsch:


erikcorry (approved):
Do you need to mkdir -p it somewhere?

floitsch:
No. The health-check rules create the directory when necessary.

PR #503, opened 2022-02-23
Run CI on pull requests, but don't do work twice.

floitsch:


floitsch:
Doesn't seem to work.

PR #505, opened 2022-02-23
Fix many compile errors in new heap.

erikcorry:
In Toit we need a reference to the program in order
to determine the size and layout of objects.  Some of
this change is threading that pointer through to the
places that need it.

In the long run we might want to find a way to avoid
this issue - probably the GC will be faster if the
size/layout information is more self-contained in the
heap objects.


floitsch (approved):
LGTM.

PR #506, opened 2022-02-23
Add two_space_heap.cc and two_space_heap.h from Dartino.

erikcorry:
Unedited.  These were called heap.h and heap.cc in Dartino.


floitsch (approved):
LGTM.

PR #508, opened 2022-02-24
Add firmware update writer (OTA support)

kasperl:


src/primitive_esp32.cc
```diff
@@ -72,6 +72,184 @@ namespace toit {
...
+    if (err != ESP_OK) {
+      ESP_LOGE("Toit", "esp_image_verify failed (%s)!", esp_err_to_name(err));
+      ota_partition = null;
+      OUT_OF_BOUNDS;
+    }
+
+    // The system SHA256 checksum is optional, so we add an explicit verification
+    // that we control.  (There is also a non-optional checksum, but it is only one
+    // byte, and so not really reliable.)
+    if (expected->is_byte_array()) {
```
erikcorry, line 213: This should use expected->byte_content so that it works with slices too and CoW-byte arrays too.

erikcorry (approved):

PR #509, opened 2022-02-24
Always free sent external message payloads

kasperl:
Not sure if this is the most useful API, but now it is consistent.

### DIFF
--- a/src/messaging.cc
+++ b/src/messaging.cc
@@ -483,7 +483,10 @@ bool ExternalSystemMessageHandler::send(int pid, int type, void* data, int lengt
   }
 
   uint8* buffer = unvoid_cast<uint8*>(malloc(buffer_size));
-  if (buffer == null) return false;
+  if (buffer == null) {
+    free(data);
+    return false;
+  }
   MessageEncoder encoder(buffer);
   encoder.encode_byte_array_external(data, length);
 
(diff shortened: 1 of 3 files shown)


src/messaging.cc
```diff
@@ -483,7 +483,10 @@ bool ExternalSystemMessageHandler::send(int pid, int type, void* data, int lengt
   }
 
   uint8* buffer = unvoid_cast<uint8*>(malloc(buffer_size));
-  if (buffer == null) return false;
+  if (buffer == null) {
+    free(data);
```
mikkeldamsgaard, line 487: Not sure that freeing the data on allocation error is the best approach. As a consumer of the Api, if this returns false, the most obvious remediation action would be to retry the operation after doing other corrective actions (like invoking the GC). If the data is free'ed then the client code needs to reallocate the memory and copy over the payload.

kasperl: I agree. Maybe it would be better to have a flag that controls this (a boolean passed to send, maybe with a reasonable default to not deallocate on errors)?

kasperl:
PTAL @erikcorry and @mikkeldamsgaard.

mikkeldamsgaard (commented):
LGTM

erikcorry (approved):

PR #513, opened 2022-02-25
Add support for sequential RPCs across tasks

kasperl:


lib/rpc/rpc.toit
```diff
@@ -2,31 +2,33 @@
...
-    if arguments is RpcSerializable: arguments = arguments.serialize_for_rpc
-    return synchronizer_.send -1: | id pid |
-      process_send_ pid SYSTEM_RPC_REQUEST_ [ id, name, arguments ]
-
-  invoke pid/int name/int arguments/any -> any:
+  invoke pid/int name/int arguments/any --sequential/bool -> any:
     if arguments is RpcSerializable: arguments = arguments.serialize_for_rpc
-    return synchronizer_.send pid: | id pid |
+    send ::= : synchronizer_.send pid: | id pid |
       process_send_ pid SYSTEM_RPC_REQUEST_ [ id, name, arguments ]
```
lask, line 30: I prefer this formatting as I find it more clear what belongs to the inner block.
```suggestion
    send ::= : 
      synchronizer_.send pid: | id pid |
        process_send_ pid SYSTEM_RPC_REQUEST_ [ id, name, arguments ]
```

tests/rpc_test.toit
```diff
@@ -216,6 +217,41 @@ test_blocking myself/int broker/RpcBroker tasks/int [test] -> none:
...
+test_sequential myself/int broker/RpcBroker -> none:
+  tasks ::= 10
+  name ::= 800
+  latches ::= {:}
+  concurrency := 0
+  broker.register_procedure name:: | args |
+    result := null
+    try:
+      concurrency++
+      expect.expect_equals 1 concurrency  // Should be sequential!
```
lask, line 229: wouldn't it be better with a `yield` here to allow the other tasks to jump in?

lask (approved):
lgtm

tests/rpc_test.toit
kasperl: The last.get call blocks until explicitly told to continue from the outside. That was intended to be (even) more robust than yielding.

lib/rpc/rpc.toit
kasperl: Done.

tests/rpc_test.toit
lask: Perfect.

PR #514, opened 2022-02-25
Remove weak pointer handling from Dartino heap.

erikcorry:
Finalizers are rather different in Toit, so this code
is not useful.


kasperl (approved):
Looks good to me.

PR #515, opened 2022-02-25
Remove ScavengeVisitor.

erikcorry:
This class was only used for single-space heaps.
We will be using only two-space generational heaps.


floitsch (commented):
LGTM.

PR #516, opened 2022-02-25
Adapt CloneInToSpace from Dartino to Toit.

erikcorry:


src/objects.h
```diff
@@ -215,6 +215,10 @@ class HeapObject : public Object {
     return HeapObject::cast(_at(HEADER_OFFSET));
   }
 
+  INLINE void set_forwarding_address(HeapObject* destination) {
+    _at_put(HEADER_OFFSET, destination);
```
floitsch, line 219: Add a comment to the `HEADER_OFFSET` explaining that it can contain the forward address.

floitsch (approved):
LGTM.

PR #517, opened 2022-02-25
Add a debug artifact.

floitsch:


floitsch:
We have seen some flakes with an external test not seeing its packages.
This PR starts to add debug information so we can narrow it down.

erikcorry (approved):

PR #519, opened 2022-02-25
Merge Heap and TwoSpaceHeap.

erikcorry:
We only have two-space heaps so no need to
have this difference with lots of virtual
methods.  Merges the classes, but doesn't
move code around yet because that would make
the CL unreadable.  (There's still a non-Dartino
class called Heap that needs to be merged.)


floitsch (approved):
LGTM.

PR #520, opened 2022-02-25
Rename space_ to semi_space_.

erikcorry:
Also rearranges TwoSpaceHeap so it has a more
conventional public-protected-private ordering.


floitsch (approved):
LGTM.

PR #524, opened 2022-02-28
Yield if Writer.write can't write everything

erikcorry:


floitsch (approved):
LGTM.

floitsch:
Alternatively, yielding when there is no progress (return 0) would also work.

PR #529, opened 2022-03-01
url.encode returns strings.

erikcorry:
The whole point of url.encode is that the result is
a normal string, and it's more convenient this way for
concatenating ?-& URLs.


floitsch (approved):
LGTM.

PR #534, opened 2022-03-03
Add more documentation to 'process_stats'.

floitsch:


floitsch:
Note that my lists always have 0 in the first entries:
```
0
0
0
0
1856
3
3
```
This is for the following program:
```
main:
  foo := [1, 2, 3]
  process_stats.do: print it
  print foo
```
I would have expected the allocated memory to be non-zero as there is a pointer to `foo`.

kasperl:
If allocated/reserved is zero, then maybe some of @erikcorry's GC changes have messed with this?

kasperl (approved):
Looks good to me.

PR #554, opened 2022-03-09
Treat WSL paths as absolute.

floitsch:
This might also work for network drives.

### DIFF
--- a/src/compiler/filesystem_local_win.cc
+++ b/src/compiler/filesystem_local_win.cc
@@ -52,10 +52,16 @@ char FilesystemLocal::path_separator() {
 }
 
 char* FilesystemLocal::root(const char* path) {
-  char* result = new char[4];
-  memcpy(result, path, 3);
-  result[3] = '\0';
-  return result;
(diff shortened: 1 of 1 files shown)


src/compiler/filesystem_local_win.cc
```diff
@@ -42,18 +42,27 @@ bool FilesystemLocal::is_absolute(const char* path) {
   if (SourceManager::is_virtual_file(path)) return true;
   int length = strlen(path);
   if (length < 3) return false;
-  return path[1] == ':' && path[2] == PATH_SEPARATOR;
+  // Either "c:\" or "\\wsl$"
```
kasperl, line 45: You don't check for the `wsl$` part so the comment is a bit confusing for me.

kasperl (approved):
Looks good to me.

erikcorry (approved):

src/compiler/filesystem_local_win.cc
floitsch: Added better comments.

PR #560, opened 2022-03-10
We should trigger a GC when LwIP has trouble allocating

erikcorry:


erikcorry:
Probably needs the same code for UDP, and perhaps we need to be able to trigger a system-wide GC instead of just a process GC.

kasperl (approved):
Looks good to me.

PR #561, opened 2022-03-11
Fix lost monitor notifications

kasperl:


lib/core/monitor_impl.toit
```diff
@@ -55,54 +56,67 @@ class __Monitor__:
...
+          // await condition. Without this check, we will be constantly re-evaluating
+          // conditions because evaluating any condition will lead to infinite evaluations
+          // of all other conditions. The state can change before the first evaluation
+          // of the await condition, so we take care of that.
+          resume_ --state_changed=first
+          first = false
+          // Check for task cancelation and timeout. We have to do this here because we may
+          // have been notified while evaluating the condition in which case it would be wrong
+          // to just block and wait for the next notification.
+          if is_non_critical and self.is_canceled_: throw CANCELED_ERROR
```
kasperl, line 87: This is changed in the loop.

lib/core/monitor_impl.toit
```diff
@@ -55,54 +56,67 @@ class __Monitor__:
...
+      is_non_critical ::= self.critical_count_ == 0
+      first := true
+      while true:
+        // Check for task cancelation and timeout.
+        if is_non_critical and self.is_canceled_: throw CANCELED_ERROR
+        if deadline and Time.monotonic_us >= deadline: return false
+        if not owner_:
+          // Re-take the lock.
+          owner_ = self
+          // Evaluate the condition. This is usually a quick operation, but it may take time
```
kasperl, line 69: Comment updated.

lib/core/monitor_impl.toit
```diff
@@ -55,54 +56,67 @@ class __Monitor__:
...
-      // Check for task timeout.
-      if deadline and Time.monotonic_us >= deadline: throw DEADLINE_EXCEEDED_ERROR
-      // If the monitor isn't owned by anyone at this point, we are ready
-      // to take it.
-      if not owner_: break
-      // Wait until notified. When we get back the monitor might be owned by
-      // someone else.
-      wait_ self
+    try:
+      while true:
```
kasperl, line 108: No logical changes in the loop body (minor reformatting and indentation due to try-finally).

lib/monitor.toit
```diff
@@ -246,13 +246,15 @@ monitor ResourceState_:
...
   notify_:
-    if not resource_: return
-    state := read_state_ group_ resource_
-    state_ |= state
-    // TODO(kasper): Only notify if we added new bits?
+    resource := resource_
+    if resource:
+      state := read_state_ group_ resource
+      state_ |= state
+    // Always call the super implementation to avoid getting
```
kasperl, line 256: This may not have been a real problem, because leaving the dispose call will notify waiters (state changed) and so anyone waiting for resource changes will be unblocked and the condition rechecked. However, it feels iffy to not let the super class know about the notification given that it is the super class that requested the timeout/cancelation notifications in the first place.

lib/core/task.toit
```diff
@@ -182,9 +182,33 @@ class Task_:
...
+    return Timer_
+
+  // Releasing a timer will make it available for reuse or close it if the
+  // single element cache is already filled.
+  release_timer_ timer/Timer_:
+    existing := timer_
+    if existing:
+      timer.close
+    else:
+      timer_ = timer
```
lask, line 211: I what cases would it happen that we release a timer and it is closed because another timer already exists?

Is it correctly understood that we only reuse the Timer object. We do not reuse the deadline?

tests/task_critical_region_test.toit
```diff
@@ -29,3 +30,14 @@ test_cancel_can_timeout:
       with_timeout --ms=1:
         critical_do:
           sleep --ms=10000
+
+test_deadline_in_critical:
+  with_timeout --ms=100:
+    expect_not_null task.deadline
+    critical_do:
+      expect_not_null task.deadline
```
lask, line 38: Maybe also add a test that the timeout triggers,
```
  expect_throw "DEADLINE_EXCEEDED": 
    with_timeout --ms=10:
      critical_do:
        while true: yield

File: tests/task_critical_region_test.toit [hunk 7]
```diff
@@ -29,3 +30,14 @@ test_cancel_can_timeout:
...
+
+  with_timeout --ms=100:
+    expect_not_null task.deadline
+    critical_do --no-respect_deadline:
+      expect_null task.deadline
```
lask, line 43: Maybe add a test that shows that the deadline doesn't trigger
```
Maybe also add a test that the timeout triggers,
```
  expect_no_throw:
    with_timeout --ms=10:
      critical_do:
        sleep --ms=50
```

Comment (lask) [inline tests/task_critical_region_test.toit:43]: Maybe add a test that shows that the deadline doesn't trigger
```
  expect_no_throw:
    with_timeout --ms=10:
      critical_do:
        sleep --ms=50
```

**Review by lask — COMMENTED** [review]
Some initial comments

File: lib/core/task.toit (hunk 5 above)
Comment (kasperl) [reply lib/core/task.toit:211]: If we end up calling acquire more than once before we start releasing. Stuff like this:

```
with_timer_:
  with_timer_:
    // do stuff
```

File: tests/task_critical_region_test.toit (hunk 6 above)
Comment (kasperl) [reply tests/task_critical_region_test.toit:38]: It should not. If it does it will throw.

File: tests/task_critical_region_test.toit (hunk 7 above)
Comment (kasperl) [reply tests/task_critical_region_test.toit:43]: Any unexpected exceptions make the tests fail, so we do check for this.

File: lib/core/utils.toit [hunk 8]
```diff
@@ -178,12 +178,18 @@ Within $block, the current task won't be interrupted by cancellation exceptions.
...
+  deadline/int? := null
   self.critical_count_++
+  if not respect_deadline:
+    deadline = self.deadline_
+    self.deadline_ = null
   try:
     block.call
   finally:
+    if not respect_deadline:
+      self.deadline_ = deadline
```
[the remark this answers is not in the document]
kasperl: Our intention is to only do block-structured changes to deadline. I think we do that correctly, so no.

tests/task_critical_region_test.toit
kasperl: Ah, you mean from within the critical region. I will try that.

tests/task_critical_region_test.toit
kasperl: I've added a different test case (with sleep) that triggers a timeout.

tests/task_critical_region_test.toit
kasperl: I've extended the testing with a case that sleeps longer than the timeout allows, but from within a critical region that doesn't respect deadlines.

kasperl: Done.

tests/task_critical_region_test.toit
```diff
@@ -0,0 +1,55 @@
...
+    critical_do:
+      expect_not_null task.deadline
+
+  expect_throw DEADLINE_EXCEEDED_ERROR:
+    with_timeout --ms=100:
+      expect_not_null task.deadline
+      critical_do:
+        sleep --ms=1000  // Deadline respected. Exception thrown!
+
+  with_timeout --ms=100:
```
floitsch, line 46: If the machine is very loaded, could this lead to a flaky test?

tests/task_critical_region_test.toit
```diff
@@ -0,0 +1,55 @@
...
+    expect_not_null task.deadline
+    critical_do --no-respect_deadline:
+      expect_null task.deadline
+
+  with_timeout --ms=100:
```
floitsch, line 51: Ditto: could this lead to a flaky test?

floitsch (approved):
LGTM.

tests/task_critical_region_test.toit
kasperl: Yeah, most of these timeout test have that problem. I have tried to tweak the parameters so it is very unlikely to happen in practice, but you're right.

tests/task_critical_region_test.toit
kasperl: Like above :)

PR #568, opened 2022-03-14
Improve error message for bad static accesses.

floitsch:
We treated `A.foo` as `(A).foo` and complained when `A` didn't have any
unnamed constructor. Now we report that we can't find any 'foo' static
or constructor.

### DIFF
--- a/src/compiler/resolver_method.cc
+++ b/src/compiler/resolver_method.cc
@@ -2609,7 +2609,10 @@ void MethodResolver::_visit_potential_call_dot(ast::Dot* ast_dot,
   // We know that this isn't a constructor call, as the `visit_potential_call` would have
   // caught that one.
   auto ast_receiver = ast_dot->receiver();
-  if (ast_receiver->is_Identifier() || scope()->is_prefixed_identifier(ast_receiver)) {
+  // If this is for the LSP just follow the normal path.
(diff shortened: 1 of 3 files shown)


src/compiler/resolver_method.cc
```diff
@@ -2602,6 +2602,28 @@ void MethodResolver::_visit_potential_call_identifier(ast::Node* ast_target,
...
+  // We know that this isn't a constructor call, as the `visit_potential_call` would have
+  // caught that one.
+  auto ast_receiver = ast_dot->receiver();
+  if (ast_receiver->is_Identifier() || scope()->is_prefixed_identifier(ast_receiver)) {
+    auto candidates = _compute_target_candidates(ast_receiver, scope());
+    if (!candidates.encountered_error &&
+        (candidates.klass != null && candidates.nodes.is_empty())) {
+      auto klass = candidates.klass;
+      auto class_interface = klass->is_interface() ? "Interface" : "Class";
+      report_error(ast_dot, "%s '%s' does not have any static member or constructor with name '%s'.",
```
kasperl, line 2618: Looks like the . at end is inconsistent with the "only has named constructors" case.

kasperl (approved):
Looks good to me.

src/compiler/resolver_method.cc
floitsch: done.

PR #578, opened 2022-03-18
Clarify which objects can be added to sets/maps

erikcorry:


lib/core/collections.toit
```diff
@@ -2177,7 +2177,22 @@ simple_rebuild_hash_index_ old_index index_ -> none:
...
+  that objects that test equal also have the same hash code.
+  However, objects that test unequal are not required to have
+  different hash codes: Hash code clashes are allowed, but should
+  be rare to maintain good performance.
+Strings and numbers fulfill these requirements and can be used as
+  keys in sets.
+Very small sets may not call the hash_code method, and therefore
+  very small sets may appear to work with keys that are missing
+  the method.  However, as soon as a they hit a certain size (currently
+  2) an error will be thrown.
```
floitsch, line 2194: I wonder if we should just call `hash_code`.

lib/core/collections.toit
```diff
@@ -2448,6 +2463,22 @@ class IdentitySet extends Set:
...
+  that objects that test equal also have the same hash code.
+  However, objects that test unequal are not required to have
+  different hash codes: Hash code clashes are allowed, but should
+  be rare to maintain good performance.
+Strings and numbers fulfill these requirements and can be used as
+  keys in maps.
+Very small maps may not call the hash_code method, and therefore
+  very small maps may appear to work with keys that are missing
+  the method.  However, as soon as a they hit a certain size (currently
+  2) an error will be thrown.
```
floitsch, line 2480: As before. I think we should just call `hash_code`.

floitsch (approved):
LGTM.

floitsch (approved):
Still LGTM.

PR #583, opened 2022-03-21
Mikkel i2s

erikcorry:
Copy of #530 so that we can run tests and commit.


mikkeldamsgaard:
I just pushed more changes to the remote i2s branch, for loading data to a predefined buffer and to fix the build error you just discovered

PR #585, opened 2022-03-22
Add hash_code to fonts

erikcorry:


lib/font/font.toit
```diff
@@ -40,10 +43,20 @@ class Font:
     // The primitive takes a small array (not a list) of byte arrays, so we
     // convert to that.
     array := create_array_ glyph_byte_data
+    hash_code = calculate_hash_code_ array
     proxy_ = font_get_nonbuiltin_ resource_freeing_module_ array
     add_finalizer this:: this.finalize_
 
-  constructor.protected_ .proxy_:
+  static calculate_hash_code_ array/Array_ -> int:
```
kasperl, line 50: Maybe I'm missing something, but why does it make sense to have a hash code that is based on the font bits if == is based on identity?

lib/font/font.toit
```diff
@@ -4,11 +4,15 @@
...
   proxy_ := ?
+  hash_code / int
+
+  static hash_code_counter_ := 1234
 
   proxy: return proxy_
 
   constructor.get font_name/string:
     proxy_ = font_get_ resource_freeing_module_ font_name
+    hash_code = hash_code_counter_++
```
kasperl, line 15: Can this be put as an initializer expression near the field instead?

kasperl (approved):
Looks good to me.

PR #588, opened 2022-03-22
ifdef out code that is not in the new GC

erikcorry:


src/heap.cc
```diff
@@ -51,17 +51,31 @@ class ScavengeScope : public Locker {
...
 Heap::Heap(Process* owner, Program* program, Block* initial_block)
     : RawHeap(owner)
     , _program(program) {
   if (initial_block == null) return;
   _blocks.append(initial_block);
 }
+#else
+Heap::Heap(Process* owner, Program* program)
+    : _program(program)
+    , _in_gc(false)
```
kasperl, line 64: I think you should use the field initializers for this.

kasperl (approved):
Looks good to me.

PR #589, opened 2022-03-22
Bring in Dartino code from program.cc: Steps in a garbage collection.

erikcorry:
1:1 copy of code from program.cc of Dartino.


kasperl (approved):
Looks good to me.

PR #591, opened 2022-03-23
Added support for reading to a user supplied buffer in i2s.

mikkeldamsgaard:


erikcorry:
Moved to #607 so the tests can be run.

PR #596, opened 2022-03-24
Bring over HeapObjectPointerVisitor from Dartino.

erikcorry:
This is just a small adapter class.

Also a bunch of compile fixes for the new GC - still doesn't
link.


floitsch (approved):
LGTM.

PR #603, opened 2022-03-25
Avoid new in new gc

erikcorry:
This meant the FreeList class had to be moved.  It was not
changed.

Currently we don't share the unused newspaces between
processes.  At some point we will want to have a single
newspace and make sure we don't do parallel scavenges.


floitsch (approved):
LGTM.

PR #605, opened 2022-03-26
Use services framework in the system process

kasperl:


kasperl:
Next step: Clean up "descriptor" and CloseableProxy handling and let the services framework take care of that too.

lib/system/api/containers.toit
```diff
@@ -0,0 +1,61 @@
+// Copyright (C) 2022 Toitware ApS. All rights reserved.
+// Use of this source code is governed by an MIT-style license that can be
+// found in the lib/LICENSE file.
+
+import uuid
+import system.services show ServiceClient
+
+interface ContainersService:
```
erikcorry, line 8: I would call it ContainerService.

It's dog food, not dogs food, even though more than one dog eats it.  It's an apple tree, not an apples tree.

system/containers.toit
```diff
@@ -118,18 +118,98 @@ class ContainerImageFlash extends ContainerImage:
...
+    if not image: return null
+    return image.start.id
+
+  uninstall_image id/uuid.Uuid -> none:
+    image/ContainerImage? := lookup_image id
+    if not image: return
+    image.delete
+
+  image_writer_open size/int -> int:
+    unreachable  // <-- TODO(kasper): Nasty.
```
erikcorry, line 170: Why are they unreachable?

system/containers.toit
```diff
@@ -118,18 +118,98 @@ class ContainerImageFlash extends ContainerImage:
...
+    writer := lookup_descriptor client handle
+    image := add_flash_image writer.commit
+    return image.id
+
+  image_writer_close client/int handle/int -> none:
+    writer := lookup_descriptor client handle
+    unregister_descriptor client handle
+    writer.close
+
+class ContainerManager extends ContainersServiceDefinition implements SystemMessageHandler_:
```
erikcorry, line 202: This class and ContainersServiceDefinition would benefit from a line or two explaining what they are for.

erikcorry (approved):

lib/system/api/containers.toit
kasperl: Yeah, I think you're right. Will fix in follow up PR.

system/containers.toit
kasperl: They are so that the service definition appears to implement the **client** interface. I'm working on a way to improve this.

system/containers.toit
kasperl: You're right. I'm still really messing with things right now, but they do deserve toitdoc comments.

PR #606, opened 2022-03-26
Allocate initial chunks for new GC heap early.

erikcorry:
A new heap must immediately have some space to
allocate, since some allocations happen immediately
and can't trigger a GC.

But in Toit any allocation can fail, and that includes the
initial pages for a new heap for a new process.  We handle
this by allocating the first page in the primitive before
trying to create the new process.  This way we can report
back allocation failures to the primitive machinery that
retries the primitive.

This CL extends that logic to the new GC.  In the case of the
new GC we start each two-space heap off with a page per
semispace, which means we will be able to perform a young
space GC.

### DIFF
--- a/src/heap.h
+++ b/src/heap.h
@@ -32,6 +32,44 @@ namespace toit {
 
 class ObjectNotifier;
 
+#ifdef LEGACY_GC
+typedef Block InitialMemory;
+#else
+struct InitialMemory {
+  Chunk* chunk_1;
+  Chunk* chunk_2;
+};
+#endif
+
+// A class that uses a RAII destructor to free memory already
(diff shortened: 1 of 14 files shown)


src/heap.h
```diff
@@ -32,6 +32,44 @@ namespace toit {
...
+#ifdef LEGACY_GC
+    initial_memory = null;
+#else
+    chunks.chunk_1 = null;
+    chunks.chunk_2 = null;
+#endif
+  }
+
+  // Allocates initial pages for heap.  Returns success.
+  bool Allocate();
```
kasperl, line 67: allocate

kasperl (approved):
Looks good to me!

PR #611, opened 2022-03-28
Add more configuration options to RMT

lask:


lib/rmt.toit
```diff
@@ -205,34 +204,20 @@ class Channel:
     rmt_config_rx_ pin.num num mem_block_num clk_div flags idle_threshold filter_en filter_ticks_thresh rx_buffer_size
     idle_threshold_ = idle_threshold
     rx_buffer_size_ = rx_buffer_size
-
-  /**
```
lask, line 209: This was just moved.

lib/rmt.toit
```diff
@@ -296,9 +301,12 @@ The $rx channel must be configured for receiving (see $Channel.config_rx).
 The $tx channel must be configured for transferring (see $Channel.config_tx).
 */
 transfer_and_receive --rx/Channel --tx/Channel signals/Signals max_returned_bytes/int -> Signals:
-  if max_returned_bytes > rx.rx_buffer_size: throw "maximum returned buffer size greater than allocated RX buffer size"
+  if not rx.rx_buffer_size or not rx.rx_clk_div: throw "rx channel not configured"
```
erikcorry, line 304: This is the same as `not (rx.rx_buffer_size or not rx.rx_clk_div)` which I think is not what you wanted.

lib/rmt.toit
```diff
@@ -296,9 +301,12 @@ The $rx channel must be configured for receiving (see $Channel.config_rx).
...
 
-  result := rmt_transfer_and_read_ tx.num rx.num signals.bytes_ max_returned_bytes
+  if not rx.rx_buffer_size or max_returned_bytes > rx.rx_buffer_size: throw "maximum returned buffer size greater than allocated RX buffer size"
```
erikcorry, line 306: We are rechecking rx.rx_buffer_size for nullness, which may not be what you want.

src/resources/rmt_esp32.cc
```diff
@@ -153,6 +152,13 @@ PRIMITIVE(config_rx) {
   return process->program()->null_object();
 }
 
+PRIMITIVE(set_idle_threshold) {
+  ARGS(int, channel_num, uint16, threshold)
+  esp_err_t err = rmt_get_rx_idle_thresh(static_cast<rmt_channel_t>(channel_num), &threshold);
```
erikcorry, line 157: It's called `set`, but it calls `get`.  Is this what you want?

lib/rmt.toit
lask: Changed it to `not rx.rx_buffer_size and rx.rx_clk_div`.

src/resources/rmt_esp32.cc
lask: I want `set`, so I can dynamically change the threshold. I should have caught this when I made the second argument a pointer.

lib/rmt.toit
lask: Removed the second check.

erikcorry (approved):

PR #614, opened 2022-03-28
Handle finalizers in new-gen of new heap

erikcorry:
Some formatting is wrong in order to make it easier to review.

Will be fixed in a later formatting-only CL.


src/heap.cc
```diff
@@ -493,6 +493,13 @@ int ObjectHeap::gc() {
 
 #else
 
+class HasForwardingAddress : public LivenessOracle {
+ public:
+  virtual bool is_alive(HeapObject* object) {
```
kasperl, line 498: override

src/os.cc
```diff
@@ -153,7 +153,10 @@ void* OS::allocate_pages(uword size) {
   if (result == null) return null;
   uword result_end = reinterpret_cast<uword>(result) + size;
   int attempt = 0;
-  while (result < _single_range.address || result_end > reinterpret_cast<uword>(_single_range.address) + _single_range.size) {
+  uword numeric = reinterpret_cast<uword>(result);
```
kasperl, line 156: address? numeric_address?

kasperl (approved):
Looks good to me.

PR #616, opened 2022-03-28
Unbork ESP32 after 'Allocate initial chunks ... early'

erikcorry:


lask (approved):
LGTM! (also tested it and it works)

PR #618, opened 2022-03-29
Make transfer and receive signals optional in transfer and receive primitive

lask:


erikcorry (approved):
Maybe (next CL) transmit/receive or send/receive?

PR #621, opened 2022-03-29
Fix compilation of new-GC.

erikcorry:
The heap histogram was not implemented for the new heap.

There are two ways to iterate heaps.
1) Using ObjectHeap::Iterator (not implemented for the new heap).
2) Using a visitor object, passed to iterate_objects (not implemented
   for the old heap).

We fix this by making a third version that calls one of the others:
3) Passing std::function to a do-like method.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,8 +18,11 @@ file(GLOB toit_core_SRC
   "*.cc"
   )
 
+# Exclude files with a main(), and files that are not used in the
+# compiler.
 list(FILTER toit_core_SRC EXCLUDE REGEX "/(toit|toit_run_image|vm).cc$")
 
+# Files that are used in the VM, but not in the compiler.
 file(GLOB toit_vm_runtime_SRC
   "vm.h"
   "vm.cc"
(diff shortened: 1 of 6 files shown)


src/CMakeLists.txt
```diff
@@ -18,7 +18,12 @@ file(GLOB toit_core_SRC
   "*.cc"
   )
 
-list(FILTER toit_core_SRC EXCLUDE REGEX "/(toit|toit_run_image).cc$")
+list(FILTER toit_core_SRC EXCLUDE REGEX "/(toit|toit_run_image|vm).cc$")
```
kasperl, line 21: Add a quick comment to explain why we're dividing the source code up into parts here?

src/heap.h
```diff
@@ -95,6 +95,16 @@ class ObjectHeap : public RawHeap {
   Iterator object_iterator() { return Iterator(_blocks, _program); }
 
   static int max_allocation_size() { return Block::max_payload_size(); }
+
+  inline void do_objects(const std::function<void (HeapObject*)>& func) {
```
kasperl, line 99: Does this have to be inline?

kasperl (approved):
Looks good to me.

PR #622, opened 2022-03-29
Use service for printing

kasperl:


kasperl:
With this change, printing only works if the service discovery mechanism works. This requires the system process to be present. We can make it work without for a while, but over time, we should make it great to run with the system process for everyone.

lib/core/print.toit
```diff
@@ -77,3 +80,24 @@ Does not yield the currently running task.
...
 write_on_stderr_ message/string add_newline/bool -> none:
   #primitive.core.write_string_on_stderr
+
+/**
+Print service used by $print.
+*/
+service_value_/PrintService? := null
+service_mutex_/monitor.Mutex ::= monitor.Mutex
+
+service_ -> PrintService:
```
kasperl, line 90: Is there a better way of doing this? It feels super complex and annoying.

lib/core/print.toit
```diff
@@ -2,14 +2,17 @@
 // Use of this source code is governed by an MIT-style license that can be
 // found in the lib/LICENSE file.
 
+import monitor
+import system.api.print show PrintService PrintServiceClient
+
 /**
 Prints the $message.
 
 /* TODO(florian): the following sentence will probably get stale soon. */
```
floitsch, line 11: Remove the TODO.
Now that the `stringify` is even in the function, it's probably stable.

lib/core/print.toit
floitsch: I think this is what you have to do now...
And this is already an improvement: https://github.com/toitware/toit/issues/1932

lib/core/print.toit
```diff
@@ -77,3 +80,24 @@ Does not yield the currently running task.
...
+service_ -> PrintService:
+  service := service_value_
+  if service: return service
+  return service_mutex_.do:
+    service_value_ = (PrintServiceClient --no-open).open or
+        StandardPrintService_
+
+/**
+Standard print service used when the system print service cannot
+  be resolved.
```
floitsch, line 99:
```suggestion
Standard print service used when the system print service cannot be resolved.
```

floitsch (approved):
LGTM!

PR #629, opened 2022-03-30
Rename GC things that don't make sense.

erikcorry:
Dartino distinguised between the shared heap and the
regular per-process heap, but we just have one heap
so the "shared" makes no sense.

Dartino had generational scavenge (for data) and
non-generational scavenge (for program), but we only
have one scavenge, and it's generational, so no need
to mention that fact.


kasperl (approved):
Looks good to me.

PR #634, opened 2022-03-30
Lazy initialization waits for other tasks to finish.

floitsch:


lib/core/objects.toit
```diff
@@ -187,24 +187,91 @@ class LazyInitializer_:
...
+
+  do_waiting [block]:
+    if id_or_task_ is Task_: return
+    if not id_or_task_ is InitializingTask_: throw "BAD_LAZY_INITIALIZER_CALL"
+    current := (id_or_task_ as InitializingTask_).waiting
+    while current:
+      block.call current.task
+      current = current.next
+
+
```
kasperl, line 208: Drop one newline.

lib/core/objects.toit
```diff
@@ -187,24 +187,91 @@ class LazyInitializer_:
...
+      this_task := task
+      if initializing_task == this_task:
+        // The initializer of the variable is trying to access the global
+        // that is currently initialized.
+        initialization_in_progress_failure_ global
+
+      // Another task is already initializing this global.
+      // Mark us as waiting.
+      initializer.add_waiting this_task
+      next := this_task.suspend_
```
kasperl, line 247: I've got a sneaking suspicion that you have to increment/decrement task_blocked_ (task.toit) around here.

lib/core/objects.toit
```diff
@@ -187,24 +187,91 @@ class LazyInitializer_:
...
-    return result
-  finally:
-    // Either store the computed result, if the initializer succeeded, or
-    // store the original initializer if it failed.
-    __store_global_with_id__ global result
+  while true:
+    if initializer.id_or_task_ is not int:
+      // There is already an initialization in progress.
+      initializing_task := initializer.initializing
+      this_task := task
```
kasperl, line 238: Move this out one level and use when creating the RAM-allocated LazyInitializer?

tests/global_parallel2_test.toit
```diff
@@ -0,0 +1,86 @@
...
+  PARALLEL_COUNT.repeat:
+    task_id := it
+    started_task := monitor.Latch
+    task::
+      started_task.set "started"
+      if task_id < fail_first:
+        expect_throw "try again": read_global.call
+      else:
+        // Implicitly makes sure that the order is correct.
+        expect_equals task_id  read_global.call
```
kasperl, line 61:
```suggestion
        expect_equals task_id read_global.call
```

lib/core/objects.toit
```diff
@@ -187,24 +187,91 @@ class LazyInitializer_:
...
+  do_waiting [block]:
+    if id_or_task_ is Task_: return
+    if not id_or_task_ is InitializingTask_: throw "BAD_LAZY_INITIALIZER_CALL"
+    current := (id_or_task_ as InitializingTask_).waiting
+    while current:
+      block.call current.task
+      current = current.next
+
+
+class InitializingTask_:
```
kasperl, line 209: I would have written this without the two new Task wrappers. Tasks can be chained together themselves while blocked (they have a field for that) and I'd be okay searching for the current task in a chained list to throw the recursive in-progress exception.

So the task in the LazyInitializer becomes the head of a linked list of tasks (the last one isn't blocked before evaluating the initializer). Not sure I care about the order of waking these up, so I'd probably just do it LIFO.

kasperl (approved):
Looks good to me.

lib/core/objects.toit
floitsch: done

lib/core/objects.toit
floitsch: done.

lib/core/objects.toit
floitsch: done.

tests/global_parallel2_test.toit
floitsch: done.

lib/core/objects.toit
floitsch: Tried that and it didn't work. (Spent quite some time to understand why).
The initializing task can run more code, and in particular, it can run code that involves mutexes, ...
Thinking even more, the currently uploaded PR doesn't work either if a task has already another task in the `blocked` list...
Will revert to the initial approach.

floitsch:
After discussion, it looks like we should be able to use the task's blocked list.
This PR is now ready.

kasperl (approved):
Still looks good to me.

PR #636, opened 2022-04-01
Add wifi service

kasperl:


system/extensions/esp32/wifi.toit
```diff
@@ -0,0 +1,201 @@
...
+        state := res.wait
+        res.dispose
+        if (state & WIFI_CONNECTED) != 0:
+          logger_.debug "connected"
+          return
+        else if (state & WIFI_RETRY) != 0:
+          reason ::= wifi_disconnect_reason_ resource
+          logger_.info "retrying" --tags={"reason": reason}
+          wifi_disconnect_ resource_group_ resource
+          sleep --ms=1_000  // Retry with 1s delay.
```
floitsch, line 150: Might as well make this a global next to the other TIMEOUT constants.

floitsch (approved):
LGTM.

system/extensions/esp32/wifi.toit
kasperl: Done.

PR #637, opened 2022-04-01
Add logging service

kasperl:


lib/log/target.toit
```diff
@@ -2,27 +2,57 @@
 // Use of this source code is governed by an MIT-style license that can be
 // found in the lib/LICENSE file.
 
+import bytes
+import monitor
+import system.api.logging show LoggingService LoggingServiceClient
+
 import .level
 
 interface Target:
```
floitsch, line 11: What does the "target" stand for? (why this name?)

floitsch (approved):
LGTM.

lib/log/target.toit
kasperl: Not entirely sure. Maybe we can get rid of the Target interface and just use the LoggingService instead.

PR #641, opened 2022-04-04
Use CoW byte array for blit identity lookup table

kasperl:


erikcorry (approved):

erikcorry:
After offline discussion, replaced with #642

PR #646, opened 2022-04-04
Fix wrong GC loop

erikcorry:


kasperl (approved):
Looks good to me! Thanks :)

PR #651, opened 2022-04-05
Write barrier 2

erikcorry:


src/interpreter.cc
```diff
@@ -134,6 +135,7 @@ Object** Interpreter::gc(Object** sp, bool malloc_failed, int attempts, bool for
     if (VM::current()->scheduler()->is_boot_process(_process)) {
       OS::out_of_memory("Out of memory in system process");
     }
+    GcMetadata::insert_into_remembered_set(_process->task()->stack());
```
kasperl, line 138: Why is this necessary?

src/objects.h
```diff
@@ -386,22 +393,23 @@ class Array : public HeapObject {
     *extra_bytes = 0;
   }
 
-  void fill(int from, Object* filler) {
-    int len = length();
-    for (int index = from; index < len; index++) {
-      at_put(index, filler);
-    }
-  }
+  inline void fill(int from, Object* filler);
```
kasperl, line 396: INLINE or inline? It is probably fine as is.

src/objects.h
```diff
@@ -386,22 +393,23 @@ class Array : public HeapObject {
...
 
  private:
   static const int LENGTH_OFFSET = HeapObject::SIZE;
   static const int HEADER_SIZE = LENGTH_OFFSET + WORD_SIZE;
 
   void _set_length(int value) { _word_at_put(LENGTH_OFFSET, value); }
-  void _initialize(int length, Object* filler) {
+
+  // Can only be called on newly allocated objects that will be either
```
kasperl, line 404: Can this be asserted easily?

src/objects.h
```diff
@@ -1154,10 +1162,18 @@ class Instance : public HeapObject {
     return _at(_offset_from(index));
   }
 
-  void at_put(int index, Object* value) {
+  inline void at_put(int index, Smi* value) {
```
kasperl, line 1165: This inline seems unnecessary. It is implied. Did you mean INLINE?

src/top.h
```diff
@@ -27,7 +27,7 @@
 #endif
 
 // Use semispace non-generational GC.
-#define LEGACY_GC 1
+//#define LEGACY_GC 1
```
kasperl, line 30: Are we ready for this?

kasperl (approved):
Looks good to me.

src/objects.h
erikcorry: The function is also called from the compiler where there are no write barriers, so no.  I fixed the comment.

PR #652, opened 2022-04-06
Stack trace origin

kasperl:


tests/negative/gold/illegal_system_call_test.gold
```diff
@@ -0,0 +1,9 @@
+As check failed: null is not a ServiceResource.
```
kasperl, line 1: Before this change, this test would have generated a complete nonsense stack trace.

src/compiler/toitc.cc
```diff
@@ -98,225 +76,211 @@ int main(int argc, char **argv) {
...
-    }
-    char* image_filename = argv[2];
-    char* bundle_filename = argv[3];
-    auto bundle = SnapshotBundle::read_from_file(bundle_filename);
-    if (!bundle.is_valid()) print_usage(1);
-    write_image_from_bundle(image_filename, bundle);
-    free(bundle.buffer());
-  } else {
-    char* bundle_filename = null;
+  char* bundle_filename = null;
```
kasperl, line 79: Only indentation changed below here.

tests/containers_test.toit
```diff
@@ -7,7 +7,8 @@ import expect show *
 
 main:
   images/List := containers.images
-  expect_list_equals [containers.current] images
+  expect_equals 2 images.size
```
kasperl, line 10: Not really sure this is the behavior I want, but for now we do return the system image in the list of containers.

floitsch (approved):
LGTM.

PR #654, opened 2022-04-06
Don't leave uninitialized objects on the heap

erikcorry:
Be a bit more defensive about having uninitialized objects on the heap, that can't be traversed.

With a GC stress test that eagerly puts random objects in the old space we would see
junk in arrays allocated by messaging.cc, then abandoned when deserialization was
restarted after an allocation failure.


kasperl (approved):
Looks good to me.

PR #661, opened 2022-04-12
Fix allocation stats in new GC.

erikcorry:
Also deprecates a util that has a poor (non-reentrant) API.

Improves process_stats:
* You can now pass a list to avoid allocation.
* Handles Smi overflow better.
* Somewhat futureproofed if more stats are added later.
* Doesn't return a bare Array to user code any more.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@
 
 # General options.
 HOST=host
-BUILD_TYPE=Debug
+BUILD_TYPE=Release
 
 # Use 'make ESP32_ENTRY=examples/mandelbrot.toit esp32' to compile a different
 # example for the ESP32 firmware.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -18,7 +18,7 @@
 
 # General options.
 HOST=host
-BUILD_TYPE=Release
```
floitsch, line 21: I think this should be reverted.

lib/core/utils.toit
```diff
@@ -286,18 +290,16 @@ Variant of $(process_stats).
...
-process_stats_ group id:
-  #primitive.core.process_stats:
-    return null
+process_stats_ list group id:
+  #primitive.core.process_stats
 
 /**
 Returns the number of bytes allocated, since the last call to this function.
 For the first call, returns number of allocated bytes since system start.
+#Deprecated, because it doesn't nest.  Use $process_stats instead.
```
floitsch, line 302:
```suggestion

Deprecated. This function doesn't nest.  Use $process_stats instead.
```

floitsch (approved):
LGTM.

PR #671, opened 2022-04-22
Return leading bytes if buffer is too small

floitsch:
If the RMT ring-buffer has too much data, only copy as much as we
want/can handle. If the user wants to know whether there was more data,
they should just request data with a higher `max_output_len` and
look at the size of the returned bytearray.


kasperl (approved):
Looks good to me.

PR #674, opened 2022-04-25
add(lib): function to check string length

snxx-lppxx:


floitsch:
We generally don't add global functions, when a method works.
In this case, the string already has a `.size` so adding an additional `len` function isn't necessary. It would also make it hard for new users to know which one to use.

You could write a package with it, but for "typical" Toit code I recommend `.size` as this will make it easier for other Toit developers to understand your code.

PR #675, opened 2022-04-25
Add .editorconfig and avoid using eol=lf for source files

snxx-lppxx:


.gitattributes
```diff
@@ -1,8 +1,10 @@
+* text=auto eol=lf
+
 # Source files should not be normalized to avoid confusing clang-format-diff.py
-*.c text eol=lf
-*.cc text eol=lf
-*.cpp text eol=lf
-*.h text eol=lf
+*.c -diff
```
floitsch, line 4: According to the documentation this would "[...] cause Git to generate Binary files differ (or a binary patch, if binary patches are enabled) instead of a regular diff."

That's not what we want.
We just want git not to change the line endings.

.gitattributes
```diff
@@ -1,8 +1,10 @@
+* text=auto eol=lf
```
floitsch, line 1: We actually want Git to change the line-endings for Toit files, so that we have test-coverage for Windows line-endings on the build bots.
We won't be able to prevent users to use non-lf line endings on Windows, so we need to test it too.

floitsch (changes requested):
Thanks.
I will take your editorconfig idea and adapt it for the template.

.editorconfig
```diff
@@ -0,0 +1,13 @@
+# Syntax: editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 4
```
floitsch, line 8:
```suggestion
indent_size = 2
```

kasperl:
Closing due to inactivity.

PR #679, opened 2022-04-25
Share second semispace between all processes

erikcorry:


src/third_party/dartino/two_space_heap.h
```diff
@@ -136,17 +132,27 @@ class TwoSpaceHeap {
...
         record_(&dummy_record_),
         water_mark_(heap->water_mark_) {}
 
+  SemiSpace* to_space() { return &to_; }
+
+  void complete_scavenge() {
+    bool work_found = true;
+    while (work_found) {
+      work_found = to_.complete_scavenge(this);
+      work_found |= old_->complete_scavenge(this);
```
kasperl, line 152: It is a little subtle that this wouldn't work with ||= even when you're using bools. Maybe it would be simpler to just write:

```
if (old_->complete_scavenge(this)) work_found = true;
```

kasperl (approved):
Looks good to me.

PR #680, opened 2022-04-25
Avoid creating an old-space if it is not needed.

erikcorry:
If you use set_max_heap_size_ to set the max heap
to a single page (4k on embedded, 32k on desktop/server)
then no old space will be created, and the heap of the
process will be limited to one page, even if some objects
are long lived.

If the max heap size never exceeds a half page (2k or 16k)
then we also create no old-space.

### DIFF
--- a/src/third_party/dartino/two_space_heap.cc
+++ b/src/third_party/dartino/two_space_heap.cc
@@ -25,7 +25,6 @@ TwoSpaceHeap::TwoSpaceHeap(Program* program, ObjectHeap* process_heap, Chunk* ch
 uword TwoSpaceHeap::max_expansion() {
   if (!process_heap_->has_max_heap_size()) return UNLIMITED_EXPANSION;
   uword limit = process_heap_->limit();
-  printf("limit %dbytes\n", (int)limit);
   if (limit <= TOIT_PAGE_SIZE) return 0;
   limit -= TOIT_PAGE_SIZE;  // New space is one page.
   if (limit < old_space()->used()) return 0;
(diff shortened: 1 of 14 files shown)


src/third_party/dartino/two_space_heap.cc
```diff
@@ -53,6 +55,18 @@ HeapObject* TwoSpaceHeap::allocate(uword size) {
 
 void TwoSpaceHeap::swap_semi_spaces(SemiSpace& from, SemiSpace& to) {
   water_mark_ = to.top();
+  if (old_space()->is_empty() && to.used() < TOIT_PAGE_SIZE / 2) {
+    // Don't start promoting to old space until a the post GC heap size
```
kasperl, line 59: a the -> the

kasperl (approved):
Looks good to me.

src/third_party/dartino/two_space_heap.cc
```diff
@@ -20,12 +20,14 @@ TwoSpaceHeap::TwoSpaceHeap(Program* program, ObjectHeap* process_heap, Chunk* ch
...
       semi_space_(program, chunk) {
   semi_space_size_ = TOIT_PAGE_SIZE;
-  max_size_ = 256 * TOIT_PAGE_SIZE;
 }
 
 uword TwoSpaceHeap::max_expansion() {
+  if (!process_heap_->has_max_heap_size()) return UNLIMITED_EXPANSION;
   uword limit = process_heap_->limit();
-  if (limit == 0) return UNLIMITED_EXPANSION;
+  printf("limit %dbytes\n", (int)limit);
```
kasperl, line 28: Drop the debug printing.

PR #681, opened 2022-04-25
Further reduce the stack size for OTA updates.

erikcorry:
This change moves the SHA256 state onto the malloc heap,
instead of having it on top of the stack during OTA
commit.

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -403,4 +403,14 @@ struct Defer {
   ~Defer() { fun(); }
 };
 
+template <typename T>
+class DeferDelete {
+ public:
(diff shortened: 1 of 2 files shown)


src/utils.h
```diff
@@ -403,4 +403,14 @@ struct Defer {
   ~Defer() { fun(); }
 };
 
+template <typename T>
+class DeferDelete {
+ public:
+  DeferDelete(T* obj) : _obj(obj) {}
```
kasperl, line 409: I would spell object out.

kasperl (approved):
Looks good to me.

PR #684, opened 2022-04-27
Preliminary SPI support for Linux

kasperl:


kasperl:
I think I will go ahead and put this into a new primitive module instead of reusing 'spi'.

floitsch (approved):
LGTM.

PR #688, opened 2022-04-28
Misc new-GC changes.

erikcorry:
Compiles in debug mode on ESP32 (but doesn't fit in IRAM).

Fix missing initialization of water_mark_ in scavenge.

Fix spurious out-of-memory caused by reluctance to open old
space (didn't allow allocation in old space during scavenge).

Don't always do a mark-sweep right after the first old-space
promotion (initialize with an allocation budget of one page).

Fix missing initialization of new-space mark bits.

Remove unused initialize() function on TwoSpaceHeap.

### DIFF
--- a/src/os_linux.cc
+++ b/src/os_linux.cc
@@ -75,7 +75,7 @@ bool OS::use_virtual_memory(void* addr, uword sz) {
   // Calls to use_virtual_memory are rounded up by one due to the single-word
   // object problem, but we don't want to poison data belonging to the next
   // page's metadata.
-  if (sz != 0) memset(addr, 0xc1, sz - 1);
+  memset(addr, 0xc1, sz - 1);
 #endif
   if (result == 0) return true;
   if (errno == ENOMEM) return false;
--- a/src/os_linux.cc
+++ b/src/os_linux.cc
@@ -71,6 +71,12 @@ bool OS::use_virtual_memory(void* addr, uword sz) {
   uword rounded = Utils::round_down(address, getpagesize());
   uword size = Utils::round_up(end - rounded, getpagesize());
   int result = mprotect(reinterpret_cast<void*>(rounded), size, PROT_READ | PROT_WRITE);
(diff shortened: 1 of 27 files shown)


src/os_linux.cc
```diff
@@ -71,6 +71,12 @@ bool OS::use_virtual_memory(void* addr, uword sz) {
   uword rounded = Utils::round_down(address, getpagesize());
   uword size = Utils::round_up(end - rounded, getpagesize());
   int result = mprotect(reinterpret_cast<void*>(rounded), size, PROT_READ | PROT_WRITE);
+#ifdef DEBUG
+  // Calls to use_virtual_memory are rounded up by one due to the single-word
+  // object problem, but we don't want to poison data belonging to the next
+  // page's metadata.
+  if (sz != 0) memset(addr, 0xc1, sz - 1);
```
kasperl, line 78: How can `sz` be zero here?

src/third_party/dartino/object_memory.cc
```diff
@@ -27,6 +27,10 @@ Chunk::Chunk(Space* owner, uword start, uword size, bool external)
         scavenge_pointer_(start_) {
   if (GcMetadata::in_metadata_range(start)) {
     GcMetadata::initialize_overflow_bits_for_chunk(this);
```
kasperl, line 29: It feels like you should share the code for the initialization of chunks. You have two different versions of this (one with 2 initialization calls and one with 3). Maybe add a shared initializer function that takes a parameter instead?

src/third_party/dartino/object_memory_copying.cc
```diff
@@ -35,6 +32,8 @@ SemiSpace::SemiSpace(Program* program, Chunk* chunk)
   ASSERT(chunk);
```
kasperl, line 32: Seems pretty useless.

kasperl (approved):
Looks good to me.

PR #690, opened 2022-04-28
Fix Makefile and Cmake file for image dependencies

floitsch:
Using the environment variable to set an important parameter of the
cmake file doesn't work, as cmake rebuilds itself when it detects that
its files have changed. At that point the environment isn't set and it
would generate cache files with the wrong arguments.


kasperl (approved):
Looks good to me!

PR #691, opened 2022-04-28
Automatically source export.sh

floitsch:
When the cross-compiler isn't found, automatically source export.sh.


kasperl (approved):
Looks good to me.

PR #693, opened 2022-04-29
Rename 'signal_X' to just 'X'.

floitsch:
Also change the order of the arguments to the `Signal.do` block.


floitsch:
Note that this changes the order of `level` and `period` in the `do` call.
Personally, I found having the `level` first just much more intuitive.

kasperl (approved):
Looks good to me.

PR #699, opened 2022-05-03
Share code for receive and transmit_and_receive.

floitsch:


src/resources/rmt_esp32.cc
```diff
@@ -362,6 +333,44 @@ PRIMITIVE(transmit_and_receive) {
...
+  int receive_items_length = receive_bytes.length() / 4;
+
+  return transmit_and_receive(tx_channel, rx_channel,
+                              transmit_items, transmit_items_length,
+                              receive_items, receive_items_length,
+                              max_output_len,
+                              receive_timeout_ms,
+                              process);
+}
+
```
kasperl, line 373: Drop extra newline.

src/resources/rmt_esp32.cc
```diff
@@ -362,6 +333,44 @@ PRIMITIVE(transmit_and_receive) {
...
+  ARGS(RMTResource, tx, RMTResource, rx, Blob, transmit_bytes, Blob, receive_bytes, int, max_output_len, int, receive_timeout_ms)
+  if (transmit_bytes.length() % 4 != 0) INVALID_ARGUMENT;
+  if (receive_bytes.length() % 4 != 0) INVALID_ARGUMENT;
+
+  rmt_channel_t rx_channel = rx->channel();
+  rmt_channel_t tx_channel = tx->channel();
+
+  const rmt_item32_t* transmit_items = reinterpret_cast<const rmt_item32_t*>(transmit_bytes.address());
+  const rmt_item32_t* receive_items = reinterpret_cast<const rmt_item32_t*>(receive_bytes.address());
+  int transmit_items_length = transmit_bytes.length() / 4;
```
kasperl, line 363: Should you pass null items if the lengths are zero?

kasperl (approved):
Looks good to me.

src/resources/rmt_esp32.cc
floitsch: Good find.
Switched to checking for the length in the shared function instead, this way I avoid having a check here.

src/resources/rmt_esp32.cc
floitsch: done.

PR #700, opened 2022-05-03
Rename configure to config.

floitsch:
This makes the name more consistent with the `Pin` method.


floitsch:
Abandoning this PR.

Turns out we already use `configure` in some other places. (bma2x2 cellular drivers, ...)

PR #706, opened 2022-05-04
Fix data race when marking on two threads at once

erikcorry:
No discernable performance difference on Linux (if anything a little faster).


kasperl (approved):
Looks good to me.

PR #718, opened 2022-05-05
Activate new GC

erikcorry:


src/top.h
```diff
@@ -27,7 +27,7 @@
 #endif
 
 // Use semispace non-generational GC.
```
kasperl, line 29: Update the comment!

kasperl (approved):
Looks good to me. Ship it!

PR #732, opened 2022-05-12
Work around spurious timer notifications by re-arming timer

kasperl:


lib/core/monitor_impl.toit
```diff
@@ -49,7 +49,7 @@ class __Monitor__:
     if deadline:
       timer = self.acquire_timer_
       // Arrange for the notify_ method to be called if the timeout expires.
-      timer.arm this deadline
+      timer.set_target this
```
erikcorry: Should this be moved into acquire_timer_ so you can't forget it?

lib/core/monitor_impl.toit
```diff
@@ -88,6 +88,7 @@ class __Monitor__:
           if deadline and Time.monotonic_us >= deadline: return false
         // Wait until notified. When we get back the monitor might be owned by
         // someone else.
+        if timer: timer.arm deadline
         await_ self
     finally:
       if timer: self.release_timer_ timer
```
erikcorry, line 93: Perhaps move the null check inside release_timer_ ?

erikcorry (approved):

lib/core/monitor_impl.toit
kasperl: Done.

lib/core/monitor_impl.toit
kasperl: Keeping as is for consistency with the other 'if timer' checks.

PR #735, opened 2022-05-16
Fix bug in GC in debug mode that could crash.

erikcorry:
In debug mode we accidentally overwrote data that we needed
in the free areas of the heap.


kasperl (approved):
Looks good to me, but I'd support dropping the DEBUG-mode only extra function.

erikcorry:
Superseded by a similar change.

PR #737, opened 2022-05-16
Roll esp-idf to get heap range directly from malloc.

erikcorry:
Also rolls in fix to malloc heap stats.


kasperl (approved):
Looks good to me.

PR #743, opened 2022-05-18
tls_test_slow be less picky about the error message

erikcorry:
The important thing is that wrong sites fail to connect.
By not specifying the exact error message we make the test
less flaky.


floitsch (approved):
LGTM.

PR #748, opened 2022-05-19
Print more info if heap is allocated outside expected range

erikcorry:


kasperl (approved):
Looks good to me. I'd love to get rid of the debug printing at the start again at some point, but for now this is excellent.

PR #749, opened 2022-05-20
Removing extra thread and allow deinit of BLE

mikkeldamsgaard:
The BLE implementation seems to have spun up an additional thread that was not terminated correctly on shutdown. This PR removes that extra threads and moves the initialisation around to make it work with the one thread.

### DIFF
--- a/src/event_sources/ble_esp32.cc
+++ b/src/event_sources/ble_esp32.cc
@@ -50,7 +50,7 @@ bool BLEEventSource::start() {
   ASSERT(_resources_changed == null);
   _resources_changed = OS::allocate_condition_variable(mutex());
   if (_resources_changed == null) return false;
-  if (!spawn()) {
+  if (!spawn(NIMBLE_STACK_SIZE)) {
     OS::dispose(_resources_changed);
     _resources_changed = null;
     return false;
@@ -87,7 +87,6 @@ void BLEEventSource::entry() {
         nimble_port_run();
       }
 
-      nimble_port_freertos_deinit();
       _running = false;
       OS::signal(_resources_changed);
     }
--- a/src/event_sources/ble_esp32.cc
+++ b/src/event_sources/ble_esp32.cc
@@ -36,7 +36,7 @@ BLEEventSource* BLEEventSource::_instance = null;
 
 BLEEventSource::BLEEventSource()
     : LazyEventSource("BLE", 1)
-    , Thread("BLE") {
+    , Thread("BLEEventSource") {
   _instance = this;
 }
 
@@ -79,8 +79,6 @@ void BLEEventSource::entry() {
 
   while (!_stop) {
     if (_should_run) {
-      nimble_port_init();
-
       _running = true;
       OS::signal(_resources_changed);
 
--- a/src/resources/ble_esp32.cc
+++ b/src/resources/ble_esp32.cc
@@ -183,13 +183,12 @@ class BLEResourceGroup : public ResourceGroup {
 };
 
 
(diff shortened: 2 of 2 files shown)


src/resources/ble_esp32.cc
```diff
@@ -414,10 +409,6 @@ int BLEResourceGroup::init_server() {
     }
 
     ble_hs_cfg.reset_cb = ble_on_reset;
-    // Start the host thread.
-    if (!_server_config->spawn(NIMBLE_STACK_SIZE)) {
```
kasperl, line 418: Is the NIMBLE_STACK_SIZE constant now unused? Should it be passed to the BLEEventSource thread constructor?

src/resources/ble_esp32.cc
```diff
@@ -546,8 +533,7 @@ PRIMITIVE(init) {
 
   ble_hs_cfg.sync_cb = ble_on_sync;
 
-  group->register_resource(gap);
-  group->set_gap(gap);
+  nimble_port_init();
```
kasperl, line 536: Should we deinit if setting the server config fails?

src/event_sources/ble_esp32.cc
```diff
@@ -79,8 +79,6 @@ void BLEEventSource::entry() {
 
   while (!_stop) {
     if (_should_run) {
-      nimble_port_init();
```
kasperl, line 82: I'd prefer keeping the nimble_port_init call here, because of the symmetry with the deinit call below. Is there a reason you moved this?

src/resources/ble_esp32.cc
kasperl, line 536: Does the BLEEventSource thread start running (_should_run == true) at the BLEEventSource::instance() call? If so, It feels a bit weird (and potentially riddled with race conditions) that we're not done with the configuration and we haven't called init yet, but we still let it start.

kasperl (approved):

src/event_sources/ble_esp32.cc
mikkeldamsgaard: I see your point, however, unfortunately, the gatt server setup requires nimble_port_init to be called before it sets up its resources (init_server). I'll add a comment.

src/resources/ble_esp32.cc
mikkeldamsgaard: Yes. Good point.

src/resources/ble_esp32.cc
mikkeldamsgaard: The BLEResourceGroup destructor does that for us.

src/event_sources/ble_esp32.cc
mikkeldamsgaard: Actually, looking at it, the nimble_port_freertos_deinit is not needed at all, as we are not running the freertos task, but our own thread. So, I'll go ahead and remote the deinit call and symmetry is restored

src/resources/ble_esp32.cc
mikkeldamsgaard: No. It starts when a GAP resource is registered. That is why the register_resource call is moved after the server init.

kasperl: Got it! Maybe add that as a comment just above the call to `group->register_resource(gap)`?

kasperl (approved):
Still looks good!

kasperl:
Trying to land this through https://github.com/toitlang/toit/tree/ble-fixes.

kasperl:
Closing in favour of #761.

PR #757, opened 2022-05-30
Xtensa does not have popcount, this is faster

erikcorry:
popcount is used in compact and sweep.  This speeds up compact by about 10%.

I also tried terminating the loop early if u was zero, but that was slower than this.

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -81,7 +81,7 @@ class Utils {
     }
   }
 
-  static uint8 popcount_table[256];
+  static const uint8 popcount_table[256];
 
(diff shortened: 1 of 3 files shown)


src/utils.h
```diff
@@ -81,10 +81,21 @@ class Utils {
     }
   }
 
+  static uint8 popcount_table[256];
```
kasperl, line 84: Should this be const? It feels weird that it is mutable.

kasperl (approved):
Looks good to me. A test of the table-based popcount would be great.

PR #758, opened 2022-05-30
Improve latch

floitsch:
Makes it possible to see whether a latch has already a value.

Also, we now allow to read the value multiple times.


kasperl (approved):
Looks good to me.

PR #761, opened 2022-05-31
Remove extra thread and allow deinit of BLE

kasperl:
All the work done by @mikkeldamsgaard in #749.


src/event_sources/ble_esp32.cc
```diff
@@ -89,7 +87,6 @@ void BLEEventSource::entry() {
         nimble_port_run();
       }
 
-      nimble_port_freertos_deinit();
```
floitsch, line 92: Is this not needed anymore?

src/resources/ble_esp32.cc
```diff
@@ -559,6 +546,9 @@ PRIMITIVE(init) {
     }
   }
 
+  group->register_resource(gap);
```
floitsch, line 549: If the `gap` is only registered here, do we then need to deallocate it explicitly when the init-server fails (line 544/545)?

floitsch (approved):
LGTM.

src/resources/ble_esp32.cc
kasperl: Yeah. I believe we do need to `delete gap` before returning in line 545. Thoughts, @mikkeldamsgaard?

src/event_sources/ble_esp32.cc
kasperl: Yeah, it turns out that this isn't needed.

PR #762, opened 2022-05-31
Provide usage message when arguments fail to parse

erikcorry:
Sample output:

```
$ toit.run cli.toit  
Usage:
<toit.run cli.toit [--value|-v=<value>] [--flag] install  <app-name> <image-file>
<toit.run cli.toit [--value|-v=<value>] [--flag] uninstall  <app-name>
<toit.run cli.toit [--value|-v=<value>] [--flag] set-max-offline  <offline-time-in-seconds>
$ toit.run cli.toit install
Too few arguments
Usage:
toit.run cli.toit install  <app-name> <image-file>
```

### DIFF
--- a/src/flags.h
+++ b/src/flags.h
@@ -81,6 +81,8 @@ class Flags {
  public:
   FLAGS_DO(DECLARE_DEBUG_FLAG, DECLARE_DEPLOY_FLAG)
 
+  static const char* command;
+
 #ifndef IOT_DEVICE
   static int process_args(int* argc, char** argv);
 #endif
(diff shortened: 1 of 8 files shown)


src/flags.h
```diff
@@ -81,6 +81,8 @@ class Flags {
  public:
   FLAGS_DO(DECLARE_DEBUG_FLAG, DECLARE_DEPLOY_FLAG)
 
+  static const char* command;
```
kasperl, line 84: Maybe just call this program_name?

kasperl (approved):
Looks good to me.

PR #763, opened 2022-05-31
Allow the use of ADC2 with a flag.

floitsch:


floitsch:
I will write a test, if you agree with this approach.

src/resources/adc_esp32.cc
```diff
@@ -58,6 +55,75 @@ static int get_adc2_channel(int pin) {
...
+    case 20: return ADC2_CHANNEL_9;
+    default: return adc2_channel_t(-1);
+  }
+}
+
+#elif CONFIG_IDF_TARGET_ESP32C3
+
+static int get_adc1_channel(int pin) {
+  switch (pin) {
+    case 0: return ADC1_CHANNEL_0;
```
floitsch, line 96: Note that the pin numbers for the ESP32C2 were wrong.

lib/gpio/adc.toit
```diff
@@ -42,12 +44,18 @@ class Adc:
   Note that chip-specific limitations apply, generally the precision at
     various voltage ranges.
 
-  On the ESP32, only the ADC1 (pins 32-39) is supported. The ADC2 has too many
-    restrictions (cannot be used when WiFi is active, and some of the pins are
-    strapping pins), and is therefore disabled.
+  If $allow_dangerous is provided, actives ADC pins that are dangerous to use.
```
kasperl, line 47: Is it really dangerous or just unreliable, unstable, restricted, special-purpose?

kasperl (approved):
Looks good to me.

floitsch:
Still need to get the test running...

PR #765, opened 2022-05-31
Remove warning in debug mode.

floitsch:
The warning correctly sees that we are using a pointer that might not
exist anymore. But this is just for debugging where we want to see
whether the pointer has changed.


kasperl (approved):
Looks good to me. Maybe add a comment to explain why.

floitsch:
As discussed, switched to storing the pointer in a word.
Also added a comment.

PR #766, opened 2022-05-31
Remove some heap code no longer in use

erikcorry:


src/linked.h
```diff
@@ -562,6 +563,43 @@ class DoubleLinkedList {
     Element* _entry;
   };
 
+  class ConstIterator {
+   public:
+    explicit ConstIterator(const Element* entry)
+        : _entry(entry) {}
+
+    inline const T* operator->() {
```
kasperl, line 571: I wouldn't typically put inline on all these.

kasperl (approved):
Looks good to me.

PR #767, opened 2022-06-03
Log trace

mikkeldamsgaard:
Adding trace to log messages, to prepare for allowing a loggingservice to send fatal traces remotely.
Also changed the system boot to use logging when a system mirror message is received.


kasperl:
FYI @mikkeldamsgaard: I plan to start landing this in some chunks over the next days, once I've gotten the cellular drivers to pop up as proper network drivers. This way, stack traces can be transmitted over cellular!

kasperl:
I've looked at this for a while now and I've decided to come up with an alternative approach: #1019. I'd like to avoid forcing everyone who hooks into the log service framework to have to deal with encoded stack traces and the likes.

Let me know what you think, @mikkeldamsgaard!

PR #772, opened 2022-06-07
Work around 'unexpected end of file' build error on macOS

kasperl:


kasperl:
TBR'ed.

Makefile
```diff
@@ -173,9 +173,7 @@ endif
 
 .PHONY: esp32
 esp32:
-	if [ "$(shell command -v xtensa-esp32-elf-g++)" == "" ]; then
-		source $(IDF_PATH)/export.sh;
-	fi;
+	if [ "$(shell command -v xtensa-esp32-elf-g++)" = "" ]; then source $(IDF_PATH)/export.sh; fi
```
floitsch, line 176: as commented here: https://github.com/toitlang/toit/commit/f6159ae45aba711738dbe2f98bef451f015cb733

that shouldn't be necessary with a recent GNU Make.

PR #773, opened 2022-06-07
Add ble.RemoteCharacteristic.write_value

kasperl:


src/resources/ble_esp32.cc
```diff
@@ -876,6 +876,21 @@ PRIMITIVE(request_data) {
   }
 }
 
+PRIMITIVE(send_data) {
+  ARGS(GATTResource, gatt, uint16, handle, Object, value);
+
+  os_mbuf* om = null;
+  Object* error = object_to_mbuf(process, value, &om);
```
floitsch, line 883: The `object_to_mbuf` function seems to convert null pointers to null pointers.
I don't think the `ble_gattc_write` function can handle null pointers.

floitsch (approved):
LGTM.

src/resources/ble_esp32.cc
kasperl: Yeah, that is a valid point. I'm trying to reuse the functionality to get the string/byte-array conversion done the same way as elsewhere in the BLE code. I'll probably keep it as is for now.

PR #775, opened 2022-06-07
Add hash_code method to ByteArray

erikcorry:


lib/core/collections.toit
```diff
@@ -1263,6 +1268,12 @@ abstract class ByteArrayBase_ implements ByteArray:
     if other is not ByteArray: return false
     #primitive.core.blob_equals
 
+  /**
+  Returns a hash code that depends on the content of this ByteArray.
+  */
+  hash_code -> int:
+    #primitive.core.string_slice_hash_code
```
floitsch, line 1275: We should probably rename this primitive.
I would probably not do it in this PR, but add a comment to the primitive to explain that it is also used for byte arrays (despite the name).

floitsch (approved):
LGTM.

PR #778, opened 2022-06-08
Improve BLE error message.

floitsch:
When the BLE device is already in use, it used to throw an
OUT_OF_BOUNDS. Now the error is "ALREADY_IN_USE".


kasperl (approved):
Looks good to me.

PR #781, opened 2022-06-08
Work around gcc optimization.

floitsch:
Gcc somehow found a way to remove a zero check for smi pointers.


kasperl:
Is this obsolete?

PR #782, opened 2022-06-08
Change `is_type` checks of heap objects.

floitsch:
The type tests must not be instance methods, as the compiler is allowed
to assume that instance calls were never done on null-pointers.

If we have a Smi object of value 0, then `o->is_X` lets the compiler
assume that `o` is not equal to `null` (or 0). If we then want to use
`o` as a 0-smi (for example by doing a value check:
`if (Smi::cast(o)->value() == 0)` then the compiler optimizes the check
away as it "knows" that `o` can't be null.


kasperl (approved):
Looks good to me.

PR #783, opened 2022-06-08
Avoid left shift of negative number.

floitsch:


kasperl (approved):
Looks good to me.

erikcorry:
Interesting that C++20 fixed this, and it's OK to shift signed numbers "now".
https://en.cppreference.com/w/cpp/language/operator_arithmetic

PR #784, opened 2022-06-13
LSP: Exit loop when there is an error.

floitsch:
Once there was an issue (like "file already closed"), the server would
go into an infinite loop, wasting CPU.


kasperl (approved):
Looks good to me.

PR #786, opened 2022-06-13
Add more duration constants.

floitsch:


lib/core/time.toit
```diff
@@ -53,6 +53,27 @@ class Duration implements Comparable:
   /** The number of nanoseconds per hour. */
   static NANOSECONDS_PER_HOUR        ::= 3600_000_000_000
 
+  /** The number of microseconds per millisecond. */
+  static MICROSECONDS_PER_MILLISECOND ::=        1_000
```
kasperl, line 57: Should these constants have int types?

kasperl (approved):
Looks good to me.

lib/core/time.toit
floitsch: The compiler should eventually figure it out itself, but for now adding the type is better.
done.

PR #787, opened 2022-06-13
Make the ADC non-blocking

floitsch:


lib/gpio/adc.toit
```diff
@@ -61,7 +63,20 @@ class Adc:
   Measures the voltage on the Pin.
   */
   get --samples=64 -> float:
-    return adc_get_ state_ samples
+    if samples < 1: throw "OUT_OF_BOUNDS"
+    if samples <= MAX_SAMPLES_PER_CALL_: return adc_get_ state_ samples
+    // Sample in chunks of 64, so we don't spend too much time in
+    // the primitive.
+    sampled := 0
```
kasperl, line 70: I'd move this down to just before the loop (even after the result definition).

kasperl (approved):
Looks good to me.

lib/gpio/adc.toit
floitsch: done.

PR #788, opened 2022-06-14
Add sha256 sum to snapshot

erikcorry:


src/snapshot_bundle.cc
```diff
@@ -24,6 +25,7 @@ namespace toit {
 
 static const char* const MAGIC_NAME = "toit";
 static const char* const MAGIC_CONTENT = "like a tiger";
+static const char* const SHA_NAME = "sha256";
```
kasperl, line 28: Let's turn this into uuid id, so all other tools can use the same and it fits into Program::Header.

src/snapshot_bundle.cc
```diff
@@ -53,7 +55,15 @@ SnapshotBundle::SnapshotBundle(List<uint8> snapshot,
     if (status != 0) FATAL("Couldn't create snapshot");
   };
 
+  mbedtls_sha256_context sha_context;
+  mbedtls_sha256_init(&sha_context);
+  mbedtls_sha256_update_ret(&sha_context, snapshot.data(), snapshot.length());
```
kasperl, line 60: I'd argue that we need the source map in here too.

src/snapshot_bundle.cc
```diff
@@ -53,7 +56,30 @@ SnapshotBundle::SnapshotBundle(List<uint8> snapshot,
...
+  uint8 length_bytes[sizeof(uint32)];
+  Utils::write_unaligned_uint32_le(length_bytes, snapshot.length());
+  mbedtls_sha256_update_ret(&sha_context, length_bytes, sizeof(uint32));
+  mbedtls_sha256_update_ret(&sha_context, snapshot.data(), snapshot.length());
+  mbedtls_sha256_update_ret(&sha_context, source_map_data.data(), source_map_data.length());
+  uint8 sum[SHA256_HASH_LENGTH];
+  mbedtls_sha256_finish_ret(&sha_context, sum);
+  mbedtls_sha256_free(&sha_context);
+
+  // Fix checksum bytes to make a legal UUID.
```
kasperl, line 77:
```suggestion
  // Fix checksum bytes to make a legal UUID5.
```

kasperl (approved):
Looks good to me!

PR #792, opened 2022-06-14
Bundle multiple images with the VM and run them on startup

kasperl:


tools/snapshot_to_image.toit
```diff
@@ -126,18 +87,16 @@ print_usage parser/arguments.ArgumentParser:
 
 main args:
   parser := arguments.ArgumentParser
-  parser.describe_rest ["snapshot-file"]
+  parser.describe_rest --max=arguments.UNLIMITED ["snapshot-files"]
```
erikcorry, line 90: The behaviour is exactly the same, but you could also do this as

```
parser.describe_rest ["snapshot-files", "..."]
```

erikcorry (approved):

floitsch (commented):
LGTM.

PR #794, opened 2022-06-15
Use dkhostmaster.dk instead of oracle.com for TLS test

erikcorry:
For make glorious stability of test!


kasperl (approved):
Looks good to me.

PR #805, opened 2022-06-20
Histogram decode

erikcorry:


src/primitive_debug.cc
```diff
@@ -60,9 +54,40 @@ PRIMITIVE(object_histogram) {
...
+  uword bytes_needed = 0;
+  for (int i = 0; i < length; i++) {
+    if (data[i].size > 0) {
+      non_trivial_entries++;
+      bytes_needed += Encoder::bytes_to_encode(i);
+      bytes_needed += Encoder::bytes_to_encode(data[i].count);
+      bytes_needed += Encoder::bytes_to_encode(data[i].size);
+    }
+  }
+  MallocedBuffer encoding_buffer(100 + bytes_needed);
```
kasperl, line 70: 100?

tools/mirror.toit
```diff
@@ -329,6 +329,45 @@ class Profile extends Mirror:
...
+        entries.add
+          HistogramEntry class_name json[i + 1] json[i + 2]
+    entries.sort --in_place: | a b | b.size - a.size
+    super json program
+
+  table:
+    return entries.join "\n"
+
+  stringify -> string:
+    return "Object heap Histogram:\n"
```
kasperl, line 363:
```suggestion
    return "Object heap histogram:\n"
```

lib/core/utils.toit
```diff
@@ -440,11 +440,19 @@ Converts a number between 0 and 15 to an upper case
...
+  requirements.  The histogram is sent as a system
+  mirror message, which means it is usually printed on
+  the console.
+*/
 print_objects marker/string="" gc/bool=true:
   if gc:
     before := gc_count
     while gc_count == before: RecognizableFiller_
-  print_histogram_ marker object_histogram_
+  encoded_histogram := object_histogram_
```
kasperl, line 453: So the marker is just ignored now? It is pretty useful to understand where a specific histogram came from.

src/primitive_debug.cc
```diff
@@ -22,34 +22,28 @@ namespace toit {
...
+struct PerClass {
+  uint32 count;
+  uint32 size;
+};
+
 PRIMITIVE(object_histogram) {
   static const int UINT32_PER_ENTRY = 2;
   Program* program = process->program();
   int length = program->class_bits.length();
   int size = length * UINT32_PER_ENTRY * sizeof(uint32);
```
kasperl, line 34: Use sizeof(PerClass).

src/primitive_debug.cc
```diff
@@ -22,34 +22,28 @@ namespace toit {
 
 MODULE_IMPLEMENTATION(debug, MODULE_DEBUG)
 
+struct PerClass {
+  uint32 count;
+  uint32 size;
+};
+
 PRIMITIVE(object_histogram) {
   static const int UINT32_PER_ENTRY = 2;
```
kasperl, line 31: Drop this constant.

src/primitive_debug.cc
```diff
@@ -22,34 +22,28 @@ namespace toit {
 
 MODULE_IMPLEMENTATION(debug, MODULE_DEBUG)
 
+struct PerClass {
```
kasperl, line 25: I don't like the PerClass name much. 

I think it should contain the name 'instances' somehow because that is what we're accumulating information about.

InstancesData?

tools/mirror.toit
```diff
@@ -329,6 +329,45 @@ class Profile extends Mirror:
...
+  stringify -> string:
+    return "  │ $(%7d count) │ $(%6d size >> 10)k $(%4d size & 1023)b │ $(%-45s class_name)│"
+
+class Histogram extends Mirror:
+  static tag ::= 'O'  // For Objects.
+
+  entries /List := []
+
+  constructor json program/Program [on_error]:
+    pos := 4
```
kasperl, line 348: ?

tools/mirror.toit
```diff
@@ -329,6 +329,45 @@ class Profile extends Mirror:
...
+
+  constructor .class_name .count .size:
+
+  stringify -> string:
+    return "  │ $(%7d count) │ $(%6d size >> 10)k $(%4d size & 1023)b │ $(%-45s class_name)│"
+
+class Histogram extends Mirror:
+  static tag ::= 'O'  // For Objects.
+
+  entries /List := []
```
kasperl, line 345: ::=

tools/mirror.toit
```diff
@@ -329,6 +329,45 @@ class Profile extends Mirror:
...
+
+  constructor json program/Program [on_error]:
+    pos := 4
+    entries = List
```
kasperl, line 349: Are you trying to clear it here? This looks weird.

tools/mirror.toit
```diff
@@ -329,6 +329,45 @@ class Profile extends Mirror:
...
+
+    for i := 1; i < json.size; i += 3:
```
kasperl, line 351: What is at index 0?

src/encoder.h
```diff
@@ -114,6 +114,8 @@ class Encoder {
   void write_byte_array_header(int length);
   void write_string(const char* string);
 
+  static int bytes_to_encode(int64 value);
```
kasperl, line 117: encoded_size? compute_encoded_size?

src/primitive_debug.cc
```diff
@@ -60,9 +54,40 @@ PRIMITIVE(object_histogram) {
...
+      non_trivial_entries++;
+      bytes_needed += Encoder::bytes_to_encode(i);
+      bytes_needed += Encoder::bytes_to_encode(data[i].count);
+      bytes_needed += Encoder::bytes_to_encode(data[i].size);
+    }
+  }
+  MallocedBuffer encoding_buffer(100 + bytes_needed);
+  if (!encoding_buffer.has_content()) MALLOC_FAILED;
+
+  ProgramOrientedEncoder encoder(program, &encoding_buffer);
```
kasperl, line 73: Maybe it would be easier just do the encoding twice, the first time with a zero-length buffer. We count the number of bytes in the encoder even after overflow.

src/primitive_debug.cc
```diff
@@ -60,9 +54,40 @@ PRIMITIVE(object_histogram) {
...
+
+  encoder.write_header(non_trivial_entries * 3, 'O');  // O for objects.  See mirror.toit.
```
kasperl, line 75: I think I'd move this to the Encoder implementation to make it consistent with the other high-level encoder methods.

erikcorry: This is done in the same way as profiles.  It feels strange to put something so specific in a generic place like encoder.cc

src/primitive_debug.cc
```diff
@@ -60,10 +54,44 @@ PRIMITIVE(object_histogram) {
...
-  return result;
+  int non_trivial_entries = 0;
+  int bytes_needed = 0;
+  bool size_known = false;
+  // Encode twice.  First time with a dry run to calculate the size of the
+  // encoded data.
+  for (int attempt = 0; attempt < 2; attempt++) {
+    MallocedBuffer encoding_buffer(size_known ? bytes_needed : 1);
+    if (!encoding_buffer.has_content()) MALLOC_FAILED;
+    ProgramOrientedEncoder encoder(program, &encoding_buffer);
```
kasperl, line 68: I think I'd move this out in a helper and call that twice instead of wrapping it in a loop with special handling for both first and second iteration.

tools/mirror.toit
```diff
@@ -329,6 +329,48 @@ class Profile extends Mirror:
...
+
+    for i := first_entry; i < json.size; i += 3:
+      class_name := program.class_name_for json[i]
+      if class_name != "RecognizableFiller_":
+        entries.add
+          HistogramEntry class_name json[i + 1] json[i + 2]
+    entries.sort --in_place: | a b | b.size - a.size
+    super json program
+
+  table:
```
kasperl, line 361: I'd probably just inline this into stringify. Is it useful on its own?

kasperl (approved):

PR #810, opened 2022-06-22
Disable spurious use-after-free warning

erikcorry:


floitsch (approved):
LGTM.
Please change the old_ptr back to a `void*` then.

PR #812, opened 2022-06-22
Add warning when instance variable resembles constant.

floitsch:
Warn the user when it's likely they forgot a `static`.


kasperl (approved):
Looks good to me.

PR #815, opened 2022-06-24
Avoid uninstalling services on client close

kasperl:


lib/system/services.toit
```diff
@@ -159,24 +161,23 @@ abstract class ServiceDefinition:
...
-    _uninstalled_.get
-
   _open_ client/int -> List:
     _clients_.add client
     catch --trace: on_opened client
     return [ name, _version_, client ]
 
   _close_ client/int -> none:
     _clients_.remove client
+    _clients_closed_++
```
kasperl, line 171: Maybe removing the client should really be at the end of this method.

kasperl: Fixed.

lib/monitor.toit
```diff
@@ -53,81 +56,109 @@ monitor Latch:
...
+    if condition.call: return
+    wait_ condition
+
+  /**
+  Raises the signal and unblocks the tasks that are already waiting.
 
-  count_ := 0
+  If $max is provided and not null, no more than $max tasks are
+    woken up in the order in which they started waiting (FIFO).
+    The most common use case is to wake waiters up one at a time.
```
floitsch, line 140: The current implementation only runs the condition of `max` waiters.

It feels like actually waking `max` waiters makes more sense. I do realize that complicates the `wait_` method, though.

floitsch (approved):
LGTM.

lib/monitor.toit
kasperl: Not sure. I think the max parameter is useful to control the amount of re-evaluation -- not unblock a particular number of tasks. I could probably be convinced otherwise :)

lib/monitor.toit
```diff
@@ -53,81 +56,109 @@ monitor Latch:
...
 This class must not be extended.
 */
-monitor Semaphore:
+monitor Signal:
+  current_ /int := 0
+  awaited_ /int := 0
+
   /**
-  Increments an internal counter.
+  Waits until the signal has been raised while this task has been waiting.
```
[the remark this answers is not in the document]
kasperl: I wanted to convey that it doesn't help that the signal was raised before you started waiting. Maybe that is covered well enough below.

PR #820, opened 2022-06-24
Make Writer helper class robust about UTF-8 chars.

erikcorry:


floitsch (approved):
LGTM.
Might be more efficient to only send the remaining chars of the multi-byte string as a byte-array.

lib/writer.toit
```diff
@@ -37,7 +37,15 @@ class Writer:
...
       from += writer_.write data[from..to]
-      if from != to: yield
+      if from != to:
+        yield
+        while data is string and from != to and data[from] == null:  // Not on a character boundary.
+          // Can't slice here, would cut up UTF-8 string in an illegal way.
+          cut_point := from + 1
+          while cut_point != to and data[cut_point] == null:
+            cut_point++
+          snip := data.to_byte_array from cut_point
```
floitsch, line 47: Does this mean that we can `to_byte_array` at non-char boundaries?
maybe add a comment.

floitsch (commented):
LGTM.

PR #821, opened 2022-06-27
Deprecate the SDK version of arguments.toit

erikcorry:


floitsch (approved):
LGTM, but there seems to be a test for it that has a warning now.

PR #822, opened 2022-06-27
Add support for AP mode for WiFi

kasperl:


src/rtc_memory_esp32.cc
```diff
@@ -131,7 +131,7 @@ static void reset_rtc(const char* reason) {
 
 // Patch the primordial entrypoint of the image (before launching FreeRTOS).
 extern "C" void IRAM_ATTR start_cpu0() {
-  ets_printf("[toit] Starting <%s>\n", toit::vm_git_version());
+  ets_printf("[toit] INFO: starting <%s>\n", toit::vm_git_version());
```
kasperl, line 134: I think this looks nicer with the other lines, so I will update Jaguar if this lands.

src/resources/wifi_esp32.cc
```diff
@@ -52,28 +52,59 @@ ResourcePool<int, kInvalidWifi> wifi_pool(
...
       RtcMemory::set_wifi_channel(0);
     }
+
+    esp_err_t err = esp_wifi_set_mode(WIFI_MODE_STA);
+    if (err != ESP_OK) return err;
+
+    wifi_config_t config;
+    memset(&config, 0, sizeof(config));
+    strncpy(char_cast(config.sta.ssid), ssid, sizeof(config.sta.ssid));
+    strncpy(char_cast(config.sta.password), password, sizeof(config.sta.password));
```
erikcorry, line 83: I think we should be copying the min of the strlen and the size of the destination.
And perhaps zero the last byte of config.sta.password to ensure that it is null terminated for a too-long password.

src/resources/wifi_esp32.cc
```diff
@@ -52,28 +52,59 @@ ResourcePool<int, kInvalidWifi> wifi_pool(
...
+  }
+
+  esp_err_t establish(const char* ssid, const char* password, bool broadcast, int channel) {
+    esp_err_t err = esp_wifi_set_mode(WIFI_MODE_AP);
+    if (err != ESP_OK) return err;
+
+    wifi_config_t config;
+    memset(&config, 0, sizeof(config));
+    strncpy(char_cast(config.ap.ssid), ssid, sizeof(config.ap.ssid));
+    strncpy(char_cast(config.ap.password), password, sizeof(config.ap.password));
```
erikcorry, line 98: Same

src/resources/wifi_esp32.cc
```diff
@@ -305,23 +337,47 @@ PRIMITIVE(connect) {
...
+
+  WifiEvents* wifi = _new WifiEvents(group);
+  if (wifi == null) MALLOC_FAILED;
+
+  group->register_resource(wifi);
+
+  esp_err_t err = group->establish(ssid, password, broadcast, channel);
+  if (err != ESP_OK) {
+    group->unregister_resource(wifi);
+    return Primitive::os_error(err, process);
```
erikcorry, line 359: Doesn't this leak wifi?

src/resources/wifi_esp32.cc
```diff
@@ -352,13 +408,26 @@ PRIMITIVE(disconnect_reason) {
...
+  if (!group->has_ip_address()) {
+    return process->program()->null_object();
+  }
+
+  ByteArray* result = process->object_heap()->allocate_internal_byte_array(4);
+  if (!result) ALLOCATION_FAILED;
+  ByteArray::Bytes bytes(result);
+
+  uint32 address = group->ip_address();
+  bytes.at_put(0, (address >>  0) & 0xff);
```
erikcorry, line 421: Utils::write_unaligned_uint32_le(bytes.address(), address);

kasperl: Done.

src/resources/wifi_esp32.cc
kasperl: No. Unregistering the resource deletes it.

PR #826, opened 2022-06-28
Update MQTT gold files for deprecated put_byte

erikcorry:
This seems preferable to changing MQTT so it becomes incompatible with older SDKs.


floitsch (approved):
LGTM.

PR #829, opened 2022-06-28
Add IPv6 DNS record type

erikcorry:


lib/net/modules/dns.toit
```diff
@@ -43,6 +43,7 @@ INTERNET_CLASS  ::= 1
 
 A_RECORD        ::= 1
```
kasperl, line 44: I'd consider changing these to RECORD_A, RECORD_CNAME, and RECORD_AAAA.

kasperl (approved):
Looks good to me.

PR #831, opened 2022-06-29
Sample based profiler

kasperl:


src/os_posix.cc
```diff
@@ -94,19 +94,15 @@ class ConditionVariable {
...
   }
 
-  bool wait(int timeout_in_ms) {
-    if (timeout_in_ms == 0) {
-      // No timeout.
-      wait();
-      return true;
-    }
+  bool wait_us(int64 us) {
+    if (us <= 0LL) return false;
```
erikcorry, line 98: Do you really need the "LL" here?

src/os_posix.cc
```diff
@@ -94,19 +94,15 @@ class ConditionVariable {
...
 
     // TODO: We really should use monotonic time here.
     struct timespec deadline = { 0, };
     if (!OS::get_real_time(&deadline)) {
       FATAL("cannot get time for deadline");
     }
-    OS::timespec_increment(&deadline, timeout_in_ms * 1000000LL);
+    OS::timespec_increment(&deadline, us * 1000LL);
```
erikcorry, line 105: Or here?

erikcorry (approved):
LGTM

Might be worth considering just counting the location of the back edge instead of the start of the method - saves some space in the byte codes because we don't need to store the offset, but costs some space when profiling because we have more different locations to count ticks.

PR #839, opened 2022-06-30
Add string.substitute

erikcorry:


tests/string_test.toit
```diff
@@ -1636,3 +1638,29 @@ test_hash_code:
...
+  expect_equals "fixed" result
+
+  result = "42foobarfizz103".substitute --open="foo" --close="fizz": "BAR"
+  expect_equals "42BAR103" result
+
+  result = "{{variable}} is not variable".substitute: MAP[it]
+  expect_equals "fixed is not variable" result
+
+  // Check that we remember to stringify.
+  "The time is {{time}} now.".substitute: Time.now.local
```
floitsch, line 1666: Add a test with spaces around the variable name.

Add a test where the variable name is equal to the open-token. (prevents an implementation that does `str.split open_token`).

floitsch (approved):
LGTM.

PR #843, opened 2022-07-01
Use map for WiFi configuration

kasperl:


lib/net/wifi.toit
```diff
@@ -7,16 +7,27 @@ import system.api.wifi show WifiServiceClient
 
 import .impl
 
+CONFIG_SSID      /string ::= "ssid"
+CONFIG_PASSWORD  /string ::= "password"
+
+CONFIG_BROADCAST /string ::= "broadcast"
+CONFIG_CHANNEL   /string ::= "channel"
+
+OPEN_KEYS_      /List ::= [CONFIG_SSID, CONFIG_PASSWORD]
```
kasperl, line 16: These beauties will go away when we can serialize/deserialize maps in the messaging primitives.

erikcorry (approved):

PR #846, opened 2022-07-01
Map serialize

erikcorry:


lib/core/collections.toit
```diff
@@ -2047,15 +2047,18 @@ abstract class HashedInsertionOrderedCollection_:
...
         return APPEND_
       else:
-        assert: size_ == 1
-        k := backing_[0]
-        if k is not Tombstone_:
-          if compare_ key k:
-            return 0
-          append_position = not_found.call
-          rebuild_ 1 --allow_shrink
+        if size_ != 1:
```
kasperl, line 2050: Make this part of the previous else:?

```
else if size_ != 1:
  ...
else:
  ...
```

lib/expect.toit
```diff
@@ -59,16 +59,52 @@ expect_not_null actual:
...
+/**
+Expects the $actual object to be structurally equal to the $expected object.
+Understands operator equality and memberwise structural equality of lists and
+  maps.
+*/
+expect_structural_equals expected/Object actual/Object:
+  if not structural_equals expected actual:
+    expect false --message="Expected <$expected>, but was <$actual>"
+
+structural_equals expected actual -> bool:
```
kasperl, line 71: Maybe the xxx_equals methods belong somewhere else? Not sure.

src/messaging.cc
```diff
@@ -115,11 +116,19 @@ bool MessageEncoder::encode(Object* object) {
...
         Array* array = Array::cast(backing);
-        return encode_array(array, Smi::cast(instance->at(1))->value());
+        Object* size = instance->at(Instance::LIST_SIZE_OFFSET);
+        if (!is_smi(size)) return false;
+        return encode_array(array, Smi::cast(size)->value());
+      } else if (class_id == _program->large_array_class_id()) {
+        printf("Can't serialize large array\n");
       }
+    } else if (class_id == _program->map_class_id()) {
+      return encode_map(Instance::cast(object));
```
kasperl, line 131:
```suggestion
      return encode_map(instance);
```

src/messaging.cc
```diff
@@ -167,6 +176,43 @@ bool MessageEncoder::encode_array(Array* object, int size) {
...
+  if (is_smi(bk)) return false;
+  HeapObject* backing = HeapObject::cast(bk);
+  Object* sz = instance->at(Instance::MAP_SIZE_OFFSET);
+  if (!is_smi(sz)) return false;
+  word size = Smi::cast(sz)->value();
+  write_cardinal(size);
+  if (size == 0) return true;  // Do this before looking at the backing, which may be null.
+  Smi* class_id = backing->class_id();
+  if (class_id == _program->list_class_id()) {
+    bk = Instance::cast(backing)->at(Instance::LIST_ARRAY_OFFSET);
```
kasperl, line 191: Maybe you should just replace bk and sz with object and use that consistently as a pattern -- or maybe we should have a few helpers on Instance to do this?

src/messaging.cc
```diff
@@ -167,6 +176,43 @@ bool MessageEncoder::encode_array(Array* object, int size) {
...
+  class_id = backing->class_id();
+  if (class_id != _program->array_class_id()) {
+    if (class_id == _program->large_array_class_id()) {
+      printf("Can't serialize large map\n");
+    }
+    return false;
+  }
+  Array* array = Array::cast(backing);
+  int i = 0;
+  for (int index = 0; i < size; index += 2) {
```
kasperl, line 204: I would do 

`for (int i = 0; ...)`

and rename the outer `i` to count.

src/messaging.cc
```diff
@@ -167,6 +176,43 @@ bool MessageEncoder::encode_array(Array* object, int size) {
...
+  }
+  class_id = backing->class_id();
+  if (class_id != _program->array_class_id()) {
+    if (class_id == _program->large_array_class_id()) {
+      printf("Can't serialize large map\n");
+    }
+    return false;
+  }
+  Array* array = Array::cast(backing);
+  int i = 0;
```
kasperl, line 203: What are you using this for? You could ASSERT that it is the same as size after the loop?

tests/rpc_test.toit
```diff
@@ -268,6 +269,20 @@ test_sequential myself/int broker/RpcBroker -> none:
   broker.unregister_procedure name
   expect.expect_throw "No such procedure registered: 800": rpc.invoke myself name []
 
+test_map myself/int -> none:
+  m := {"foo": 42, "bar": [1, 2]}
```
kasperl, line 273: I'd test more cases. Maps inside maps. Empty maps.

tests/rpc_test.toit
```diff
@@ -511,7 +526,9 @@ test myself/int arguments/any:
   else:
     actual = rpc.invoke myself PROCEDURE_ECHO arguments
   if arguments is List:
-    expect.expect_list_equals expected actual
+    expect.expect_structural_equals expected actual
```
kasperl, line 529: Combine these two? List or map.

kasperl (approved):
Looks good to me!

PR #849, opened 2022-07-02
Change ARRAY_CANNOT_CHANGE_SIZE to COLLECTION_CANNOT...

erikcorry:
We also throw this exception when someone attempts to
change the size of a map that has been through RPC, so
it makes sense for the exception to be more generic.


kasperl (approved):
Looks good to me.

PR #850, opened 2022-07-04
Do print_ in one operation.

erikcorry:
Because of multithreading we often see another print between the text
and the newline.
This is an attempt to get the newline together with the text more often.


kasperl (approved):
Looks good to me.

PR #851, opened 2022-07-04
Upgrade to ESP-IDF 4.4

floitsch:


floitsch:
A few notes:
During merge some of the `sorted` functions in the esp-idf were lost:
```
for (s, v) in sorted(sections.items())
for (target, sections) in sorted(scheme_dictionary['default'].items()
```
We should have another look at https://github.com/toitware/esp-idf/commit/60cb20449876d817942c32de46827b7cc9f60b26 and make sure it still does the same as before.

Another thing we should still do: the esp-idf/patch-head-4.4-1 uses lwip-fix-2.1.2-esp-4.3-4.
We should tag that commit as lwip-fix-2.1.2-esp4.4-1 to sync the two repositories.

kasperl (approved):
Looks good to me.

PR #860, opened 2022-07-08
Add a gate to the monitor library.

floitsch:


lib/monitor.toit
```diff
@@ -165,6 +165,60 @@ monitor Signal:
...
+    if is_open: return
+    open_ = true
+    signal_.raise
+
+  /**
+  Closes the gate.
+
+  Any task that is trying to $enter will block until the gate is opened again.
+  */
+  close:
```
kasperl, line 201: I'm not super excited about this being called 'close'. It is very different from other uses of close.

Maybe we can use lock/unlock instead?

kasperl (approved):
Looks good to me.

lib/monitor.toit
floitsch: done.

PR #861, opened 2022-07-08
Rewrite channel and add getters.

floitsch:


lib/monitor.toit
```diff
@@ -222,12 +221,23 @@ monitor Channel:
...
 
+  /**
+  The capacity of the channel.
+  */
+  capacity -> int: return buffer_.size
+
+  /**
+  The amount of messages that are currently queued in the channel.
+  */
+  buffered_count -> int: return buffered_count_
```
kasperl, line 239: I kinda like just calling this size. What do you think?

kasperl (approved):
Looks good to me.

lib/monitor.toit
floitsch: I'm hesitant. I can easily see myself asking for "a channel of size 8".
To me the "size" often refers to the capacity of the channel, and not how many items are in it.

floitsch: Changed to `size`.
done.

PR #863, opened 2022-07-08
Make 'Task' public

floitsch:
Also deprecate `task` and change it to `Task.current`.


lib/core/task.toit
```diff
@@ -53,19 +55,56 @@ The Toit programming language is cooperatively scheduled, so it is important
...
+
+Tasks are created by calling $(task code). They can either terminate by themselves
+  (gracefully or with an exception) or by being canceled from the outside with
+  $cancel.
+
+If a task finishes with an exception it brings down the whole program.
+*/
+interface Task:
+  /**
+  Returns the $Task for the current task.
```
kasperl, line 76:
```suggestion
  Returns the current task.
```

lib/core/task.toit
```diff
@@ -53,19 +55,56 @@ The Toit programming language is cooperatively scheduled, so it is important
...
+  */
+  static current -> Task:
+    #primitive.core.task_current
+
+  /**
+  Cancels the task.
+
+  If the task has `finally` clauses, those are executed.
+  However, these must not yield, as the task won't run again. Use `do_critical` to
```
kasperl, line 85:
```suggestion
  However, these must not yield, as the task won't run again. Use $critical_do to
```

kasperl (approved):
Looks good to me.

PR #865, opened 2022-07-12
Make fail-case of external tests more robust.

floitsch:
CMake would complain if it can't find the test we try to mark as
failing. With the new setup it only looks if it a new test is in the
failing-test list.


kasperl (approved):
Looks good to me.

PR #873, opened 2022-07-18
Try to use the Jaguar device name.

floitsch:
The output of the name was a stringified byte-array. For example:
```
 #[0x98, 0xcd, 0xac, 0x61, 0x71, 0xcc]
```

Now it either picks the name that got stored with Jaguar, or uses the
hardware ID:
```
96980af0-3b26-59d2-be76-fda35e89df2
```


kasperl (approved):
Looks good to me.

PR #874, opened 2022-07-18
Add ceil, floor and truncate to floats.

### LINKED ISSUE #866: Add ceil and floor methods to the float type.

floitsch:
Fixes #866.

### DIFF
--- a/src/primitive_core.cc
+++ b/src/primitive_core.cc
@@ -1353,6 +1353,21 @@ PRIMITIVE(float_sqrt) {
   return Primitive::allocate_double((double) sqrt(receiver), process);
 }
 
+PRIMITIVE(float_ceil) {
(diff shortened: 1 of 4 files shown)


src/primitive_core.cc
```diff
@@ -1353,6 +1353,21 @@ PRIMITIVE(float_sqrt) {
   return Primitive::allocate_double((double) sqrt(receiver), process);
 }
 
+PRIMITIVE(float_ceil) {
+  ARGS(double, receiver);
+  return Primitive::allocate_double((double) ceil(receiver), process);
```
kasperl, line 1358: Why do we need the C-style casts of the result here and in other places?

kasperl (approved):
Looks good to me.

src/primitive_core.cc

PR #875, opened 2022-07-18
Give warning when assigning local to itself.

floitsch:


src/compiler/resolver_method.cc
```diff
@@ -4022,7 +4022,30 @@ ir::Expression* MethodResolver::_assign_identifier(ast::Binary* node,
...
+    auto assig = result->as_AssignmentLocal();
+    auto local = assig->local();
+    auto right = assig->right();
+    if (right->is_ReferenceLocal() && right->as_ReferenceLocal()->target() == local) {
+      if (_method->is_constructor() || _method->is_instance()) {
+        auto fields = _method->holder()->fields();
+        for (int i = 0; i < fields.length(); i++) {
+          auto field_name = fields[i]->name();
+          if (field_name.is_valid() && field_name == local->name()) {
+            diagnostics()->report_warning(node, "Assignment of local to itself, did you forget 'this.'?");
```
kasperl, line 4037: Maybe we can find a better warning? It sounds very informational right now. 

Assigning local to itself has no effect

kasperl (approved):
Looks good to me.

PR #877, opened 2022-07-18
Fixes bad resolution of index-calls with parameters

floitsch:
Fixes a confusion in the resolver.
Take the call `o[3] 4`

The parser would correctly parse this as a call of `o[3]` with argument
`4`.

However, the resolver tried to reuse functionality and gave a
call-resolving function just a "target" and "arguments".

For a normal call the "target" was the target of the call (for `foo 1
2`, the `foo` identifier).
For an index call, the "target" was the node itself: `o[3]`.

Once the target and arguments were resolved, the resolver would then
again check whether it was trying to solve a normal call or an index
call. It did this, by looking at the target-node. If it was and
index-node, it assumed that it was resolving and index-call. Otherwise,
it assumed it was for a call-node.

However, this led to the following bug, where a normal node had an
index-call as target. For our example `o[3] 4`, the resolver now got
confused and thought that it was doing work for an index node (since
the target of the call was `o[3]`, and index node). It assumed that
the argument `4` was the argument to the index (as if the user had
written `o[4]`). The actual arguments to the index-node were completely
dropped.

### DIFF
--- a/src/compiler/resolver_method.cc
+++ b/src/compiler/resolver_method.cc
@@ -2717,16 +2717,16 @@ void MethodResolver::_visit_potential_call_dot(ast::Dot* ast_dot,
   }
 }
 
-void MethodResolver::_visit_potential_call_index(ast::Index* ast_index,
+void MethodResolver::_visit_potential_call_index(ast::Node* ast_target,
                                                  CallBuilder& call_builder) {
-  auto receiver = resolve_expression(ast_index->receiver(),
+  auto receiver = resolve_expression(ast_target,
                                      "Can't use the index operator on a block");
   push(call_builder.call_instance(_new ir::Dot(receiver, Symbols::index)));
 }
 
-void MethodResolver::_visit_potential_call_index_slice(ast::IndexSlice* ast_index_slice,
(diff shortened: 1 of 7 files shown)


src/compiler/resolver_method.cc
```diff
@@ -2968,31 +2969,39 @@ void MethodResolver::_visit_potential_call(ast::Node* ast_target,
...
-                              call_builder,
-                              named_lsp_selection);
-  } else if (ast_target->is_Index()) {
-    _visit_potential_call_index(ast_target->as_Index(), call_builder);
-  } else if (ast_target->is_IndexSlice()) {
-    _visit_potential_call_index_slice(ast_target->as_IndexSlice(), call_builder);
-  } else if (is_literal_super(ast_target) ||
-             (ast_target->is_Dot() && is_constructor_super_call)) {
-    _visit_potential_call_super(ast_target, call_builder, is_constructor_super_call);
+  if (potential_call->is_Index()) {
```
floitsch, line 2972: This PR is significantly easier to review with whitespace changes turned off.
I moved a chunk into an 'else' branch and thus nested it.

kasperl (approved):
Looks good to me.

PR #879, opened 2022-07-19
Add a way to keep CS active.

### LINKED ISSUE #809: The SPI module for the ESP32 does not allow you to choose if the CS pin should be held after transmission
There are many different IC's that use SPI as a communication method, and they have different implementations of the protocol.
Some implementations will give different responses to the same command based on the previous command and the time since the last command.

An example of this is the Si4455 radio transceiver (https://www.silabs.com/documents/public/data-sheets/Si4x55-C.pdf).
In the SPI implementation for the transceiver, you have to poll the chip's status command until it is ready. When the status commands indicate that the chip is ready, data can be read from the chip.

The current implementation of the SPI driver allows you to select a CS pin that can be controlled with the SPI peripheral register. However the exposed primitives for the SPI driver does not allow you to select wether or not the CS line should be held after a transmission. Therefore you must manually control the CS pin in such applications:

    poll_cts_ -> int:

        data := #[READ_CMD_BUFF_, 0xFF]
        
        nsel_.set 0
        device_.transfer data --read=true

        return data[1]

    read_response_ data/ByteArray:
        cts := 0

        data.fill 0xFF

        while cts != 0xFF:
            cts = poll_cts_
            if cts == 0xFF:
                device_.transfer data --read=true
            nsel_.set 1

The ESP32 has a bit that controls whether or not to hold the CS line after the given transmission. 
This flag is located in the PI_PIN_REG (0x34) as described in Register 7.14 on page 136 in the ESP32 reference manual (https://www.espressif.com/sites/default/files/documentation/esp32_technical_reference_manual_en.pdf).

This should be simple to add to the SPI transfer primitive and if a default is added to the SPI implementation in the library, it should be possible to add this feature while keeping the backward compatibility for the existing code base.

floitsch:
Fixes #809.

### DIFF
--- a/lib/spi.toit
+++ b/lib/spi.toit
@@ -4,6 +4,7 @@
 
 import gpio
 import serial
+import monitor
 
 /**
 SPI is a serial communication bus able to address multiple devices along a main 3-wire bus with an additional 1 wire per device.
@@ -39,6 +40,15 @@ Each device on the bus is enabled with its own chip-select pin. See $Bus.device.
 */
 class Bus:
   spi_ := ?
+  /**
+  Mutex to serialize reservation attempts of multiple devices.
+  See $Device.reserve_bus.
+
+  When trying to acquire the bus, the ESP-IDF currently (as of 2022-07-19) does not allow to set a timeout.
+    This means that the program would be stuck in the primitive. We thus use this mutex to avoid that
+    situation.
+  */
+  reservation_mutex_/monitor.Mutex ::= monitor.Mutex
 
   /**
   Constructs a new SPI bus using the given $mosi, $miso, and $clock pins.
@@ -111,6 +121,10 @@ interface Device extends serial.Device:
     value of $dc.
   If a commands and/or address sections was defined, use $command and
     $address to set the values.
+
+  When $keep_cs_active is true, then the chip select pin is kept active
+    after the transfer. This functionality is only allowed when the
+    bus is reserved for this device. See $reserve_bus.
   */
   transfer
       data/ByteArray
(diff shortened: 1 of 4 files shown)


lib/spi.toit
```diff
@@ -120,6 +134,21 @@ interface Device extends serial.Device:
...
+  Starts by acquiring the bus. Once that's succeeded, executes the $block. Finally, releases
+    the bus before returning.
+
+  Reserving the bus can be useful in two contexts:
+  1. The CS pin is controlled by the user. Since the hardware only supports a limited number of
+    automatic CS pins, it might be necessary to set some CS pins by hand. This should be done
+    after the bus has been reserved.
+  2. When using the `--keep_cs_active` flag of the $transfer function, the bus must be reserved.
+  */
+  reserve_bus [block]
```
kasperl, line 151: Alternative names that makes it slightly more obvious that this takes a block argument:

reserve_bus_during
reserved_do
with_reserved_bus

src/resources/spi_esp32.cc
```diff
@@ -247,6 +251,23 @@ PRIMITIVE(transfer) {
...
+  esp_err_t err = spi_device_acquire_bus(device->handle(), portMAX_DELAY);
+  if (err != ESP_OK) {
+    return Primitive::os_error(err, process);
+  }
+  return process->program()->null_object();
+}
+
+PRIMITIVE(release_bus) {
+  ARGS(SPIDevice, device);
+  printf("Releasing the bus\n");
```
kasperl, line 265: Drop debug printing.

src/resources/spi_esp32.cc
```diff
@@ -247,6 +251,23 @@ PRIMITIVE(transfer) {
...
+  spi_device_release_bus(device->handle());
+  printf("Released\n");
```
kasperl, line 267: Drop debug printing.

tests/esp32/spi_keep_active.toit
```diff
@@ -0,0 +1,93 @@
...
+      --mosi=gpio.Pin MOSI
+
+  device := bus.device
+      --cs=gpio.Pin CS
+      --frequency=10_000_000
+
+  in := gpio.Pin IN_CS --input
+
+  2.repeat: | test_iteration/int |
+    print "starting round $test_iteration"
```
kasperl, line 52: Maybe drop this - or capitalize for consistency.

kasperl (approved):
Looks good to me.

src/resources/spi_esp32.cc
floitsch: done.

src/resources/spi_esp32.cc
floitsch: done.

tests/esp32/spi_keep_active.toit
floitsch: removed.

lib/spi.toit
floitsch: Went with `with_reserved_bus`.

PR #880, opened 2022-07-19
Fix error message for methods in super classes

floitsch:
When providing helpful error messages of why a method call didn't work,
     we didn't take super classes into account.


kasperl (approved):
Looks good to me!

PR #884, opened 2022-07-20
Improve configuration of pins.

floitsch:
Also rename from `config` to `configure`.

### DIFF
--- a/lib/gpio/pin.toit
+++ b/lib/gpio/pin.toit
@@ -137,7 +137,6 @@ class Pin:
     the corresponding libraries (like the i2c library) already take care of setting this
     configuration for you.
   Note that it is not safe to ground an $open_drain pin and to connect it externally to VCC.
-
   Also note, that only one entity on an open-drain bus needs to pull the bus high. As such,
     it can be useful to set $open_drain without $pull_up.
   */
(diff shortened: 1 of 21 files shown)


lib/gpio/pin.toit
```diff
@@ -91,11 +101,60 @@ class Pin:
...
+  If a pin is configured to be an input, it can have a $pull_up or $pull_down.
+
+  If $open_drain is set, then the pin can only pull the pin to the ground. Together, with
+    a $pull_up resistor this still allows the pin to emit both 0 and 1. In this configuration,
+    connected devices can also safely pull the pin to ground without damaging the microcontroller.
+    This configuration is typically used in communications that only use one data bus for
+    input and output, such as the DHT11/DHT22, the i2c bus, and the one-wire bus. Note, that
+    the corresponding libraries (like the i2c library) already take care of setting this
+    configuration for you.
+  Note that it is not safe to ground an $open_drain pin and to connect it externally to VCC.
```
kasperl, line 139: When do you add a newline between the paragraphs and when do you not?

kasperl (approved):
Looks good to me.

lib/gpio/pin.toit
floitsch: I don't think there is a big rule, and the toitdoc currently doesn't make a difference.
Here, the "Note" is related to the 'open_drain' so I kept it together, but you are right: the next note should be attached too.
done.

Issue #130, opened 2021-12-13
Clang 13 on macOS mis-compiles third_party/esp-idf/components/mbedtls


kasperl:
The Toit virtual machine compiled with Clang 13 on macOS, does not have a working TLS implementation. Clang 12.0.5 generates the correct code.

Could be related to https://github.com/ARMmbed/mbedtls/issues/5052.

kasperl:
At least one problem is here: https://github.com/espressif/mbedtls/blob/6465247f67167518b8813ae2faaf422704e4b1a3/include/mbedtls/bn_mul.h#L258.

This should be fixed like here: https://github.com/ARMmbed/mbedtls/commit/d337fbc4cb1ded1d58a1c4fe92aa9988d68ec1f3.

kasperl:
The GitHub actions we use to build releases now use Clang 12, which makes this less critical.

See 25dc42a790cabf69519b9a0c0f2a9e8581ff8795.

floitsch:
Has been fixed by upgrading ESP-IDF.

PR #886, opened 2022-07-21
Move to newer builder.

floitsch:
Windows-latest is now Windows-2022.
The fix for mbdetls has been merged. So we don't need an older macos
anymore. See https://github.com/toitlang/toit/issues/130.


kasperl (approved):
Looks good to me.

PR #889, opened 2022-07-22
Also check for EAGAIN.

floitsch:
Practically speaking they are nowadays the same error (`#define
    EWOULDBLOCK EAGAIN`), but documentation states that they could be
different and should be evaluated both.


floitsch:
In practice this PR probably doesn't make any difference. I'm fine with dropping it.

kasperl (approved):
Looks good to me.

PR #890, opened 2022-07-22
Remove work-around for uart_get_buffered_data_len

floitsch:
The race condition has been fixed: https://github.com/espressif/esp-idf/issues/6397


floitsch:
The workaround isn't costing much. We could leave it in.

kasperl (approved):
Looks good to me.

PR #891, opened 2022-07-22
If pin has already value exit wait_for immediately

floitsch:
No need to set the interrupt if the pin has already the correct value.


kasperl (approved):
Looks good to me.

PR #897, opened 2022-07-27
Run tests on windows with bash

floitsch:
Powershell doesn't abort execution when a command fails, so we were not
correctly testing on Windows.


kasperl (approved):
Looks good to me.

PR #907, opened 2022-08-01
Add basic support for updating WiFi credentials

kasperl:


system/extensions/esp32/wifi.toit
```diff
@@ -59,6 +66,12 @@ class WifiServiceDefinition extends NetworkServiceDefinitionBase:
     if module.ssid != ssid or module.password != password:
       throw "wifi already connected with different credentials"
 
+    if save:
+      if config:
+        store_.set WIFI_CONFIG_STORE_KEY config
+      else:
+        store_.delete WIFI_CONFIG_STORE_KEY
```
floitsch, line 73: This is a bit weird.
You would basically say `connect --save`, and we would connect to the saved configuration, but then remove it.

floitsch (approved):
LGTM.

lib/net/wifi.toit
```diff
@@ -17,7 +17,7 @@ service_/WifiServiceClient? ::= (WifiServiceClient --no-open).open
 
 open --ssid/string --password/string -> net.Interface
     --save/bool=false:
-  return open {
+  return open --save=save {
```
floitsch, line 20: good catch...

PR #908, opened 2022-08-01
Cleaner close of UDP socket.

erikcorry:
When we are stuck in receive and a different task closes our
socket we just want to return null, and not throw an exception.


floitsch (approved):
LGTM.

floitsch (approved):
Still LGTM.

PR #910, opened 2022-08-02
Update patch-head to 4.4-2

floitsch:
This includes changes to fix WiFi issues we have encountered.


kasperl (approved):
Looks good to me.

PR #921, opened 2022-08-03
Move run functions into their own file.

floitsch:


kasperl (approved):
Looks good to me.

Maybe move the stuff we need to run from image in here too?

PR #922, opened 2022-08-03
Embed the boot snapshot in the executable.

floitsch:


src/run.cc
```diff
@@ -37,7 +42,20 @@ ProgramImage read_image_from_bundle(SnapshotBundle bundle) {
 }
 
 int run_program(const char* boot_bundle_path, SnapshotBundle application_bundle, char** argv) {
-  auto boot_bundle = SnapshotBundle::read_from_file(boot_bundle_path, true);
+  if (boot_bundle_path != null) {
+    auto boot_bundle = SnapshotBundle::read_from_file(boot_bundle_path, true);
+    int result = run_program(boot_bundle, application_bundle, argv);
+    // TODO(florian): we should free the boot_bundle buffer here, but that's already
+    // done by the `run_program` as the buffer is sent in a message and then freed as an external byte array.
+    // That should not happen.
```
kasperl, line 50: What does that mean? That we will fix it?

kasperl (approved):
Looks good to me.

src/run.cc
floitsch: removed that confusing line.

PR #923, opened 2022-08-03
Allow to compile toit programs to executables.

floitsch:


floitsch:
If we are happy with this, I would then modify our tools to become executables.
This would require changing some cmake-rules. At that point I would add tests.

kasperl (approved):
Looks good to me.

kasperl (approved):
Looks good to me.

PR #926, opened 2022-08-04
Update tpkg dependency.

floitsch:
This update fixes the environment entry in the package.yaml file.


kasperl (approved):
Looks good to me.

PR #937, opened 2022-08-08
Sign executables on Mac.

floitsch:


src/compiler/executable.cc
```diff
@@ -44,6 +51,34 @@ static const uint8 VESSEL_TOKEN[] = { VESSEL_TOKEN_VALUES };
...
+  if (posix_spawnp(&child_pid, "codesign", null, null, argv, environ) != 0) goto fail;
+  do {
+    if (waitpid(child_pid, &status, 0) == -1) goto fail;
+  } while (WIFEXITED(status) == 0 && WIFSIGNALED(status) == 0);
+
+  free(out_path_mutable);
+  if (WIFEXITED(status) != 0) return WEXITSTATUS(status);
+  if (WIFSIGNALED(status) != 0) return -1;
+
+  fail:
```
kasperl, line 75: I'd indent the label by only one space and let the following lines have only two space indent.

src/compiler/executable.cc
```diff
@@ -44,6 +51,34 @@ static const uint8 VESSEL_TOKEN[] = { VESSEL_TOKEN_VALUES };
 // just much more complicated for something that doesn't change that frequently.
 static int VESSEL_SIZES[] = { 128, 256, 512, 1024, 8192, };
 
+static int sign_if_necessary(const char* out_path) {
+#ifndef TOIT_DARWIN
+  return 0;
+#else
+  char codesign[] = { "codesign" };
```
kasperl, line 58: How does one install codesign? Do we need to update a README?

kasperl (approved):
Looks good to me.

src/compiler/executable.cc
floitsch: I think/hope it's preinstalled. If it isn't I would probably just ship the zsign executable with Toit instead.
This is the approach I would take for cross-compilation too. (That said: currently we ship with non-ARM executables, so the precompiled executables shouldn't complain).

src/compiler/executable.cc
floitsch: done.

PR #942, opened 2022-08-10
Take a lock while printing to the console/UART

erikcorry:


erikcorry:
This doesn't work - the prints from the logger in jaguar still get mixed up with other output.

To test, run the following on the device with jag monitor and jag run:

```
main:
  foo

foo:
  bar

bar:
  throw "a fish"
```

src/os_posix.cc
```diff
@@ -220,6 +220,7 @@ void OS::set_up() {
   ASSERT(sizeof(void*) == sizeof(pthread_t));
   (void) pthread_key_create(&thread_key, null);
   Thread::ensure_system_thread();
+  _print_mutex = allocate_mutex(-1, "Print mutex");
```
kasperl, line 223: As discussed, this should probably be 0.

kasperl (approved):
Looks good to me.

kasperl:
I assume this isn't going in? Feel free to reopen.

PR #948, opened 2022-08-15
Process messages before suspending tasks

kasperl:


lib/core/task.toit
```diff
@@ -247,41 +245,37 @@ class Task_ implements Task:
...
+  static STATE_SUSPENDED  /int ::= 2
+  state_ := null  // TODO(kasper): Document this.
 
 // ----------------------------------------------------------------------------
 
 task_new_ lambda/Lambda -> Task_:
   #primitive.core.task_new
 
-task_transfer_ to/Task_ detach_stack:
+task_transfer_to_ to/Task_ detach_stack:
```
floitsch, line 280: It's otherwise not possible to know what the boolean for each `task_transfer_to_` call is.
```suggestion
task_transfer_to_ to/Task_ --detach_stack/bool:
```

That said: this relies on the fact that named parameters are added after positional parameters, which is an implementation detail.
Alternatives are to have a function with named parameter that redirects to one without, or to keep the code as is.

lib/core/task.toit
```diff
@@ -170,57 +172,53 @@ class Task_ implements Task:
...
-    previous := previous_running_
-    if previous == this:
-      // If we encounted a root-error, terminate the process.
-      if exit_with_error_: __exit__ 1
-      // Check whether no service definitions and only background tasks are running.
-      if ServiceManager_.is_empty and task_count_ == task_background_: __halt__
-      is_task_idle_ = true
+    processed := false
+    state_ = STATE_SUSPENDING
+    if identical this previous_running_:
```
floitsch, line 188: Is it necessary to make this 'if'?
Is it, so we give every task a chance to run when one task exited with an error? If yes, add a comment.

lib/core/task.toit
```diff
@@ -170,57 +172,53 @@ class Task_ implements Task:
...
+        if not identical this previous_running_: break
         __yield__
     else:
-      // Unlink from the linked of running tasks.
-      next := next_running_
-      previous.next_running_ = next
-      next.previous_running_ = previous
-      next_running_ = previous_running_ = this
-      return next
+      process_messages_
```
floitsch, line 199: Can a `process_messages_` lead to a task being removed from the list?
If yes, can we end up being the lone survivor?

lib/core/task.toit
```diff
@@ -170,57 +172,53 @@ class Task_ implements Task:
...
-
     // Link the task into the linked list of running tasks
     // at the very end of it.
-    previous := current.previous_running_
-    current.previous_running_ = this
-    previous.next_running_ = this
-    previous_running_ = previous
-    next_running_ = current
+    if state_ == STATE_SUSPENDED:
+      current/Task_ := Task_.current
```
floitsch, line 215:
```suggestion
      current := Task_.current
```

lib/core/task.toit
```diff
@@ -247,41 +245,37 @@ class Task_ implements Task:
...
+  // Deadline and cancelation support.
   deadline_ := null
   is_canceled_ := null
   critical_count_ := null
 
   // If the task is blocked in a monitor, this reference that monitor.
   monitor_ := null
 
-  // All running tasks are chained together in linked list.
+  // All running tasks are chained together in doubly linked list.
```
floitsch, line 260:
```suggestion
  // All running tasks are chained together in a doubly linked list.
```

floitsch (approved):
LGTM.

lib/core/task.toit
kasperl: I thought about this. Will keep as is for now.

lib/core/task.toit
kasperl: Simplified.

lib/core/task.toit
kasperl: Fixed the behavior to eagerly exit.

floitsch (approved):
Still LGTM.

PR #951, opened 2022-08-15
Remove support for deprecated A.member.

floitsch:
We now required the constructor to be in parenthesis: `(A).member`.
By removing the deprecated support we can provide better error messages
when the left-hand side points to a constructor.

In the following example we used to give bad messages:
```
class A:
  static FOO ::= 499
  constructor x:
main:
  print A.FOOO
```

```
bad.toit:6:9: error: Argument mismatch for 'A'
Too few arguments provided
  print A.FOOO
        ^
```


kasperl (approved):
Looks good to me.

PR #954, opened 2022-08-16
Activate PSRAM on ESP32s that have it

erikcorry:


src/os.cc
```diff
@@ -35,6 +35,7 @@ namespace toit {
 Mutex* OS::_global_mutex = null;
 Mutex* OS::_scheduler_mutex = null;
 Mutex* OS::_resource_mutex = null;
+int    OS::_cpu_revision = 999;
```
kasperl, line 38: Why is this a reasonable default value? -1? 0? Add a comment if 999 makes sense.

src/os_esp32.cc
```diff
@@ -50,7 +53,7 @@ namespace toit {
 
 // Flags used to get memory for the Toit heap, which needs to be fast and 8-bit
 // capable.
-static const int TOIT_HEAP_CAPS_FLAGS = MALLOC_CAP_8BIT | MALLOC_CAP_INTERNAL | MALLOC_CAP_DMA;
+static int toit_heap_caps_flags = 0;
```
kasperl, line 56: Extend comment to tell that this is dynamically computed and updated based on the capabilities of the platform we run on.

src/os_esp32.cc
```diff
@@ -338,6 +341,9 @@ void OS::set_up() {
   _global_mutex = allocate_mutex(0, "Global mutex");
   _scheduler_mutex = allocate_mutex(4, "Scheduler mutex");
   _resource_mutex = allocate_mutex(99, "Resource mutex");
+#ifdef CONFIG_IDF_TARGET_ESP32
+  _cpu_revision = esp_efuse_get_chip_ver();
```
kasperl, line 345: What can this return?

src/os_esp32.cc
```diff
@@ -381,8 +389,25 @@ bool OS::use_virtual_memory(void* address, uword size) {
...
 OS::HeapMemoryRange OS::get_heap_memory_range() {
-  multi_heap_info_t info {};
-  heap_caps_get_info(&info, TOIT_HEAP_CAPS_FLAGS);
+  multi_heap_info_t info = { 0 };
+
+  toit_heap_caps_flags = MALLOC_CAP_8BIT | MALLOC_CAP_SPIRAM;
+  heap_caps_get_info(&info, toit_heap_caps_flags);
+
+  if (info.lowest_address == null || _cpu_revision < 3) {
+    if (info.lowest_address != null) {
```
kasperl, line 398: I feel like this should be the cpu revision check. Maybe put it in a bool.

```
bool cpu_supports_spiram = _cpu_revision >= 3;
```

src/os_esp32.cc
```diff
@@ -381,8 +389,25 @@ bool OS::use_virtual_memory(void* address, uword size) {
...
+      printf("[toit] INFO: SPIRAM not supported on CPU revision %d, using only internal RAM\n", _cpu_revision);
+    }
+    // If there is no SPI RAM try again with flags for internal RAM.
+    toit_heap_caps_flags = MALLOC_CAP_8BIT | MALLOC_CAP_INTERNAL | MALLOC_CAP_DMA;
+    heap_caps_get_info(&info, toit_heap_caps_flags);
+  } else {
+    uword lo = reinterpret_cast<uword>(info.lowest_address);
+    uword hi = reinterpret_cast<uword>(info.highest_address);
+    // Round the reported SPIRAM to the nearest MB.
+    printf("[toit] INFO: Using SPIRAM %dMbytes for heap.\n",
```
kasperl, line 408: Do we typically print this as Mbytes? How do we print kilobytes?

src/os_esp32.cc
```diff
@@ -393,6 +418,8 @@ OS::HeapMemoryRange OS::get_heap_memory_range() {
     return range;
   }
 
+  // In this case used hard coded ranges for internal RAM.
+
```
kasperl, line 422: Drop newline.

src/os_esp32.cc
```diff
@@ -393,6 +418,8 @@ OS::HeapMemoryRange OS::get_heap_memory_range() {
     return range;
   }
 
+  // In this case used hard coded ranges for internal RAM.
```
kasperl, line 421:
```suggestion
  // In this case use hard coded ranges for internal RAM.
```

toolchains/esp32/sdkconfig
```diff
@@ -378,6 +379,7 @@ CONFIG_ESP_TLS_USING_MBEDTLS=y
 #
 # ESP32-specific
 #
+CONFIG_ESP32_ECO3_CACHE_LOCK_FIX=y
```
kasperl, line 382: Do we need this?

kasperl (approved):
Looks good to me.

PR #958, opened 2022-08-17
Add popcount and move bit manipulation to int class

erikcorry:


lib/core/numbers.toit
```diff
@@ -971,6 +971,62 @@ abstract class int extends num:
...
+  int.MAX.count_leading_zeros   // => 1
+  ```
+  */
+  count_leading_zeros -> int:
+    #primitive.core.count_leading_zeros
+
+  /**
+  Returns the number of trailing zeros in the binary representation of the
+    integer.
+  The inteter is treated as an unsigned 64 bit number.
```
kasperl, line 993:
```suggestion
  The integer is treated as an unsigned 64 bit number.
```

kasperl (approved):
Looks good to me.

PR #959, opened 2022-08-17
Enable SPIRAM workarounds.

erikcorry:
I was under the impression that these were unnecessary for
CPU revision 3 and up, but I get crashes when locating the
heap (not the metadata) in SPIRAM unless I have these flags.


src/third_party/dartino/gc_metadata.cc
```diff
@@ -63,6 +63,11 @@ void GcMetadata::set_up_singleton() {
   // page boundaries.
   metadata_ = reinterpret_cast<uint8*>(OS::grab_virtual_memory(null, metadata_size_));
 
+  if (metadata_ == null) {
+    printf("[Toit] ERROR Failed to allocate GC metadata\n");
```
kasperl, line 67: I don't think this is the conventional format for our messages.

```suggestion
    printf("[toit] ERROR: failed to allocate GC metadata\n");
```

kasperl (approved):
Looks good to me.

PR #966, opened 2022-08-18
Set wifi to no-SPIRAM mode if the SPI is not present

erikcorry:


src/resources/wifi_esp32.cc
```diff
@@ -250,6 +251,13 @@ uint32 WifiResourceGroup::on_event(Resource* resource, word data, uint32 state)
 
 MODULE_IMPLEMENTATION(wifi, MODULE_WIFI)
 
+static bool has_external_psram() {
```
kasperl, line 254: Base this on the decision to use SPIRAM for Toit heaps.

kasperl (approved):
Looks good to me.

src/os.h
```diff
@@ -277,6 +277,10 @@ class HeapTagScope {
   uword old;
 };
 
+#ifdef TOIT_FREERTOS
+extern bool use_spiram_for_heap;
```
kasperl, line 281: Maybe this should be accessed through a static method on OS? Seems more consistent with the rest of the system.

kasperl (approved):
Still looks good to me.

src/os_esp32.cc
```diff
@@ -54,7 +54,18 @@ namespace toit {
 // Flags used to get memory for the Toit heap, which needs to be fast and 8-bit
 // capable.  We will set this to the most useful value when we have detected
 // which types of RAM are available.
-static int toit_heap_caps_flags = 0;
+bool OS::_use_spiram_for_heap = false;
+
+static int EXTERNAL_CAPS = MALLOC_CAP_8BIT | MALLOC_CAP_SPIRAM;
```
kasperl, line 59: const?

src/resources/wifi_esp32.cc
```diff
@@ -17,6 +17,7 @@
 
 #ifdef TOIT_FREERTOS
 
+#include <esp_heap_caps.h>
```
kasperl, line 20: Drop include again.

kasperl (approved):
Still looks good.

PR #968, opened 2022-08-18
Infer the types of globals.

floitsch:


src/compiler/type_check.cc
```diff
@@ -103,12 +107,61 @@ class TypeChecker : public ReturningVisitor<Type> {
...
+    bool detected_cycle = _globals_cycle_detector.check_cycle(node, [&](const std::vector<ir::Global*>& cycle) {
+      report_cyclic_global_types(cycle);
+      _reported_cyclic_globals.insert(node);
+    });
+    if (detected_cycle) return Type::invalid();
+    _method = node;
+    // TODO(florian): this is a bit hacky, but we have already rewritten the expression of
+    // the global, so we need to extract it now again.
+    auto body = node->body();
+    if (!body->is_Sequence()) UNREACHABLE();
```
kasperl, line 129: Should we have a CHECK macro of sorts?

src/compiler/type_check.cc
```diff
@@ -103,12 +107,61 @@ class TypeChecker : public ReturningVisitor<Type> {
...
+    auto expressions = body->as_Sequence()->expressions();
+    if (expressions.length() != 1) UNREACHABLE();
```
kasperl, line 131: Should we have a CHECK macro of sorts?

src/compiler/type_check.cc
```diff
@@ -103,12 +107,61 @@ class TypeChecker : public ReturningVisitor<Type> {
...
+    auto last = expressions.last();
+    if (last->is_CallStatic()) {
+      // Call to "uninitialized_global_failure_".
+      if (last->as_CallStatic()->target()->as_ReferenceMethod()->target()->name() != Symbols::uninitialized_global_failure_) UNREACHABLE();
```
kasperl, line 135: Should we have a CHECK macro of sorts?

src/compiler/type_check.cc
```diff
@@ -103,12 +107,61 @@ class TypeChecker : public ReturningVisitor<Type> {
     if (node->has_body()) visit(node->body());
     return Type::invalid();
   }
+  Type visit_Global(Global* node) {
```
kasperl, line 110: All these methods always return Type::invalid(), so they are only after the side effects. It seems a little bit surprising at first glance, so maybe it makes sense to think of a good place to document that?

third_party/benchmarks/toit/pystone.toit
```diff
@@ -30,8 +30,8 @@ class Record:
   copy -> Record:
     return Record PtrComp Discr EnumComp IntComp StringComp
 
-TRUE ::= 1
-FALSE ::= 0
+TRUE ::= true
```
kasperl, line 33: Why is this change necessary? I'm just curious.

src/compiler/type_check.cc
```diff
@@ -431,9 +498,13 @@ class TypeChecker : public ReturningVisitor<Type> {
   }
 
   Type visit_Return(Return* node) {
-    auto return_type = visit(node->value());
+    auto value_type = visit(node->value());
+    return visit_Return(node, value_type);
+  }
+
+  Type visit_Return(Return* node, Type value_type) {
```
kasperl, line 505: Are there other callers (not just the `visit_Return(Return*)` above) of this already?

kasperl (approved):
Looks good to me.

src/compiler/type_check.cc
floitsch: added a comment.
done.

src/compiler/type_check.cc
floitsch: added TOIT_CHECK
done.

src/compiler/type_check.cc
floitsch: added `TOIT_CHECK`.
done.

src/compiler/type_check.cc
floitsch: added `TOIT_CHECK`.
done.

src/compiler/type_check.cc
```diff
@@ -103,12 +107,61 @@ class TypeChecker : public ReturningVisitor<Type> {
...
+      // The uninitialized_global_failure_ call references its own global recursively.
+      // Mark the node as handled already now and give it the 'any' type.
+      // Alternatively, we could also just not visit the body.
+      _handled_globals.insert(node);
+      node->set_return_type(Type::any());
+      visit(node->body());
+      return Type::invalid();
+    }
+    _globals_cycle_detector.start(node);
+    if (!last->is_Return()) UNREACHABLE();
```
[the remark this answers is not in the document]
floitsch: added `TOIT_CHECK`.
done.

src/compiler/type_check.cc
floitsch: not anymore.
Initially I made globals still call the `visit_Return` but I removed that.
reverted.

third_party/benchmarks/toit/pystone.toit
floitsch: In the old code:
The BoolGlob of line 44 is initialized with `FALSE` which is inferred to be of type `int`.
However, line 71 tries to assign a boolean (the result of `not`) to it. As a consequence the type-checker complains.

PR #972, opened 2022-08-19
Fix Toit/mbedtls define mismatch.

floitsch:
The Toit code and the mbedtls code did not agree on the defines that
were set. This meant that Toit believed that some of the mbedtls
structures were smaller than they needed to be. The mbedtls code was
writing into fields that were outside the allocated structure,
overwriting neighboring objects.


src/resources/wifi_esp32.cc
```diff
@@ -17,6 +17,7 @@
 
 #ifdef TOIT_FREERTOS
 
+#include <esp_heap_caps.h>
```
kasperl, line 20: Is this still needed?

CMakeLists.txt
```diff
@@ -120,7 +120,15 @@ endif()
 # sizes, but they just cause the other side to fragment packets more agressively.
 # At this end we still have to reassemble the packets in and unfragmented form
 # that fits in the buffer size below.
-set(MBEDTLS_C_FLAGS "-DMBEDTLS_SSL_IN_CONTENT_LEN=8100 -DMBEDTLS_SSL_OUT_CONTENT_LEN=3700 -DMBEDTLS_PLATFORM_MEMORY=1")
+if ("${TOIT_SYSTEM_NAME}" MATCHES "esp32")
+  # No need to set the flags, the same values are in the sdkconfig.
+  # However, we need to include the esp32 config file to make sure that Toit and the compiled
+  # mbedtls sources agree on the defines.
```
kasperl, line 126: MbedTLS for consistency.
```suggestion
  # MbedTLS sources agree on the defines.
```

kasperl (approved):
Looks good to me.

src/resources/wifi_esp32.cc
floitsch: I don't even remember that I added it... Maybe an accidental "quick fix" from the IDE.
removed; done.

CMakeLists.txt
floitsch: done.

PR #979, opened 2022-08-23
Attempt to work around spurious out-of-memory on M1 Mac.

erikcorry:
If this doesn't work we may need to reserve the memory
at the start instead of trying to guess where it will
be allocated.

### DIFF
--- a/src/os.cc
+++ b/src/os.cc
@@ -203,7 +203,7 @@ OS::HeapMemoryRange OS::get_heap_memory_range() {
     // anything in that range.
     const uword FOUR_GB = 4LL * GB;
     if (addr < FOUR_GB && addr + MAX_HEAP > FOUR_GB) {
-      _single_range.address = FOUR_GB;
+      _single_range.address = reinterpret_cast<void*>(FOUR_GB);
     } else {
 #else
     {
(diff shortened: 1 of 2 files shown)


src/os.cc
```diff
@@ -197,9 +197,23 @@ OS::HeapMemoryRange OS::get_heap_memory_range() {
     // be the last MAX_HEAP of the address space.
     _single_range.address = reinterpret_cast<void*>(-static_cast<word>(MAX_HEAP + TOIT_PAGE_SIZE));
   } else {
-    // We will be allocating within a symmetric range either side of this
-    // single allocation.
-    _single_range.address = reinterpret_cast<void*>(addr - HALF_MAX);
+    uword from = addr - MAX_HEAP / 2;
+#if defined(TOIT_DARWIN) && defined(BUILD_64)
+    uword to = addr + MAX_HEAP / 2;
```
kasperl, line 202:
```suggestion
    uword to = from + MAX_HEAP;
```

src/os.cc
```diff
@@ -197,9 +197,23 @@ OS::HeapMemoryRange OS::get_heap_memory_range() {
...
+    // MacOS never returns addresses in the first 4Gbytes, in order to flush
```
kasperl, line 203:
```suggestion
    // On macOS, we never get addresses in the first 4Gbytes, in order to flush
```

src/os.cc
```diff
@@ -197,9 +197,23 @@ OS::HeapMemoryRange OS::get_heap_memory_range() {
...
+    // out 32 bit uncleanness, so let's try to avoid having the range cover
+    // anything in that range.
```
kasperl, line 205:
```suggestion
    // both sides of the 4Gbytes boundary.
```

kasperl (approved):
Looks good to me.

PR #980, opened 2022-08-23
Add Channel.send with a block.

floitsch:


lib/monitor.toit
```diff
@@ -256,6 +256,20 @@ monitor Channel:
     buffer_[index] = value
     size_++
 
+  /**
+  Sends a message with the result of calling the given $block.
+  This operation may block if the buffer capacity has been reached. In that
+    case, this task waits until another task calls $receive. The block is
+    only called when the channel has the capacity to receive a message.
```
kasperl, line 263:
```suggestion
    only called when the channel has the capacity to buffer a message.
```

kasperl (approved):
Looks good to me.

PR #982, opened 2022-08-25
Add version to snapshot

kasperl:


erikcorry (approved):
LGTM but lets encode the length of the version too.

PR #983, opened 2022-08-25
Add UART support for darwin and added control of RTS/DTR

mikkeldamsgaard:
Support for UART on Darwin and for reading and writing control lines

### DIFF
--- a/lib/uart.toit
+++ b/lib/uart.toit
@@ -102,7 +102,7 @@ class Port implements reader.Reader:
 
   This constructor does not work on embedded devices, such as the ESP32.
 
-  The $baud_rate must match one that is supported by the operating system. See $Port.baud_rate=.
+  On some platforms the $baud_rate must match one that is supported by the operating system. See $Port.baud_rate=.
   */
   constructor device/string
       --baud_rate/int
@@ -131,6 +131,8 @@ class Port implements reader.Reader:
     following baud rates are supported: 50, 75, 110, 134, 150, 200, 300, 600, 1200,
     1800, 2400, 4800, 9600, 19200, 38400, 57600, 115200, 230400, 460800, 576000, 921600,
     1152000, 1500000, 2000000, 2500000, 3000000, 3500000, 4000000.
+
+  On macos the baud rate can be arbitrary
   */
   baud_rate= new_rate/int:
     uart_set_baud_rate_ uart_ new_rate
@@ -217,6 +219,21 @@ class Port implements reader.Reader:
       if flushed: return
       sleep --ms=1
 
+/**
+Extends the functionality of the UART Port on platforms that support configuratble RS232 devices. It allows setting
+and reading control lines.
+*/
+
+class ConfigurableDevicePort extends Port:
+  /**
+  See super
+  */
+  constructor device/string
+      --baud_rate/int
+      --data_bits/int=8
+      --stop_bits/int=Port.STOP_BITS_1
+      --parity/int=Port.PARITY_DISABLED:
+    super device --baud_rate=baud_rate --data_bits=data_bits --stop_bits=stop_bits --parity=parity
 
   static CONTROL_FLAG_LE  ::= 1 << 0            /* line enable */
   static CONTROL_FLAG_DTR ::= 1 << 1            /* data terminal ready */
@@ -232,8 +249,6 @@ class Port implements reader.Reader:
   Read the value of the given control $flag. $flag must be one of the CONTROL_ constants.
 
   Returns the state of the $flag
-
-  Note: Works only on uarts constructed from a device, i.e. in host mode. It is not avaiable on the ESP32.
   */
   read_control_flag flag/int -> bool:
     return (uart_get_control_flags_ uart_) & flag != 0
@@ -241,16 +256,12 @@ class Port implements reader.Reader:
   /**
   Read the value of all the control flags. Each bit in the returned value corresponds to the bit position indicated
   by the CONTROL_ constants.
-
-  Note: Works only on uarts constructed from a device, i.e. in host mode. It is not avaiable on the ESP32.
   */
   read_control_flags -> int:
     return uart_get_control_flags_ uart_
 
   /**
   Sets the $state of a control $flag. $flag must be one of the CONTROL_ constants.
-
-  Note: Works only on uarts constructed from a device, i.e. in host mode. It is not avaiable on the ESP32.
   */
   set_control_flag flag/int state/bool:
     flags := uart_get_control_flags_ uart_
@@ -262,8 +273,6 @@ class Port implements reader.Reader:
 
   /**
   Sets all control $flags to the specified value. Each bit in the $flags corresponds to one of the CONTROL_ constants.
-
-  Note: Works only on uarts constructed from a device, i.e. in host mode. It is not avaiable on the ESP32.
   */
   set_control_flags flags/int:
     uart_set_control_flags_ uart_ flags
--- a/lib/uart.toit
+++ b/lib/uart.toit
@@ -226,7 +226,7 @@ and reading control lines.
 
 class ConfigurableDevicePort extends Port:
   /**
-  See $super.
+    See super class constructor.
   */
   constructor device/string
       --baud_rate/int
--- a/lib/uart.toit
+++ b/lib/uart.toit
@@ -217,6 +217,57 @@ class Port implements reader.Reader:
       if flushed: return
       sleep --ms=1
 
+
+  static CONTROL_FLAG_LE  ::= 1 << 0            /* line enable */
+  static CONTROL_FLAG_DTR ::= 1 << 1            /* data terminal ready */
+  static CONTROL_FLAG_RTS ::= 1 << 2            /* request to send */
+  static CONTROL_FLAG_ST  ::= 1 << 3            /* secondary transmit */
+  static CONTROL_FLAG_SR  ::= 1 << 4            /* secondary receive */
+  static CONTROL_FLAG_CTS ::= 1 << 5            /* clear to send */
+  static CONTROL_FLAG_CAR ::= 1 << 6            /* carrier detect */
+  static CONTROL_FLAG_RNG ::= 1 << 7            /* ring */
+  static CONTROL_FLAG_DSR ::= 1 << 8            /* data set ready */
+
+  /**
+  Read the value of the given control $flag. $flag must be one of the CONTROL_ constants.
+
+  Returns the state of the $flag
+
+  Note: Works only on uarts constructed from a device, i.e. in host mode. It is not avaiable on the ESP32.
+  */
+  read_control_flag flag/int -> bool:
+    return (uart_get_control_flags_ uart_) & flag != 0
+
+  /**
+  Read the value of all the control flags. Each bit in the returned value corresponds to the bit position indicated
+  by the CONTROL_ constants.
+
+  Note: Works only on uarts constructed from a device, i.e. in host mode. It is not avaiable on the ESP32.
+  */
+  read_control_flags -> int:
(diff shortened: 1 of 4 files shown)


src/resources/uart_posix.cc
```diff
@@ -217,9 +234,16 @@ class UARTResourceGroup : public ResourceGroup {
 
  private:
   uint32_t on_event(Resource* resource, word data, uint32_t state) {
+#ifdef TOIT_LINUX
     if (data & EPOLLIN) state |= kReadState;
     if (data & EPOLLERR) state |= kErrorState;
     if (data & EPOLLOUT) state |= kWriteState;
+#elifdef TOIT_BSD
+    auto *event = reinterpret_cast<struct kevent*>(data);
```
erikcorry, line 242:
```suggestion
    auto event = reinterpret_cast<struct kevent*>(data);
```

lib/uart.toit
```diff
@@ -217,6 +217,57 @@ class Port implements reader.Reader:
       if flushed: return
       sleep --ms=1
 
+
+  static CONTROL_FLAG_LE  ::= 1 << 0            /* line enable */
```
erikcorry, line 221: It's not a great user experience for embedded programmers to have all these methods and constants that are suggested by the editor's autocomplete, but that don't work because the ESP32 doesn't have support.

How about making a subclass, perhaps called ConfigurablePort, that has the extra constants and methods?

src/resources/uart_posix.cc
```diff
@@ -297,13 +325,21 @@ PRIMITIVE(set_baud_rate) {
...
+  if (int_to_baud_rate(baud_rate, &speed) != 0) INVALID_ARGUMENT;
   struct termios tty;
   if (tcgetattr(fd, &tty) != 0) return Primitive::os_error(errno, process);
   if (cfsetospeed(&tty, speed) != 0) return Primitive::os_error(errno, process);
   if (cfsetispeed(&tty, speed) != 0) return Primitive::os_error(errno, process);
   // TCSADRAIN: let the change happen once all output written to the fd has been transmitted.
-  if (tcsetattr(fd, TCSADRAIN, &tty) != 0) return Primitive::os_error(errno, process);
+  if (tcsetattr(fd, TCSADRAIN, &tty) != 0) {
+#ifdef TOIT_DARWIN
+    // Try setting baud rate with ioctl
```
erikcorry, line 336:
```suggestion
    // Try setting baud rate with ioctl.
```

src/resources/uart_posix.cc
```diff
@@ -103,6 +117,9 @@ static int int_to_linux_baud_rate(int baud_rate, speed_t* speed) {
     case 3500000: *speed = B3500000; return 0;
     case 4000000: *speed = B4000000; return 0;
     default: return -1;
+#elifdef TOIT_DARWIN
+    default: *speed = baud_rate; return 0;
```
erikcorry, line 121: In theory there could be a clash between the baud rate and the Bxxx defines?  Perhaps should set a flag indicating that we should skip tcgetattr/tcsetattr and go straight to IOSSIOSPEED.

lib/uart.toit
```diff
@@ -131,6 +131,8 @@ class Port implements reader.Reader:
     following baud rates are supported: 50, 75, 110, 134, 150, 200, 300, 600, 1200,
     1800, 2400, 4800, 9600, 19200, 38400, 57600, 115200, 230400, 460800, 576000, 921600,
     1152000, 1500000, 2000000, 2500000, 3000000, 3500000, 4000000.
+
+  On macos the baud rate can be arbitrary
```
erikcorry, line 135:
```suggestion
  On macOS the baud rate can be arbitrary.
```

src/resources/uart_posix.cc
```diff
@@ -68,10 +76,16 @@ static int linux_baud_rate_to_int(speed_t speed) {
...
+    default:
+      return (int)speed;
+#endif
   }
 }
 
-static int int_to_linux_baud_rate(int baud_rate, speed_t* speed) {
+static int int_to_baud_rate(int baud_rate, speed_t* speed, int *arbitrary_baud_rate) {
+  // TODO: On linux using gcc, it should be possible to just set the bit rate as an integer
+  *arbitrary_baud_rate = 0;
```
erikcorry, line 88:
```suggestion
  *arbitrary_baud_rate = false;
```

src/resources/uart_posix.cc
```diff
@@ -103,6 +118,12 @@ static int int_to_linux_baud_rate(int baud_rate, speed_t* speed) {
     case 3500000: *speed = B3500000; return 0;
     case 4000000: *speed = B4000000; return 0;
     default: return -1;
+#elifdef TOIT_DARWIN
+    default:
+      *speed = baud_rate;
+      *arbitrary_baud_rate = 1;
```
erikcorry, line 124:
```suggestion
      *arbitrary_baud_rate = true;
```

src/resources/uart_posix.cc
```diff
@@ -248,7 +280,8 @@ PRIMITIVE(create_path) {
   ARGS(UARTResourceGroup, resource_group, cstring, path, int, baud_rate, int, data_bits, int, stop_bits, int, parity);
 
   speed_t speed;
-  if (int_to_linux_baud_rate(baud_rate, &speed) != 0) INVALID_ARGUMENT;
+  int arbitrary_baud_rate;
```
erikcorry, line 283:
```suggestion
  bool arbitrary_baud_rate;
```

src/resources/uart_posix.cc
```diff
@@ -297,13 +330,24 @@ PRIMITIVE(set_baud_rate) {
...
 
   speed_t speed;
-  if (int_to_linux_baud_rate(baud_rate, &speed) != 0) INVALID_ARGUMENT;
-  struct termios tty;
-  if (tcgetattr(fd, &tty) != 0) return Primitive::os_error(errno, process);
-  if (cfsetospeed(&tty, speed) != 0) return Primitive::os_error(errno, process);
-  if (cfsetispeed(&tty, speed) != 0) return Primitive::os_error(errno, process);
-  // TCSADRAIN: let the change happen once all output written to the fd has been transmitted.
-  if (tcsetattr(fd, TCSADRAIN, &tty) != 0) return Primitive::os_error(errno, process);
+  int arbitrary_rate;
```
erikcorry, line 333:
```suggestion
  bool arbitrary_rate;
```

src/resources/uart_posix.cc
```diff
@@ -297,13 +330,24 @@ PRIMITIVE(set_baud_rate) {
...
+  int result = int_to_baud_rate(baud_rate, &speed, &arbitrary_rate);
+  if (result != 0) INVALID_ARGUMENT;
+  if (!arbitrary_rate) { // 0 means use standard posix/linux line speed setup
```
erikcorry, line 336:
```suggestion
  if (!arbitrary_rate) {  // false means use standard Posix/Linux line speed setup.
```

src/resources/uart_posix.cc
```diff
@@ -68,10 +76,16 @@ static int linux_baud_rate_to_int(speed_t speed) {
...
+#elifdef TOIT_DARWIN
+    default:
+      return (int)speed;
+#endif
   }
 }
 
-static int int_to_linux_baud_rate(int baud_rate, speed_t* speed) {
+static int int_to_baud_rate(int baud_rate, speed_t* speed, int *arbitrary_baud_rate) {
+  // TODO: On linux using gcc, it should be possible to just set the bit rate as an integer
```
erikcorry, line 87:
```suggestion
  // TODO: On linux using gcc, it should be possible to just set the bit rate as an integer.
```

erikcorry (approved):

erikcorry:
Build is failing because "See $super" doesn't work in comments.

lib/uart.toit
```diff
@@ -217,6 +219,64 @@ class Port implements reader.Reader:
       if flushed: return
       sleep --ms=1
 
+/**
+Extends the functionality of the UART Port on platforms that support configuratble RS232 devices. It allows setting
```
floitsch, line 223:
```suggestion
Extends the functionality of the UART Port on platforms that support configurable RS232 devices. It allows setting
```

lib/uart.toit
```diff
@@ -217,6 +219,64 @@ class Port implements reader.Reader:
...
+and reading control lines.
```
floitsch, line 224:
```suggestion
  and reading control lines.
```

src/resources/uart_posix.cc
```diff
@@ -68,10 +76,16 @@ static int linux_baud_rate_to_int(speed_t speed) {
     case B3500000: return 3500000;
     case B4000000: return 4000000;
     default: return -1;
+#elif defined(TOIT_DARWIN)
+    default:
+      return (int)speed;
```
floitsch, line 81:
```suggestion
      return static_cast<int>(speed);
```

src/resources/uart_posix.cc
```diff
@@ -297,13 +330,24 @@ PRIMITIVE(set_baud_rate) {
...
-  struct termios tty;
-  if (tcgetattr(fd, &tty) != 0) return Primitive::os_error(errno, process);
-  if (cfsetospeed(&tty, speed) != 0) return Primitive::os_error(errno, process);
-  if (cfsetispeed(&tty, speed) != 0) return Primitive::os_error(errno, process);
-  // TCSADRAIN: let the change happen once all output written to the fd has been transmitted.
-  if (tcsetattr(fd, TCSADRAIN, &tty) != 0) return Primitive::os_error(errno, process);
+  bool arbitrary_rate;
+  int result = int_to_baud_rate(baud_rate, &speed, &arbitrary_rate);
+  if (result != 0) INVALID_ARGUMENT;
+  if (!arbitrary_rate) { // false means use standard Posix/Linux line speed setup
```
floitsch, line 336:
```suggestion
  if (!arbitrary_rate) {
    // Use standard Posix/Linux line speed setup.
```

lib/uart.toit
```diff
@@ -217,6 +219,64 @@ class Port implements reader.Reader:
...
+*/
+
+class ConfigurableDevicePort extends Port:
```
floitsch, line 227: Could we make this the `HostPort` class (someone has a better name?) or something similar?
The `Port` constructor that takes a device string could then be a factory constructor returning an instance of this class. (We would need a private constructor in the `Port` class to make this work).
This would make me more comfortable with having a new class.

lib/uart.toit
```diff
@@ -217,6 +219,64 @@ class Port implements reader.Reader:
       if flushed: return
       sleep --ms=1
 
+/**
+Extends the functionality of the UART Port on platforms that support configuratble RS232 devices. It allows setting
+and reading control lines.
+*/
+
```
floitsch, line 226:
```suggestion
```

floitsch (approved):
Nice.

lib/uart.toit
mikkeldamsgaard: With this suggestion:
`port/HostPort := Port "/dev/ttyUSB1"` fails to compile.
`port/HostPort := (Port "/dev/ttyUSB1") as HostPort` compiles. 
Seems a bit convoluted

floitsch: We could still leave the `HostPort` constructor public. So if you actually wanted the `HostPort` directly, you can go there.

Just feels unnecessary to not give the host-functionality, just because you didn't go through the `HostPort` constructor.

mikkeldamsgaard:
@floitsch Can you kick off the tests agin?

floitsch:
Kicked off the tests. Currently the buildbot is a bit overloaded, so it will take a bit of time.

floitsch:
Thanks.

PR #992, opened 2022-09-02
Update external health gold

kasperl:


kasperl:
TBR'ed and replaces #991.

PR #995, opened 2022-09-02
Fix free memory percent on heap reports with PSRAM.

erikcorry:
Also some formatting improvements, especially when
dumping heap info for individual pages of memory:
* Each page gets its own line drawn box.
* Bignum is renamed to Bignum/TLS.
* We call it "untagged allocations" when we don't know.
* Some ASCII pipes replaced with real line drawing chars.
* Each page has a heading "Page:" to make it clearer.

### DIFF
--- a/src/os_esp32.cc
+++ b/src/os_esp32.cc
@@ -719,7 +719,7 @@ class HeapSummaryCollector {
       pages_[i].print();
     }
     if (dropped_pages_ > 0) {
-      printf("\n  %d unreported pages, hit limit of %-4d \n", dropped_pages_, max_pages_);
+      printf("\n  %d unreported pages, hit limit of %d.\n", dropped_pages_, max_pages_);
     }
   }
 
--- a/src/os_esp32.cc
+++ b/src/os_esp32.cc
@@ -578,26 +578,27 @@ class HeapSummaryPage {
 
   void print() {
     if (address_ == 0) return;
-    printf("  ├────────────┼─────────────────────────────────────────────┤\n");
-    printf("  │ %p │   Largest free = %-5d                      │\n",
-        reinterpret_cast<void*>(address_), largest_free_);
-    printf("  │            ├───────────┬─────────┬───────────────────────┤\n");
+    printf("  ┌────────────┬─────────────────────────────────────────────┐\n");
+    printf("  │ Page:      │   Largest free = %-5d                      │\n",
+        largest_free_);
+    printf("  │ %p ├───────────┬─────────┬───────────────────────┤\n",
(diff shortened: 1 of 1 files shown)


src/os_esp32.cc
```diff
@@ -607,7 +608,7 @@ class HeapSummaryPage {
       case EVENT_SOURCE_MALLOC_TAG: return "event source";
       case OTHER_THREADS_MALLOC_TAG: return "other threads";
       case THREAD_SPAWN_MALLOC_TAG: return "thread spawn";
-      case NULL_MALLOC_TAG: return "null tag";
+      case NULL_MALLOC_TAG: return "untagged allocations";
```
kasperl, line 611: They are all allocations. Maybe just untagged?
```suggestion
      case NULL_MALLOC_TAG: return "untagged";
```

src/os_esp32.cc
```diff
@@ -578,26 +578,27 @@ class HeapSummaryPage {
...
-    printf("  ├────────────┼───────────┴─────────┴───────────────────────┤\n");
+    printf("  └────────────┴───────────┴─────────┴───────────────────────┘\n");
   }
 
   static const char* name_of_type(int tag) {
     switch (tag) {
       case MISC_MALLOC_TAG: return "misc";
       case EXTERNAL_BYTE_ARRAY_MALLOC_TAG: return "external byte array";
-      case BIGNUM_MALLOC_TAG: return "bignum";
+      case BIGNUM_MALLOC_TAG: return "bignum/TLS";
```
kasperl, line 601: I think I'd prefer this to be tls/bignum. TLS is the category sort of.
```suggestion
      case BIGNUM_MALLOC_TAG: return "tls/bignum";
```

src/os_esp32.cc
```diff
@@ -607,7 +608,7 @@ class HeapSummaryPage {
       case EVENT_SOURCE_MALLOC_TAG: return "event source";
       case OTHER_THREADS_MALLOC_TAG: return "other threads";
```
kasperl, line 609: Why is this plural?
```suggestion
      case OTHER_THREADS_MALLOC_TAG: return "thread/other";
```

src/os_esp32.cc
```diff
@@ -607,7 +608,7 @@ class HeapSummaryPage {
...
       case THREAD_SPAWN_MALLOC_TAG: return "thread spawn";
```
kasperl, line 610:
```suggestion
      case THREAD_SPAWN_MALLOC_TAG: return "thread/spawn";
```

kasperl (approved):
Looks good to me.

PR #1000, opened 2022-09-06
Encode arguments to main as a serialized message

kasperl:


src/process.h
```diff
@@ -226,7 +228,7 @@ class Process : public ProcessListFromProcessGroup::Element,
   Method _entry;
   Method _spawn_method;
 
-  char** _main_arguments = null;
+  uint8* _main_arguments = null;
```
erikcorry, line 231: Add comment to say how this is encoded and perhaps how it is disposed.

src/scheduler.h
```diff
@@ -71,21 +71,24 @@ class Scheduler {
...
+    Program* program,
     SnapshotBundle system,  // It is then the responsibility of the system process to launch the application.
     SnapshotBundle application,
-    char** arguments,
+    char** argv,
     int group_id);
-#endif  // TOIT_FREERTOS
+#endif
 
   // Run a new program. Returns the process ID of the root process.
```
erikcorry, line 90: Also here.

lib/core/entry.toit
```diff
@@ -9,8 +9,11 @@ __entry__main -> none:
   current := Task_.current
   current.initialize_entry_task_
   current.evaluate_:
-    args := List_.from_array_ main_arguments_
-    #primitive.intrinsics.main args
+    // TODO(kasper): Do we want to handle arrays in arrays here and
+    // can we get away from mapping to lists in general?
+    arguments := main_arguments_
+    if arguments is Array_: arguments = List.from arguments
```
erikcorry, line 15: Maybe it's OK to pass the Array_

erikcorry (approved):

PR #1006, opened 2022-09-06
Improve error message on malformed domain name

erikcorry:


lib/net/modules/dns.toit
```diff
@@ -271,7 +271,7 @@ create_query name/string query_id/int -> ByteArray:
   parts := name.split "."
   length := 1
   parts.do: | part |
-    if part.size > 63: throw (DnsException "LABEL_TOO_LARGE")
+    if part.size > 63: throw (DnsException "DNS name sections cannot exceed 63 bytes")
```
kasperl, line 274: Are these really sections? It feels like they are name parts or components.

lib/net/modules/dns.toit
```diff
@@ -271,7 +271,7 @@ create_query name/string query_id/int -> ByteArray:
...
     if part.size < 1: throw (DnsException "LABEL_TOO_SHORT")
```
kasperl, line 275: Update this too? DNS name parts/components cannot be empty?

kasperl (approved):

kasperl (approved):
Still looks good to me.

PR #1008, opened 2022-09-06
Don't mark a process idle-since-gc after a scavenge

erikcorry:


kasperl (approved):
Looks good to me.

src/scheduler.cc
```diff
@@ -464,26 +464,24 @@ void Scheduler::gc(Process* process, bool malloc_failed, bool try_hard) {
     ProcessListFromScheduler targets;
     { Locker locker(_mutex);
       for (ProcessGroup* group : _groups) {
-        bool done = false;
         for (Process* target : group->processes()) {
+          if (target->program() == null) continue;  // External process.
```
kasperl, line 468: You also need this guard in line 498 (after your changes).

kasperl (changes requested):

kasperl (approved):
Looks good to me.

PR #1010, opened 2022-09-06
Add TPKG_SDK_VERSION for pkg downloads

floitsch:
Hardcode an SDK version number for external package downloads.
The inferred SDK number is lower than the one we need to download all
dependencies in the external packages. As a work-around we hard-code a
version number that works.


kasperl (approved):
Looks good to me.

PR #1019, opened 2022-09-08
Add trace service for stack traces et al

kasperl:


lib/system/api/trace.toit
```diff
@@ -0,0 +1,33 @@
...
+    logging, or storing the message.
+
+  Returns whether the message was handled and needs no further
+    processing from the system's built-in trace message handler.
+  */
+  handle_trace message/ByteArray -> bool
+  static HANDLE_TRACE_INDEX /int ::= 0
+  // TODO(kasper): It seems nice to always have the method index
+  // after the method definition to allow for documentation comments.
+  // This should be fixed across the code base.
```
floitsch, line 23: Good TODO, but I wouldn't keep it here.

floitsch (approved):
LGTM.

lib/system/api/trace.toit
kasperl: I'll remove this when I'm done changing all the other uses (in one go -- and as a separate PR).

PR #1021, opened 2022-09-08
Add timeouts to external tests.

floitsch:


kasperl (approved):
Looks good to me.

kasperl:
Maybe this is a bit too little for the non-slow tests?

floitsch:
Added more time for the mac.
We do something similar for our own tests.

kasperl (approved):
Still looks good.

PR #1023, opened 2022-09-09
Some small fixes to tracegc and heap accounting

erikcorry:
The heap iteration can now filter out IRAM so we don't have to filter it out with a hack.


kasperl (approved):
Looks good to me.

PR #1024, opened 2022-09-09
A primitive that delivers a pagewise malloc report.

erikcorry:


src/primitive_esp32.cc
```diff
@@ -389,6 +389,145 @@ PRIMITIVE(rtc_user_bytes) {
...
+
+  static const int MALLOC_MANAGED  = 1 << 0;
+  static const int TOIT            = 1 << 1;
+  static const int EXTERNAL        = 1 << 2;
+  static const int TLS             = 1 << 3;
+  static const int BUFFERS         = 1 << 4;
+  static const int MISC            = 1 << 5;
+  static const int MERGE_WITH_NEXT = 1 << 6;
+  static const int SIZE_SHIFT_LEFT =      7;
+  static const uword FLAG_MASK       = (1 << SIZE_SHIFT_LEFT) - 1;
```
kasperl, line 488: Alignment of the = seems a bit wobbly here.

lib/esp32.toit
```diff
@@ -135,3 +135,74 @@ There is only one RTC memory on the device, so all tasks or processes have
 rtc_user_bytes -> ByteArray:
   #primitive.esp32.rtc_user_bytes
 
+/**
```
kasperl, line 138: I think we should move towards encoding the bytes (ubjson) in the VM, so this becomes more similar to profiles, histograms and stack traces.

kasperl (approved):
Looks good to me.

PR #1026, opened 2022-09-10
Avoid (caught) lookup errors in trace handling

kasperl:


lib/system/trace.toit
```diff
@@ -18,8 +18,13 @@ send_trace_message message/ByteArray -> none:
   handled := false
   try:
     service := service_
-    if not service: service = service_ = (TraceServiceClient --no-open).open
-    handled = service.handle_trace message
```
kasperl, line 22: The common case is actually that `service` is null here, so I'd rather not get another exception thrown in that case.

floitsch (approved):
LGTM.

PR #1027, opened 2022-09-11
Add support for forwarding exceptions through latches

kasperl:


lib/monitor.toit
```diff
@@ -35,34 +35,44 @@ A latch that allows one task to wait until a value (object) has been provided
...
+      rethrow value.value value.trace
+    return value
 
   /**
   Sets the $value of the latch.
 
   Calling this method unblocks any task that is blocked in the $get method of
     the same instance, sending the $value to it.
   Future calls to $get return immediately and use this $value.
-  Must be called at most once for each instance of the monitor.
```
kasperl, line 57: We could also decide to enforce this. It might be a nice property in terms of being able to reason about values returned by get.

floitsch (approved):
LGTM.

PR #1028, opened 2022-09-12
Multi part page report

erikcorry:
Deliver the page report in sections so we can make a report
for a large memory or bigger pages.

I still intend to make this a jag-decode based feature, so the
primitive will disappear.


src/primitive_esp32.cc
```diff
@@ -456,6 +453,18 @@ class PageReport {
...
+  uword next_memory_base() const {
+    if (more_above_) {
+      return memory_base_ + GRANULARITY * PAGES;
+    } else {
+      return 0;
+    }
+  }
+
+  static const int GRANULARITY_LOG2 = TOIT_PAGE_SIZE_LOG2;
+  static const uword GRANULARITY = 1 << GRANULARITY_LOG2;
```
kasperl, line 465: Should everything but `GRANULARITY` be private?

kasperl (approved):
Looks good to me.

PR #1039, opened 2022-09-19
Fix if condition variable.

floitsch:
I didn't fully investigate, but it was almost certainly a reuse of a stack slot.


kasperl (approved):
Looks good to me. 

I actually encountered this 3 days ago, but I couldn't find a short repro.

PR #1040, opened 2022-09-19
Use '-g' in debug mode.

floitsch:


floitsch:
I accidentally committed this change a while ago.

The better way to get debug-built code is:
```
rm -rf build
LOCAL_CXXFLAGS="-O0" make BUILD_TYPE=Debug
```
(and use the same line when calling `make` again).

PR #1050, opened 2022-09-23
Update external gold

kasperl:


floitsch (approved):
LGTM.
Thanks. I need to remember to do that when I update the mqtt package.

PR #1051, opened 2022-09-24
Wrap all firmware assets in the envelope

kasperl:


tools/firmware.toit
```diff
@@ -60,26 +74,31 @@ main arguments/List:
...
-  input_path := parsed["binary"]
+  input_path := parsed["firmware.bin"]
 
   // TODO(kasper): Do some sanity checks on the
   // structure of this. Can we check that we don't
   // already have stuff appended to the DROM section?
-  binary_data := file.read_content input_path
-  binary := Esp32Binary binary_data
+  firmware_bin_data := file.read_content input_path
+  Esp32Binary firmware_bin_data
```
floitsch, line 94: Is this just to check that the data is correct?

tools/firmware.toit
```diff
@@ -236,44 +255,61 @@ properties_update parsed/cli.Parsed [block] -> none:
     if properties: envelope.entries[AR_ENTRY_PROPERTIES] = json.encode properties
 
 extract_cmd -> cli.Command:
+  flags := AR_ENTRY_FILE_MAP.map: | key/string value/string |
```
floitsch, line 258: Alternative is to make each of those options:
`extract --firmware.bin=xyz --partitions.csv=...`

That would also avoid the exclusivity check in line 277.

floitsch (approved):
LGTM.

tools/firmware.toit
kasperl: Yes. It is what I intend to extend with more validation.

tools/firmware.toit
kasperl: Let's discuss the options next week. I'm all for further tweaks.

PR #1053, opened 2022-09-25
Support bundling program assets

kasperl:


tools/assets.toit
```diff
@@ -0,0 +1,115 @@
...
+  store_assets output_path existing
+
+load_assets path/string -> Map:
+  bytes := file.read_content path
+  return assets.decode bytes
+
+store_assets path/string entries/Map -> none:
+  output_stream := file.Stream.for_write path
+  bytes := assets.encode entries
+  output_stream.write bytes
```
floitsch, line 114: In theory this needs a writer.
In practice it always writes everything in one go.

tools/assets.toit
```diff
@@ -0,0 +1,115 @@
...
+  return cli.Command "create"
+      --run=:: create_assets it
+
+create_assets parsed/cli.Parsed -> none:
+  output_path := parsed[OPTION_ASSETS]
+  store_assets output_path {:}
+
+add_cmd -> cli.Command:
+  return cli.Command "add"
+      --options=[ option_output ]
```
floitsch, line 57: Add help stating that existing assets of the same name are overwritten.

tools/assets.toit
```diff
@@ -0,0 +1,115 @@
...
+      --run=:: remove_asset it
+
+remove_asset parsed/cli.Parsed -> none:
+  name := parsed["name"]
+  update_assets parsed: | entries/Map |
+    entries.remove name
+
+list_cmd -> cli.Command:
+  return cli.Command "list"
+      --run=:: list_assets it
```
floitsch, line 90: Not this PR, but could be nice to allow a "path" here. For example `list foo` lists files that are `foo` or start with `foo/`.
Question if we should allow absolute paths then.

floitsch (approved):
LGTM.

tools/assets.toit
kasperl: That is a valid point. I'm thinking that maybe we should build this on top based on the JSON output.

tools/assets.toit
kasperl: Done.

tools/assets.toit
kasperl: Added writers.

PR #1058, opened 2022-09-26
Handle decoding system messages with a nil program ID.

erikcorry:


tools/system_message.toit
```diff
@@ -46,13 +47,15 @@ main args:
...
   snapshot := args[0]
-  if not file.is_file snapshot:
-    pipe.print_to_stderr "No such snapshot file: $snapshot"
-    usage prog_name
-  snapshot_content := file.read_content snapshot
-  if not SnapshotBundle.is_bundle_content snapshot_content:
-    pipe.print_to_stderr "Not a snapshot file: $snapshot"
-    usage prog_name
+  snapshot_content := null
+  if not snapshot.ends_with "00000000-0000-0000-0000-000000000000.snapshot":
```
kasperl, line 51: This feels a bit hacky. You're letting the Go code get away with passing a non-existing file to tools/system_message as long as it has a name you can recognize. It would be more natural to let the snapshot argument be optional and do the filtering in the Go code. Is it obvious which messages you can decode without a snapshot?

tools/system_message.toit
```diff
@@ -37,35 +40,67 @@ usage prog_name:
...
       $prog_name <snapshot> <system message or heap_dump file>
       $prog_name <snapshot> -b <base64-encoded-ubjson>
+      $prog_name [--snapshot|-s <snapshot>] --message|-m <base64-encoded-ubjson>
       # Eg snapshot file can be toit.run.snapshot
 
     If no system-message file is given, the stack trace is read from stdin."""
   exit 1
 
 main args:
   prog_name := "system_message"
```
kasperl, line 50: Can't we just clean this up? We use tools/system_message in very few places, so it feels nicer just to get that sorted out. I'd also like to rename it to `decode_system_message` or something like that.

tools/system_message.toit
```diff
@@ -34,74 +34,35 @@ handle_system_message encoded_system_message snapshot_content:
...
+          by `jag decode` to provide nice output from the encoded
+          messages a device prints on the serial port.
+        """
+      --options=[
+        cli.OptionString "snapshot" --short_name="s"
+            --short_help="The snapshot file of the program that produced the message",
+        cli.OptionString "message" --short_name="m" --required
+            --short_help="The base64-encoded message from the device",
+      ]
+      --run=:: parsed = it
```
kasperl, line 56: Maybe just split the function here or type parsed with cli.Parsed?

tools/system_message.toit
```diff
@@ -34,74 +34,35 @@ handle_system_message encoded_system_message snapshot_content:
...
-      $prog_name <snapshot> <system message or heap_dump file>
-      $prog_name <snapshot> -b <base64-encoded-ubjson>
-      $prog_name [--snapshot|-s <snapshot>] --message|-m <base64-encoded-ubjson>
-      # Eg snapshot file can be toit.run.snapshot
-
-    If no system-message file is given, the stack trace is read from stdin."""
-  exit 1
-
 main args:
   prog_name := "system_message"
```
kasperl, line 38: program_name?

tools/system_message.toit
```diff
@@ -34,74 +34,35 @@ handle_system_message encoded_system_message snapshot_content:
...
-  if args.size < 1: usage prog_name
-  legacy_args := (args[0].ends_with ".snapshot") and (not args[0].starts_with "-")
+
+  // Use arguments library.
+  parsed := null
+  command := cli.Command prog_name
```
kasperl, line 42: The command name for the top-level command isn't used. Maybe use "root" instead like in other places?

kasperl (approved):
Looks good to me.

PR #1061, opened 2022-09-26
Revive the dump_heap primitive, more detailed malloc report.

erikcorry:
This report needs a bit more memory, but shows the fragmentation for each 8-byte memory cell.  For it to work we need to output the base64 data in a single line (recognized by Jaguar).

### DIFF
--- a/lib/esp32.toit
+++ b/lib/esp32.toit
@@ -137,7 +137,7 @@ rtc_user_bytes -> ByteArray:
   #primitive.esp32.rtc_user_bytes
 
 /**
-Produces (as a system message) a report over the usage of memory at the OS level.
+Sends (as a system message) a report over the usage of memory at the OS level.
 */
 memory_page_report -> none:
   report := memory_page_report_
@@ -145,3 +145,13 @@ memory_page_report -> none:
 
 memory_page_report_ -> ByteArray:
   #primitive.esp32.memory_page_report
+
+/**
+Sends (as a system message) a detailed report over the usage of memory at the OS level.
+*/
+dump_heap -> none:
+  report := dump_heap_ 300
+  send_trace_message report
+
+dump_heap_ slack/int -> ByteArray:
+  #primitive.core.dump_heap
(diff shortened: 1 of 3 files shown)


lib/esp32.toit
```diff
@@ -137,11 +137,21 @@ rtc_user_bytes -> ByteArray:
...
   send_trace_message report
 
 memory_page_report_ -> ByteArray:
   #primitive.esp32.memory_page_report
+
+/**
+Sends (as a system message) a detailed report over the usage of memory at the OS level.
+*/
+dump_heap -> none:
+  report := dump_heap_ 300
```
kasperl, line 153: Can we get rid of the versions that write to serial/flash from C code? They are unused?

src/heap_report.cc
```diff
@@ -227,12 +227,12 @@ class SerialFragmentationDumper : public HeapFragmentationDumper {
 };
 
 void dump_heap_fragmentation(output_char_t* output_char_fn) {
-  const char* p = "toit serial decode ";
+  const char* p = "jag decode ";
```
kasperl, line 230: Is this code used?

lib/esp32.toit
erikcorry: I want to revive them too.

system/containers.toit
```diff
@@ -350,8 +350,8 @@ trace_using_print message/ByteArray --from=0 --to=message.size:
     end := i >= to - BLOCK_SIZE
     prefix := i == from ? "jag decode " : ""
     base64_text := base64.encode message[i..(end ? to : i + BLOCK_SIZE)]
-    postfix := end ? "" : "\\"
-    print_ "$prefix$base64_text$postfix"
+    postfix := end ? "\n" : ""
+    write_on_stderr_ "$prefix$base64_text$postfix" false
```
kasperl, line 354: stdout for now

kasperl (approved):
Looks good to me.

PR #1062, opened 2022-09-27
Stop compiling system code to .s file

kasperl:


Makefile
```diff
@@ -202,6 +202,7 @@ esp32:
 esp32-no-env: check-env check-esp32-env build/$(ESP32_CHIP)/firmware.envelope
 
 build/$(ESP32_CHIP)/firmware.envelope: build/$(ESP32_CHIP)/toit.bin
+build/$(ESP32_CHIP)/firmware.envelope: build/$(ESP32_CHIP)/system.snapshot
 build/$(ESP32_CHIP)/firmware.envelope: tools toit-tools
 	$(FIRMWARE_BIN) -e build/$(ESP32_CHIP)/firmware.envelope create \
```
floitsch, line 207: I think this should eventually happen in the CMake files.
The Makefile should never be necessary, and just convenience to launch CMake.
Now both are necessary, which isn't great.

If I find the time I will try to move some of the critical things into cmake.

system/containers.toit
```diff
@@ -256,9 +256,10 @@ class ContainerManager extends ContainerServiceDefinition implements SystemMessa
       if allocation.type != FLASH_ALLOCATION_PROGRAM_TYPE: continue.do
       add_flash_image allocation
 
-    // Run through the bundled images in the VM.
+    // Run through the bundled images in the VM, but skip the
+    // first one which is always the system image.
     bundled := container_bundled_images_
-    for i := 0; i < bundled.size; i += 2:
+    for i := 2; i < bundled.size; i += 2:
```
floitsch, line 262: Add a comment explaining the `2`.
Each bundled entry seems to take two slots. Explain why.

floitsch (approved):
LGTM.

system/containers.toit
kasperl: Done.

PR #1063, opened 2022-09-27
Drop support for esp32.image_config

kasperl:


src/os_esp32.cc
```diff
@@ -740,21 +740,17 @@ void OS::heap_summary_report(int max_pages, const char* marker) { }
 
 #endif // def TOIT_CMPCTMALLOC
 
-static const int TOIT_IMAGE_DATA_SIZE = 1024;
-static const int TOIT_CONFIG_IMAGE_SIZE = TOIT_IMAGE_DATA_SIZE - UUID_SIZE - sizeof(uint32);
-
 class ImageData {
  public:
   uint32 image_pad = 0;
```
kasperl, line 745: Not sure what this does?

tools/firmware.toit
```diff
@@ -368,23 +368,23 @@ extract parsed/cli.Parsed -> none:
...
   containers ::= []
-
   entries := envelope.entries
+
+  properties_entry := entries.get AR_ENTRY_PROPERTIES
+  properties := properties_entry ? (json.decode properties_entry) : null
   system := entries.get AR_ENTRY_SYSTEM_SNAPSHOT
-  if system: containers.add (ContainerEntry "system" system --assets=null)
+  if system:
+    wifi := properties.get "wifi"
```
floitsch, line 378: Should there be a warning if there are other properties?

floitsch (approved):
LGTM.

src/os_esp32.cc
```diff
@@ -740,21 +740,17 @@ void OS::heap_summary_report(int max_pages, const char* marker) { }
...
-  uint32 image_magic1 = 0x7017da7a;  // "Toitdata"
-  // The data between image_magic1 and image_magic2 must be a multiple of 512
+  uint32 image_magic1 = 0x7017da7a;  // "toitdata"
+  // The data between image_magic1 and image_magic2 must be less than 256
   // bytes, otherwise the patching utility will not detect it. Search for
   // 0x7017da7a. Note when updating this restriction is baked into the SDK that
   // you are updating *from* so it can't be fixed without multiple SDK updates.
```
floitsch, line 750: Is the second part of the comment still true?
Aren't we patching from the SDK that is currently built?

src/os_esp32.cc
kasperl: Removed.

lib/device_impl.toit
```diff
@@ -22,7 +22,9 @@ class Device_ implements device.Device_:
   constructor.init_:
 
   name -> string?:
-    return esp32.image_config.get "name" --if_absent=: hardware_id.stringify
+    // TODO(kasper): Should this be Jaguar specific? Maybe.
+    // return esp32.image_config.get "name" --if_absent=: hardware_id.stringify
```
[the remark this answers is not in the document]
kasperl: Yeah. I think I like that.

tools/firmware.toit
kasperl: I think I will drop the support for properties as soon as we make it easier to work with assets. Let me think about this.

PR #1066, opened 2022-09-27
Fix pin.wait_for.

floitsch:
The wait_for sometimes wakes up even though the pin isn't set yet. Simply go back to wait for the state change.

### DIFF
--- a/lib/gpio/pin.toit
+++ b/lib/gpio/pin.toit
@@ -195,16 +195,15 @@ class Pin:
   */
   wait_for value -> none:
     if get == value: return
-    gpio_config_interrupt_ num true
     expected_state := value == 1 ? GPIO_STATE_UP_ : GPIO_STATE_DOWN_
+    state_.clear_state expected_state
+    gpio_config_interrupt_ num true
     try:
       // Make sure the pin didn't change to the expected value while we
       // were setting up the interrupt.
-      // The loop is to prevent accidental wakeups.
-      while get != value:
-        state_.wait_for_state expected_state
-        state_.clear_state expected_state
-      return
+      if get == value: return
+
+      state_.wait_for_state expected_state
     finally:
       gpio_config_interrupt_ num false
       state_.clear_state expected_state
--- a/lib/gpio/pin.toit
+++ b/lib/gpio/pin.toit
@@ -189,16 +189,18 @@ class Pin:
   wait_for value -> none:
     if get == value: return
     gpio_config_interrupt_ num true
+    expected_state := value == 1 ? GPIO_STATE_UP_ : GPIO_STATE_DOWN_
     try:
       // Make sure the pin didn't change to the expected value while we
       // were setting up the interrupt.
-      if get == value: return
-      expected_state := value == 1 ? GPIO_STATE_UP_ : GPIO_STATE_DOWN_
-      state := state_.wait_for_state expected_state
-      state_.clear_state expected_state
+      // The loop is to prevent accidental wakeups.
+      while get != value:
+        state_.wait_for_state expected_state
+        state_.clear_state expected_state
       return
     finally:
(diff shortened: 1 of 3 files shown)


tests/esp32/wait_for1.toit
```diff
@@ -26,8 +26,8 @@ ITERATIONS ::= 10_000
 main:
   pin_in := gpio.Pin PIN_IN --input --pull_down
   pin_out := gpio.Pin PIN_OUT --output
-  sleep --ms=100
 
+  pin_in.configure --input --pull_down
```
floitsch, line 30: This is due to another bug that I will (hopefully) fix next.

kasperl: Will you drop this again?

tests/esp32/wait_for2.toit
```diff
@@ -17,10 +17,12 @@ main:
   pin_in := gpio.Pin PIN_IN --input
   pin_out := gpio.Pin PIN_OUT --output
 
-  ITERATIONS.repeat:
+  ITERATIONS.repeat: | iteration |
+    if iteration % 1000 == 0: print "Iteration: $iteration"
```
kasperl, line 21: Is the printing still useful?

tests/esp32/wait_for2.toit
```diff
@@ -17,10 +17,12 @@ main:
...
     before := pin_in.get
     with_timeout --ms=2_000:
       pin_in.wait_for 1
+    if pin_in.get != 1: throw "Expected pin to be 1 - $iteration $before"
```
kasperl, line 25: The test here generally does not expect or throw. This is different. Should this use 'expect'?

lib/gpio/pin.toit
```diff
@@ -189,16 +189,18 @@ class Pin:
...
     gpio_config_interrupt_ num true
+    expected_state := value == 1 ? GPIO_STATE_UP_ : GPIO_STATE_DOWN_
     try:
       // Make sure the pin didn't change to the expected value while we
       // were setting up the interrupt.
-      if get == value: return
-      expected_state := value == 1 ? GPIO_STATE_UP_ : GPIO_STATE_DOWN_
-      state := state_.wait_for_state expected_state
-      state_.clear_state expected_state
+      // The loop is to prevent accidental wakeups.
```
kasperl, line 196: So how does the wakeup happen? I feel that I need to understand this better. With this change, it feels like you might drop signals (say the pin goes high, so we wakeup, but then it is low again before we ask for the value).

floitsch: Rewritten.

tests/esp32/wait_for1.toit
floitsch: done.

tests/esp32/wait_for2.toit
floitsch: I find it useful, as it shows me if that tests are doing something and not just hitting the timeout.

tests/esp32/wait_for2.toit
floitsch: Switched to `expect`.

lib/gpio/pin.toit
```diff
@@ -195,17 +195,18 @@ class Pin:
...
       if get == value: return
-      expected_state := value == 1 ? GPIO_STATE_UP_ : GPIO_STATE_DOWN_
-      state := state_.wait_for_state expected_state
-      state_.clear_state expected_state
-      return
+
+      state_.wait_for_state expected_state
     finally:
       gpio_config_interrupt_ num false
+      state_.clear_state expected_state
```
floitsch, line 209: For some reason this line is necessary.
I had hoped that it would be enough to clear it in line 199, but without this line the test fails.

kasperl (approved):
Looks good to me.

PR #1067, opened 2022-09-27
Fix the gpio.Pin constructor.

floitsch:
It was forgetting to forward the pull_up/pull_down arguments. Also close the pin if the configure fails.


kasperl (approved):
Looks good to me.

PR #1070, opened 2022-09-28
Fix they key/colors for the heap dump

erikcorry:


kasperl (approved):
Looks good to me. Weird, but good.

PR #1073, opened 2022-09-30
Make hex.decode and hex.encode tree-shakable.

erikcorry:
Currently these functions use two primitives, but they are rarely used, and can be rewritten with the blit primitive, saving two primitives, and saving space in the case where they are not used.


lib/encoding/hex.toit
```diff
@@ -2,8 +2,60 @@
...
+  return result
 
 decode str/string -> ByteArray:
-  #primitive.encoding.hex_decode
+  if str.size == 0: return #[]
+  if str.size == 2: return #[int.parse --radix=16 str]
+  if str.size & 1 != 0: throw "INVALID_ARGUMENT"
+  checker := #[0]
+  blit str checker str.size
+    --destination_pixel_stride=0
```
kasperl, line 46: 4 space indent for the -- lines here.

kasperl (approved):
Looks good to me.

PR #1074, opened 2022-09-30
ByteArray constructor takes filler argument.

erikcorry:
Previously you could only give an initialization block, which is much less efficient if all elements need the same filler.

This brings ByteArray into line with List, which has always had this option.


kasperl (approved):
Looks good to me.

lib/core/collections.toit
```diff
@@ -1022,9 +1022,9 @@ interface ByteArray:
   /**
   Creates a new byte array of the given $size.
 
-  All elements are initialized to 0.
+  All elements are initialized to the $filler, which default to 0.
```
floitsch, line 1025:
```suggestion
  All elements are initialized to the $filler, which defaults to 0.
```

lib/core/collections.toit
```diff
@@ -1466,7 +1466,7 @@ class ByteArray_ extends ByteArrayBase_:
 
   All elements are initialized to 0.
```
floitsch, line 1467: Update comment.

lib/core/collections.toit
```diff
@@ -1022,9 +1022,9 @@ interface ByteArray:
   /**
   Creates a new byte array of the given $size.
 
-  All elements are initialized to 0.
+  All elements are initialized to the $filler, which default to 0.
   */
-  constructor size/int:
+  constructor size/int filler/int=0:
```
floitsch, line 1027: I would prefer if the filler is a named argument.
At the moment named arguments are passed after positional ones (but in sorted order). Not sure if we want to rely on this for primitive functions, or if it's better to forward to a static function instead.

Also not sure, if `--fill_value` or `--initial_value` wouldn't be better.

floitsch (approved):
LGTM.

PR #1076, opened 2022-09-30
Unentangle JSON string escaping from general JSON encoding.

erikcorry:
This way we can escape individual strings without
pulling in the whole JSON library.

### DIFF
--- a/lib/encoding/json.toit
+++ b/lib/encoding/json.toit
@@ -170,6 +170,7 @@ This means that control characters, double quotes and backslashes have
   been replaced by backslash sequences.
 */
 escape_string str -> any:
+  if str == "" or str.size == 1 and ESCAPED_CHAR_MAP_[str[0]] == 0: return str
   counter := #[0, 0]
   bitmap.blit str counter str.size
       --destination_pixel_stride=0
--- a/lib/encoding/json.toit
+++ b/lib/encoding/json.toit
@@ -169,9 +169,9 @@ Returns a string or a byte array that has been escaped for use in JSON.
 This means that control characters, double quotes and backslashes have
   been replaced by backslash sequences.
 */
-escape_string str -> any:
+escape_string str/string -> any:
   if str == "" or str.size == 1 and ESCAPED_CHAR_MAP_[str[0]] == 0: return str
-  counter := #[0, 0]
+  counter := ByteArray 2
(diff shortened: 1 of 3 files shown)


lib/encoding/json.toit
```diff
@@ -141,6 +143,70 @@ class Buffer_:
...
+  result := ByteArray 0x100: it < ' ' ? 5 : 0
+  ONE_CHAR_ESCAPES_.do: | code escape | result[code] = 1
+  return result
+
+/**
+Returns a string or a byte array that has been escaped for use in JSON.
+This means that control characters, double quotes and backslashes have
+  been replaced by backslash sequences.
+*/
+escape_string str -> any:
```
kasperl, line 172:
```suggestion
escape_string str/string -> any:
```

lib/encoding/json.toit
```diff
@@ -248,16 +282,15 @@ class Encoder extends Buffer_:
...
     put_byte_
-      _hex_digit (code_point >> 4) & 0xf
+      hex_digit_ (code_point >> 4) & 0xf
     put_byte_
-      _hex_digit code_point & 0xf
-
+      hex_digit_ code_point & 0xf
 
-_hex_digit x: return x < 10 ? '0' + x : 'a' + x - 10
+hex_digit_ x: return x < 10 ? '0' + x : 'a' + x - 10
```
kasperl, line 293: Do we have something like this elsewhere? encoding.hex?

lib/encoding/json.toit
```diff
@@ -141,6 +143,70 @@ class Buffer_:
...
+      --lookup_table=ESCAPED_CHAR_MAP_
+      --operation=bitmap.ADD_16_LE
+  extra_chars := LITTLE_ENDIAN.uint16 counter 0
+  if extra_chars == 0: return str  // Nothing to escape.
+  if extra_chars == 0xffff:
+    // Overflow of the saturating counter :-(.  We must count manually.
+    extra_chars = 0
+    str.size.repeat: extra_chars += ESCAPED_CHAR_MAP_[str.at --raw it]
+  result := ByteArray str.size + extra_chars
+  // Index of the output in the byte array:
```
kasperl, line 186: Maybe replace these two comments with better local variable names?

kasperl (approved):
Looks good to me.

PR #1077, opened 2022-09-30
Deprecate --foo.

floitsch:
They look too much like named parameters.


src/compiler/parser.cc
```diff
@@ -1887,9 +1887,14 @@ Expression* Parser::parse_unary(bool allow_colon) {
...
       consume();
       if (!is_current_token_attached()) {
-        report_error("Can't have space between '%s' and the operand",
+        report_error(range.extend(current_range()),
+                     "Can't have space between '%s' and the operand",
                      Token::symbol(kind).c_str());
       }
+      if (kind == Token::DECREMENT) {
+        diagnostics()->report_warning(range.extend(current_range()),
+                                      "Prefixed decrement is deprecated.");
```
kasperl, line 1896:
```suggestion
                                      "Prefix decrement is deprecated.");
```

kasperl (approved):
Looks good to me.

src/compiler/parser.cc
kasperl: Fixed.

PR #1080, opened 2022-10-01
Harden TISON decoding

kasperl:


src/messaging.cc
```diff
@@ -96,7 +106,26 @@ void MessageEncoder::neuter_externals() {
...
+    write_cardinal(payload_size());
+  }
+  bool result = encode_any(object);
+  if (!result) return false;
+  // Compute the number of bytes we need to encode the payload size.
+  // Later, when we're not encoding for size, we know the payload size
+  // and will encode this before the payload.
+  if (encoding_for_size()) {
+    _payload_size = size() - sizeof(uint32);
+   write_cardinal(payload_size());
```
erikcorry, line 123: Strange indentation.

Also, this is really subtle.  We initialize the payload size to -1, which happens to have a very long cardinal encoding, which means that when we actually write, the size will be slightly smaller.  Needs comments, including at the place where the payload size is initialized to -1.

Or perhaps a version of write_cardinal that always writes in a fixed length, regardless of the number.

And a test that needs a two-byte size.

erikcorry (approved):

src/messaging.cc
kasperl: We don't write the payload size of -1. I'll add an assertion. The comment tries to explain that we compute the payload size when encoding for size and write it when encoding afterwards.

Fixed indentation.

kasperl: Oh, and there are lots of tests that use multiple bytes for encoding the payload size.

PR #1086, opened 2022-10-02
Add support for cross-compilation to the compiler.

floitsch:
Once we have the correct vessels we can cross-compile.


kasperl (approved):
Looks good to me.

PR #1088, opened 2022-10-03
Include class name in serialization error

erikcorry:


tests/negative/serialization_test.toit
```diff
@@ -0,0 +1,10 @@
+// Copyright (C) 2019 Toitware ApS.
```
kasperl, line 1:
```suggestion
// Copyright (C) 2022 Toitware ApS.
```

src/primitive_encoding.cc
```diff
@@ -144,7 +144,13 @@ PRIMITIVE(tison_encode) {
   unsigned size = 0;
   unsigned payload_size = 0;
   { TisonEncoder size_encoder(process);
-    if (!size_encoder.encode(object)) WRONG_TYPE;
+    if (!size_encoder.encode(object)) {
```
kasperl, line 147: I think we should consider doing the same for `process_send`.

tools/mirror.toit
```diff
@@ -199,6 +199,13 @@ class Error extends Mirror:
       class_name = method_info.as_class_name relative_bci
     return "As check failed: $(typed_expression_string_ expression) is not a $class_name.\n$trace"
 
+  serialization_failed_stringify -> string:
+    // message is an array [expression, id]
```
kasperl, line 203: wat?

src/messaging.h
```diff
@@ -151,6 +151,8 @@ class MessageEncoder {
   bool encode_bundles(SnapshotBundle system, SnapshotBundle application);
 #endif
 
+  int problematic_class_id() const { return problematic_class_id_; }
```
kasperl, line 154: Maybe it is time we let encode return a structured object (by value) instead? We have a number of different things we'd like to communicate.

kasperl (approved):
Oh, and looks good to me.

PR #1089, opened 2022-10-03
Extend new serialization errors to process_send

erikcorry:


src/primitive_core.cc
```diff
@@ -1804,7 +1805,20 @@ PRIMITIVE(process_send) {
 
   unsigned size = 0;
   { MessageEncoder size_encoder(process, null);
-    if (!size_encoder.encode(array)) WRONG_TYPE;
+    if (!size_encoder.encode(array)) {
+      Object* result = null;
+      int class_id = size_encoder.problematic_class_id();
```
kasperl, line 1810: Can this code be shared? Can the encoder help give the right error if asked politely?

src/primitive_encoding.cc
```diff
@@ -146,6 +146,8 @@ PRIMITIVE(tison_encode) {
   { TisonEncoder size_encoder(process);
     if (!size_encoder.encode(object)) {
       int id = size_encoder.problematic_class_id();
+      if (size_encoder.nesting_too_deep())
```
kasperl, line 149: I'd use { } for the branch here too.

src/primitive_core.cc
```diff
@@ -1841,14 +1848,37 @@ PRIMITIVE(process_send) {
...
     // TODO(kasper): Consider doing in-place shrinking of internal, non-constant
     // byte arrays and strings.
+    return process->program()->null_object();
   } else {
     // Sending failed. Free any copied bits, but make sure to not free the externals
     // that have not been neutered on this path.
     encoder.free_copied();
     message->free_data_but_keep_externals();
     delete message;
+    // TODO: We should activate this immediately after the release to see how
```
kasperl, line 1858: Isn't this already activated?

src/primitive_core.cc
```diff
@@ -1841,14 +1848,37 @@ PRIMITIVE(process_send) {
     encoder.neuter_externals();
     // TODO(kasper): Consider doing in-place shrinking of internal, non-constant
     // byte arrays and strings.
+    return process->program()->null_object();
   } else {
```
kasperl, line 1852: Maybe drop the else part and get rid of one level of nesting?

src/primitive_encoding.cc
```diff
@@ -159,7 +155,11 @@ PRIMITIVE(tison_encode) {
   if (!result) ALLOCATION_FAILED;
   ByteArray::Bytes bytes(result);
   TisonEncoder encoder(process, bytes.address(), payload_size);
-  if (!encoder.encode(object)) OTHER_ERROR;
+  if (!encoder.encode(object)) {
+    if (encoder.malloc_failed()) MALLOC_FAILED;
```
kasperl, line 159: Can't you use `create_error_object` here too?

src/primitive_core.cc
```diff
@@ -1814,18 +1819,20 @@ PRIMITIVE(process_send) {
...
   SystemMessage* message = null;
   MessageEncoder encoder(process, buffer);
-  bool encoding_succeeded = false;
-  if (encoder.encode(array)) {
-    encoding_succeeded = true;
-    message = _new SystemMessage(type, process->group()->id(), process->id(), buffer);
+  if (!encoder.encode(array)) {
+    encoder.free_copied();
+    free(buffer);
+    if (encoder.malloc_failed()) MALLOC_FAILED;
```
kasperl, line 1825: Use `create_error_object` here too?

src/primitive_core.cc
```diff
@@ -1814,18 +1819,20 @@ PRIMITIVE(process_send) {
 
   SystemMessage* message = null;
   MessageEncoder encoder(process, buffer);
-  bool encoding_succeeded = false;
-  if (encoder.encode(array)) {
-    encoding_succeeded = true;
-    message = _new SystemMessage(type, process->group()->id(), process->id(), buffer);
+  if (!encoder.encode(array)) {
+    encoder.free_copied();
```
kasperl, line 1823: Maybe we can flip it around so you have to call something on the encoder to get it to not free the copies and the incoming buffer? That way we might be able to get away with just returning here.

It is essentially the `neuter_externals` call, I think.

src/primitive_core.cc
```diff
@@ -140,11 +142,12 @@ PRIMITIVE(spawn) {
   if (!encoder.encode(arguments)) {
     encoder.free_copied();
     free(buffer);
-    if (encoder.malloc_failed()) MALLOC_FAILED;
+    // Should not happen, but just in case, throw "ERROR".
```
kasperl, line 145: Use `create_error_object` here too?

src/primitive_core.cc
```diff
@@ -1814,18 +1819,20 @@ PRIMITIVE(process_send) {
...
-    message = _new SystemMessage(type, process->group()->id(), process->id(), buffer);
+  if (!encoder.encode(array)) {
+    encoder.free_copied();
+    free(buffer);
+    if (encoder.malloc_failed()) MALLOC_FAILED;
+    // Should not happen, but just in case, throw "ERROR".
+    OTHER_ERROR;
   }
 
+  message = _new SystemMessage(type, process->group()->id(), process->id(), buffer);
```
kasperl, line 1830: Introduce the `message` variable here.

kasperl (approved):
Looks good to me, but I think we should try to make the encoder responsible for freeing the buffer and the copies on failures and rename `neuter_externals` to something similar to `AllocationManager.keep_result`.

PR #1091, opened 2022-10-03
Fix some compiler warnings.

floitsch:


src/os_esp32.cc
```diff
@@ -312,7 +315,7 @@ bool Thread::spawn(int stack_size, int core) {
   if (core == -1) core = tskNO_AFFINITY;
 
   BaseType_t res = xTaskCreatePinnedToCore(
-    reinterpret_cast<TaskFunction_t>(thread_start),
+    esp_thread_start,
```
floitsch, line 318: Warning was complaining that this is an incompatible cast.

src/primitive_font.cc
```diff
@@ -43,8 +43,8 @@ bool FontBlock::verify(const uint8* data, uint32 length, const char* name) {
   if (Utils::read_unaligned_uint32(data + 4) != length) return false;
   if (data[length - 1] != 0xff) return false;
   uint32 offset = 40;
-  uint32 from = -1;
-  uint32 to = -1;
+  int32 from = -1;
```
floitsch, line 46: There is a test in line 84 against `from < 0 || to < 0`.

erikcorry (approved):

PR #1093, opened 2022-10-04
Test with qemu.

floitsch:


toolchains/raspberry_pi64.cmake
```diff
@@ -0,0 +1,24 @@
+# Copyright (C) 2022 Toitware ApS.
```
floitsch, line 1: Currently unused, but I didn't want to lose the flags.

kasperl (approved):
Looks good to me.

PR #1097, opened 2022-10-05
Reserve 'catch', 'throw', 'rethrow' and 'switch'.

floitsch:
We probably want to make these reserved identifiers.

### DIFF
--- a/src/compiler/resolver.cc
+++ b/src/compiler/resolver.cc
@@ -97,7 +97,6 @@ ir::Program* Resolver::resolve(const std::vector<ast::Unit*>& units,
   mark_runtime(modules[core_index]);
   mark_non_returning(modules[core_index]);
 
-
   setup_inheritance(modules, core_index);
 
   fill_classes_with_skeletons(modules);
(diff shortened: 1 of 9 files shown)


src/compiler/resolver.cc
```diff
@@ -94,15 +94,18 @@ ir::Program* Resolver::resolve(const std::vector<ast::Unit*>& units,
   auto modules = build_modules(units, entry_index, core_index);
   build_module_scopes(modules);
 
-  mark_runtime_classes(modules[core_index]);
+  mark_runtime(modules[core_index]);
   mark_non_returning(modules[core_index]);
 
+
```
kasperl, line 100: Drop this newline again?

src/compiler/resolver.cc
```diff
@@ -1594,7 +1642,8 @@ void Resolver::fill_classes_with_skeletons(std::vector<Module*> modules) {
           auto method = member->as_Method();
           auto method_is_abstract = method->is_abstract();
 
-          check_method(method, ir_class, &member_name, &kind);
+          bool allow_future_reserved;
+          check_method(method, ir_class, &member_name, &kind, allow_future_reserved = false);
```
kasperl, line 1646: I prefer assigning to the local in the previous line (like you do in most other places).

kasperl (approved):
Looks good to me.

PR #1103, opened 2022-10-07
Assets get

kasperl:


tools/assets.toit
```diff
@@ -95,6 +96,52 @@ add_asset parsed/cli.Parsed -> none:
...
+  entries := load_assets input_path
+  entry := entries.get name
+  if not entry:
+    print "No such asset: $name"
+    exit 1
+  content := entry
+  exception := null
+  if format == "auto":
+    decoded := null
+    catch: decoded = decoded or "$(json.stringify (tison.decode content))\n"
```
floitsch, line 132: Would be nice to do this without exceptions.
For tison and maybe ubjson there might be a header. That said, the functionality should be next to the en/decoder and not here...

floitsch (approved):
LGTM.

tools/assets.toit
kasperl: Yeah, I thought about that too. It is super fast to fail, but it sucks to have to use exceptions. Maybe we should have a block -- or maybe we should have a "looks like this might be ubjson/base64/tison/json" methods?

PR #1105, opened 2022-10-08
Add support for process priorities

kasperl:


kasperl:
Maybe something like this, @mikkeldamsgaard?

src/process.h
```diff
@@ -144,7 +151,21 @@ class Process : public ProcessListFromProcessGroup::Element,
 
   void signal(Signal signal);
   void clear_signal(Signal signal);
-  uint32_t signals() const { return _signals; }
+  uint32 signals() const { return _signals; }
+
+  // Processes have a priority in the range [0..255]. The scheduler
+  // priorities running processes with higher priorities, so processes
```
erikcorry, line 157:
```suggestion
  // prioritizes running processes with higher priorities, so processes
```

erikcorry (approved):
As I understand it we don't offer a guarantee that a process with a higher priority will be run if it is ready to run.

The Toit priorities have no effect on the priorities of the FreeRTOS tasks that are used to run them.   Also no effect on native tasks?

Priority inversion protection is not implemented.  Eg priority inheritance for mutex monitors.

We might want to clarify that in the doc comment.

PR #1107, opened 2022-10-09
Avoid throwing in process_send for now

kasperl:


kasperl:
TBR'ed.

Without disabling this, we get weird crashes when we try to send messages during task shutdown -- outside the last-resort exception handler.

PR #1109, opened 2022-10-11
Fix build on macOS

kasperl:


floitsch (approved):
LGTM.
please add a comment why this is necessary

kasperl:
What kind of comment would make sense to you? It is just using sed the wrong way because -i always take an extension on the Mac.

floitsch:
I just looked it up and understand now.
I think it's fine this way.

PR #1113, opened 2022-10-11
Disallow use of some GPIO pins.

floitsch:
When using one of the flash/PSRAM pins the system crashes.

Resources for the documentation:
ESP32: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/gpio.html
ESP32C3: https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-reference/peripherals/gpio.html
ESP32S3: https://docs.espressif.com/projects/esp-idf/en/latest/esp32s3/api-reference/peripherals/gpio.html

### DIFF
--- a/lib/gpio/pin.toit
+++ b/lib/gpio/pin.toit
@@ -47,16 +47,23 @@ class Pin:
 
   While the Pin is in input mode, $pull_up and $pull_down resistors are applied as
     configured.
+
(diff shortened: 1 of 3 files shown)


kasperl (approved):
Looks good to me.

PR #1116, opened 2022-10-12
Compile the esptool and deliver it with our tools.

floitsch:


kasperl (approved):
Looks good to me. I assume this will automatically be part of SDK releases because we bundle up the build/.../sdk directory?

floitsch:
Correct. As soon as it's in the tools directory it should be distributed with the SDK.

PR #1117, opened 2022-10-12
Generate and upload C3 and S3 envelopes.

floitsch:


.github/workflows/ci.yml
```diff
@@ -377,6 +383,24 @@ jobs:
           tag: ${{ github.ref }}
           overwrite: true
 
+      - name: Upload ESP32C3 release firmware artifacts
```
kasperl, line 386: Maybe ESP32-C3?

.github/workflows/ci.yml
```diff
@@ -377,6 +383,24 @@ jobs:
...
+        if: (github.event_name) == 'release'
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: build/firmware-esp32c3.gz
+          tag: ${{ github.ref }}
+          overwrite: true
+
+      - name: Upload ESP32S3 release firmware artifacts
```
kasperl, line 395: Maybe ESP32-S3?

.github/workflows/ci.yml
```diff
@@ -356,19 +356,25 @@ jobs:
           source third_party/esp-idf/export.sh
           make esp32
           make ESP32_CHIP=esp32s3 esp32
+          make ESP32_CHIP=esp32c3 esp32
 
       - name: Pack firmware artifacts
         shell: bash
         run: |
           gzip --to-stdout build/esp32/firmware.envelope > build/firmware-esp32.gz
+          gzip --to-stdout build/esp32c3/firmware.envelope > build/firmware-esp32c3.gz
```
kasperl, line 365: Do we have a naming convention in mind for the variants? I think the Espressif folks usually call these ESP32-S3 to make it easier to grok, but if that conflicts with a variants, we may want to keep it as is here.

kasperl (approved):
Looks good to me.

.github/workflows/ci.yml

.github/workflows/ci.yml
floitsch: I temporarily changed it to `firmware-esp32-c3.gz` and didn't like it. Makes it feel like, the `c3` is not part of the chip name.
For now leaving as is, but willing to change.

PR #1120, opened 2022-10-12
Ble rewrite

mikkeldamsgaard:
Ble library rewrite and added support for ble on macos via toit.
Major enhancements:
  1. Discovery
  2. Long attribute writes
  3. Multiple clients for server
  4. Support on macos with the same ble.toit library
  5 more details from scans
  6 descriptor support (esp32)
  7 Subscription support


lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
-    resource_state_ = monitor.ResourceState_ device.resource_group_ gatt_
-    ble_connect_ device.resource_group_ address.raw_ gatt_
+  constructor .manager .address:
+    device := ble_connect_ manager.resource_ address
+    super manager.adapter.resource_group_ device --auto_release
     state := resource_state_.wait_for_state CONNECTED_EVENT_ | CONNECT_FAILED_EVENT_
     if state & CONNECT_FAILED_EVENT_ != 0:
+      close_
       throw "BLE connection failed"
       // TODO Possible leak of the gatt_ resource on connection failures
```
floitsch, line 308:
```suggestion
      // TODO: Possible leak of the gatt_ resource on connection failures.
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
 
   /**
-  Reads a remote service by looking up the given $service_uuid on the remote device.
-  */
-  read_service service_uuid/uuid_pkg.Uuid -> RemoteService:
-    ble_request_service_ gatt_ service_uuid.to_byte_array
-    wait_for_done_
-    result := ble_request_result_ gatt_
-    return RemoteService this service_uuid result
+  Discovers remote services by looking up the given $service_uuids on the remote device
```
floitsch, line 311:
```suggestion
  Discovers remote services by looking up the given $service_uuids on the remote device.
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
 
-  wait_for_done_:
-    state := resource_state_.wait_for_state STARTED_EVENT_
-    resource_state_.clear_state STARTED_EVENT_
+  If $service_uuids is empty all services for the device is discovered.
```
floitsch, line 313:
```suggestion
  If $service_uuids is empty all services for the device are discovered.
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
 
-/**
-Contains definitions of the services exposed by the BLE server.
-*/
-class ServerConfiguration:
-  resource_group_ := null
-  services/List := []
+  Note: Some platforms only support an empty list or a list of size 1. If the platform is limited, this method
+  will throw INVALID_ARGUMENT.
```
floitsch, line 316:
```suggestion
    throws "INVALID_ARGUMENT".
```
or maybe just
```suggestion
    throws.
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
+    state := wait_for_state_with_gc_ SERVICES_DISCOVERED_EVENT_
+                                   | DISCONNECTED_EVENT_
+                                   | DISCOVERY_OPERATION_FAILED_
+    if state & DISCONNECTED_EVENT_ != 0:
+      throw "BLE disconnected"
+    else if state & DISCOVERY_OPERATION_FAILED_ != 0:
+      throw_error_
+
+    discovered_services.add_all
+        List.from
```
floitsch, line 331: Is the `List.from` needed?
The `map` function should always give a normal `List`.

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
+
+    return order_attributes_ service_uuids discovered_services
 
-  constructor:
-    resource_group_ = ble_server_configuration_init_
-    add_finalizer this:: ble_server_configuration_dispose_ resource_group_
+  /**
+  Disconnects from the remote device.
+  */
+  disconnect:
```
floitsch, line 340: Since we don't have a `connect` function, I would go with `close` here.

If we ever get some "auto-close" functionality, it would work here too. (Not that anything is planned).

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
+    ble_disconnect_ resource_
+    resource_state_.wait_for_state DISCONNECTED_EVENT_
+    close_
 
-  add_service uuid -> Service:
-    service := Service this uuid
-    services.add service
-    return service
+  get_mtu -> int:
```
floitsch, line 345:
```suggestion
  mtu -> int:
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
+  constructor .peripheral_manager .uuid:
+    resource := ble_add_service_ peripheral_manager.resource_ uuid.encode_for_platform_
+    super peripheral_manager.adapter.resource_group_ resource --auto_release
 
-  characteristics_/Map ::= {:}
+  /**
+  Adds a characteristic to this service with the given paraemters.
+  $uuid is the uuid of the characteristic
+  $properties is one of the CHARACTERISTIC_PROPERTY_* values
+  $permissions is one of the CHARACTERISTIC_PERMISSIONS_* values
```
floitsch, line 366:
```suggestion
  $uuid is the uuid of the characteristic.
  $properties is one of the CHARACTERISTIC_PROPERTY_* values (see
    $CHARACTERISTIC_PROPERTY_BROADCAST and similar).
  $permissions is one of the CHARACTERISTIC_PERMISSION_* values (see $CHARACTERISTIC_PERMISSION_READ and similar).
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
+  if $value is specified, it will be used as an initial value for the characteristic.
```
floitsch, line 367:
```suggestion
  if $value is specified, it is used as the initial value for the characteristic.
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
 
-  constructor .server_configuration_ .uuid:
-    resource_ = ble_add_server_service_ server_configuration_.resource_group_ uuid.to_byte_array
+  NOTE: Only has effect if the corresponding service is not yet deployed
```
floitsch, line 369: Can we throw if it is already deployed?
```suggestion
  NOTE: Only has effect if the corresponding service is not yet deployed.
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
+    return LocalCharacteristic this uuid properties permissions value
 
-  add_read_only_characteristic uuid --value=#[]-> ReadOnlyCharacteristic:
-    char := ReadOnlyCharacteristic this uuid value
-    characteristics_[uuid] = char
-    return char
+  /**
+  Adds a read only characteristic with the given $uuid and $value. See $add_characteristic.
+  */
+  add_read_only_characteristic uuid/BleUUID --value/ByteArray -> LocalCharacteristic:
```
floitsch, line 381: ?
```suggestion
  /**
  Variant of $(add_characteristic uuid --properties --permissions).

  Adds a read-only characteristic with the given $uuid and $value.
  */
  add_characteristic --read_only/bool uuid/BleUUID --value/ByteArray -> LocalCharacteristic:
    if not read_only: throw "INVALID_ARGUMENT"
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
 
-  add_write_only_characteristic uuid --require_response/bool=true -> WriteOnlyCharacteristic:
-    char := WriteOnlyCharacteristic this uuid require_response
-    characteristics_[uuid] = char
-    return char
+  /**
+  Adds a write only characteristic with the given $uuid that can $require_responsew for each write.
+  See $add_characteristic.
+  */
+  add_write_only_characteristic uuid/BleUUID requires_response/bool=false -> LocalCharacteristic:
```
floitsch, line 392: ?
```suggestion
  /**
  Variant of $(add_characteristic uuid --properties --permissions).

  Adds a write-only characteristic with the given $uuid that can $require_response for each write.
  */
  add_characteristic --write_only/bool uuid/BleUUID requires_response/bool=false -> LocalCharacteristic:
    if not write_only: throw "INVALID_ARGUMENT"
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
+        --properties=requires_response?CHARACTERISTIC_PROPERTY_WRITE:CHARACTERISTIC_PROPERTY_WRITE_WITHOUT_RESPONSE
+        --permissions=CHARACTERISTIC_PERMISSION_WRITE
 
-  add_read_write_characteristic uuid --value=#[]-> ReadWriteCharacteristic:
-    char := ReadWriteCharacteristic this uuid value
-    characteristics_[uuid] = char
-    return char
+  /**
+  Adds a notification characteristic with the given $uuid that can be an indication if $indication is true,
+  otherwise it will be a notification. See $add_characteristic.
```
floitsch, line 400:
```suggestion
    otherwise it will be a notification. See $add_characteristic.
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
+  */
+  add_notification_characteristic uuid/BleUUID indication/bool=false -> LocalCharacteristic:
```
floitsch, line 402: If chosen to use the `--read_only` and `write_only` versions from above, might as well make two more:
`--for_notification` and `--for_indication`.  (Not 100% happy with the naming):
```
add_characteristic --for_notification ...
add_characteristic --for_indication ...
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
+        uuid
+        --properties=indication?CHARACTERISTIC_PROPERTY_INDICATE:CHARACTERISTIC_PROPERTY_NOTIFY
+        --permissions=CHARACTERISTIC_PERMISSION_READ
 
-  add_notification_characteristic uuid -> NotificationCharacteristic:
-    char := NotificationCharacteristic this uuid
-    characteristics_[uuid] = char
-    return char
+  /**
+  Deploys this service. After deployment, no more characteristics can be added. See $add_characteristic.
```
floitsch, line 409:
```suggestion
  Deploys this service.
  After deployment, no more characteristics can be added.
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
+    super service resource
 
-/**
-Base class of all characteristics.
-*/
-abstract class Characteristic:
   /**
-  The UUID of the characteristic.
+  Send a notification or an indication, based on the properties of the characteristic. If the characteristic
+  supports both indications and notifications, then a notification is send.
```
floitsch, line 430:
```suggestion
  Sends a notification or an indication, based on the properties of the characteristic.
  If the characteristic supports both indications and notifications, then a notification is sent.
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
-/**
-A characteristic that can only be read by clients.
-*/
-class ReadOnlyCharacteristic extends Characteristic:
-  value_/ByteArray := #[]
+  /**
+  Adds a descriptor to this characteristic.
+  $uuid is the uuid of the characteristic
+  $properties is one of the CHARACTERISTIC_PROPERTY_* values
+  $permissions is one of the CHARACTERISTIC_PERMISSIONS_* values
```
floitsch, line 454:
```suggestion
  $uuid is the uuid of the characteristic.
  $properties is one of the CHARACTERISTIC_PROPERTY_* values (see
    $CHARACTERISTIC_PROPERTY_BROADCAST and similar).
  $permissions is one of the CHARACTERISTIC_PERMISSION_* values (see $CHARACTERISTIC_PERMISSION_READ and similar).
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
-        value
-    super service resource uuid
-
-  value= value/ByteArray -> none:
-    ble_set_characteristics_value_ state_.resource value
+  constructor .characteristic .uuid .properties .permissions value:
+    resource :=  ble_add_descriptor_ characteristic.resource_ uuid properties permissions value
+    super characteristic.service resource
 
+  value= value/ByteArray:
```
floitsch, line 473:
```suggestion
  write value/ByteArray:
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
-    characteristics_[uuid] = char
-    return char
+  /**
+  Adds a notification characteristic with the given $uuid that can be an indication if $indication is true,
+  otherwise it will be a notification. See $add_characteristic.
+  */
+  add_notification_characteristic uuid/BleUUID indication/bool=false -> LocalCharacteristic:
+    return add_characteristic
+        uuid
+        --properties=indication?CHARACTERISTIC_PROPERTY_INDICATE:CHARACTERISTIC_PROPERTY_NOTIFY
```
floitsch, line 405: Too crowded. Move the properties into a variable first and then use it here.

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
-  add_notification_characteristic uuid -> NotificationCharacteristic:
-    char := NotificationCharacteristic this uuid
-    characteristics_[uuid] = char
-    return char
+  /**
+  Deploys this service. After deployment, no more characteristics can be added. See $add_characteristic.
+  */
+  deploy:
+    ble_deploy_service_ resource_
+    state := resource_state_.wait_for_state SERVICE_ADD_SUCCEEDED_EVENT_ | SERVICE_ADD_FAILED_EVENT_
```
floitsch, line 413:
```suggestion
    state := resource_state_.wait_for_state (SERVICE_ADD_SUCCEEDED_EVENT_ | SERVICE_ADD_FAILED_EVENT_)
```

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
-    characteristics_[uuid] = char
-    return char
+  /**
+  Adds a write only characteristic with the given $uuid that can $require_responsew for each write.
+  See $add_characteristic.
+  */
+  add_write_only_characteristic uuid/BleUUID requires_response/bool=false -> LocalCharacteristic:
+    return add_characteristic
+        uuid
+        --properties=requires_response?CHARACTERISTIC_PROPERTY_WRITE:CHARACTERISTIC_PROPERTY_WRITE_WITHOUT_RESPONSE
```
floitsch, line 395: Too crowded. Move the properties into a variable and then use it here.

lib/ble.toit
```diff
@@ -517,134 +512,381 @@ class Device:
...
-      ble_scan_stop_ resource_group_
+      ble_scan_stop_ resource_
+      resource_state_.wait_for_state COMPLETED_EVENT_
 
-  wait_for_client_connected -> none:
-    resource_state_.wait_for_state CONNECTED_EVENT_
-    resource_state_.clear_state CONNECTED_EVENT_
+/**
+The manager for advertising and managing local services.
+*/
```
floitsch, line 549:
```suggestion
```

lib/ble.toit
```diff
@@ -517,134 +512,381 @@ class Device:
...
 
-ble_set_preferred_mtu_ mtu:
-  #primitive.ble.set_preferred_mtu
+  /**
+  Starts advertising the $data.
+
+  The data is advertised once every $interval.
 
-ble_init_ config_resource_group:
+  The advertise will include the given $connection_mode, use one
```
floitsch, line 564:
```suggestion
  The advertise includes the given $connection_mode, use one
```

lib/ble.toit
```diff
@@ -517,134 +512,381 @@ class Device:
...
+    of the BLE_CONNECTION_MODE_* constants.
```
floitsch, line 565:
```suggestion
    of the BLE_CONNECT_MODE_* constants (see $BLE_CONNECT_MODE_NONE and similar).
```

lib/ble.toit
```diff
@@ -517,134 +512,381 @@ class Device:
...
+
+  If the adapter does not support parts of the advertise content, INVALID_ARGUMENT is thrown.
```
floitsch, line 567: ?
```suggestion
  Throws, if the adapter does not support parts of the advertise content.
```

lib/ble.toit
```diff
@@ -517,134 +512,381 @@ class Device:
...
+  For example, on MacOS manufacturing data can not be specified.
+
+  If the adapter does not allow configuration of $interval or $connection_mode an INVALID_ARGUMENT is thrown.
```
floitsch, line 570: ?
```suggestion
  Throws, if the adapter does not allow configuration of $interval or $connection_mode.
```

lib/ble.toit
```diff
@@ -517,134 +512,381 @@ class Device:
...
+      raw_service_classes
+      data.manufacturer_data
+      interval.in_us
+      connection_mode
+
+    state := resource_state_.wait_for_state ADVERTISE_START_SUCEEDED_EVENT_ | ADVERTISE_START_FAILED_EVENT_
+    if state & ADVERTISE_START_FAILED_EVENT_ != 0: throw "Failed to start advertising"
+
+  /**
+    Stops advertising.
```
floitsch, line 596:
```suggestion
  Stops advertising.
```

lib/ble.toit
```diff
@@ -517,134 +512,381 @@ class Device:
...
+  */
+  stop_advertise:
+    ble_advertise_stop_ resource_
+
+  /**
+    Adds a new service to the peripheral identified by $uuid. The returned service should be configured with
```
floitsch, line 602:
```suggestion
  Adds a new service to the peripheral identified by $uuid. The returned service should be configured with
```

floitsch (commented):
First round of comments. (Need to leave now, but will come back for a second round).

src/resources/ble_darwin.mm
```diff
@@ -0,0 +1,1297 @@
...
+  void append_data(NSData* data) {
+    _characteristic_data_list.append(_new CharacteristicData(data));
+  }
+
+  CharacteristicData *remove_first() {
+    return _characteristic_data_list.remove_first();
+  }
+
+  void put_back(CharacteristicData* data) {
+    _characteristic_data_list.prepend(data);
```
floitsch, line 198: Feels weird that the function is called `put_back` but we are calling `prepend` here.

src/resources/ble_darwin.mm
```diff
@@ -0,0 +1,1297 @@
...
+  void add_central(CBCentral *central) {
+    [_subscriptions addObject:[central retain]];
+  }
+
+  void remove_central(CBCentral *central) {
+    [_subscriptions removeObjectIdenticalTo:central];
+  }
+
+  int mtu() {
+    int min_mtu = 1 << 16; // MTU should maximum be 16 bit
```
floitsch, line 212:
```suggestion
    int min_mtu = 1 << 16; // MTU should maximum be 16 bit.
```

src/resources/ble_darwin.mm
```diff
@@ -0,0 +1,1297 @@
...
+    for (int i=0; i < [_subscriptions count]; i++) {
```
floitsch, line 213: Unless this is convention in darwin code.
```suggestion
    for (int i = 0; i < [_subscriptions count]; i++) {
```

src/resources/ble_darwin.mm
```diff
@@ -0,0 +1,1297 @@
...
+      min_mtu = MIN(min_mtu, _subscriptions[i].maximumUpdateValueLength);
+    }
+    return min_mtu==1<<16?23:min_mtu; // 23 is the default mtu value in BLE.
```
floitsch, line 216: Should this be a `<` ?
```suggestion
    return min_mtu== 1 < 16 ? 23 : min_mtu; // 23 is the default mtu value in BLE.
```

src/resources/ble_darwin.mm
```diff
@@ -0,0 +1,1297 @@
...
+- (id)initWithDevice:(toit::BLERemoteDeviceResource*)gatt {
+  self.device = gatt;
+  return self;
+}
+
+- (void)peripheral:(CBPeripheral*)peripheral didDiscoverServices:(NSError*)error {
+  if (peripheral.delegate != nil) {
+    toit::BLERemoteDeviceResource* device = ((TOITPeripheralDelegate*) peripheral.delegate).device;
+    if (error) {
+      //device->set_error(error);
```
floitsch, line 373: Not necessary?
```suggestion
      //device->set_error(error);
```

src/resources/ble_darwin.mm
```diff
@@ -0,0 +1,1297 @@
...
+
+- (void)                  peripheral:(CBPeripheral*)peripheral
+didDiscoverCharacteristicsForService:(CBService*)service
+                               error:(NSError*)error {
+  if (peripheral.delegate != nil) {
+    toit::BLERemoteDeviceResource* device = ((TOITPeripheralDelegate*) peripheral.delegate).device;
+    toit::BLEServiceResource* service_resource = device->get_or_create_service_resource(service);
+    if (service_resource == null) return;
+    if (error) {
+      //service_resource->set_error(error);
```
floitsch, line 389: not necessary?

src/resources/ble_darwin.mm
```diff
@@ -0,0 +1,1297 @@
...
+  CBPeripheralManager* _peripheral_manager;
+  NSMutableDictionary<CBUUID*, BLEResourceHolder*>* _service_resource_index;
+
+};
+
+BLECharacteristicResource* lookup_remote_characteristic_resource(CBPeripheral* peripheral, CBCharacteristic* characteristic);
+BLECharacteristicResource* lookup_local_characteristic_resource(CBPeripheralManager* peripheral_manager, CBCharacteristic* characteristic);
+}
+
+@interface TOITPeripheralDelegate : NSObject <CBPeripheralDelegate>
```
floitsch, line 357:
```suggestion
@interface ToitPeripheralDelegate : NSObject <CBPeripheralDelegate>
```

src/resources/ble_darwin.mm
```diff
@@ -0,0 +1,1297 @@
...
+    if (!custom_data) ALLOCATION_FAILED;
+    memcpy(custom_data_bytes.address(), manufacturer_data.bytes, [manufacturer_data length]);
+
+    array->at_put(4, custom_data);
+  }
+
+  NSNumber* is_connectable = peripheral->connectable();
+  array->at_put(5, (is_connectable != nil && is_connectable.boolValue == YES)
+                   ? process->program()->true_object()
+                   : process->program()->false_object());
```
floitsch, line 878: I believe this works.
```suggestion
  array->at_put(5, BOOL(is_connectable != nil && is_connectable.boolValue == YES));
```

src/resources/ble_darwin.mm
```diff
@@ -0,0 +1,1297 @@
...
+
+  resource->resource_group()->unregister_resource(resource);
+
+  return process->program()->null_object();
+}
+
+PRIMITIVE(discover_services) {
+  ARGS(BLERemoteDeviceResource, device, Array, raw_service_uuids);
+
+  Error* err;
```
floitsch, line 941: I know the `ns_uuid_array_from_array_of_strings` function does it for us, but it doesn't really cost to set it here.
```suggestion
  Error* err = null;
```

src/resources/ble_darwin.mm
```diff
@@ -0,0 +1,1297 @@
...
+
+  return array;
+}
+
+PRIMITIVE(discover_characteristics) {
+  ARGS(BLEServiceResource, service, Array, raw_characteristics_uuids);
+
+  if (!service->device()) INVALID_ARGUMENT;
+
+  Error* err;
```
floitsch, line 993:
```suggestion
  Error* err = null;
```

src/resources/ble_darwin.mm
```diff
@@ -0,0 +1,1297 @@
...
+PRIMITIVE(discover_characteristics_result) {
+  ARGS(BLEServiceResource, service);
+
+  NSArray<CBCharacteristic*>* characteristics = [service->service() characteristics];
+
+  int count = 0;
+  BLECharacteristicResource* characteristic_resources[([characteristics count])];
+  for (int i = 0; i < [characteristics count]; i++) {
+    BLECharacteristicResource* characteristic_resource
+        = service->get_or_create_characteristic_resource(characteristics[i],true);
```
floitsch, line 1012: unless this is convention in darwin code.
```suggestion
        = service->get_or_create_characteristic_resource(characteristics[i], true);
```

src/resources/ble_darwin.mm
```diff
@@ -0,0 +1,1297 @@
...
+
+PRIMITIVE(advertise_start) {
+  ARGS(BLEPeripheralManagerResource, peripheral_manager, Blob, name, Array, service_classes,
+       Blob, manufacturing_data, int, interval_us, int, conn_mode);
+  USE(interval_us);
+  USE(conn_mode);
+
+  NSMutableDictionary* data = [NSMutableDictionary new];
+
+  if (manufacturing_data.length()>0) INVALID_ARGUMENT;
```
floitsch, line 1124:
```suggestion
  if (manufacturing_data.length() > 0) INVALID_ARGUMENT;
```

floitsch (commented):
Next round of comments.
Still need to review the ESP32 code.

lib/ble.toit
```diff
@@ -4,60 +4,62 @@
...
-*/
-class Address:
-  raw_/ByteArray
-
-  constructor .raw_:
+import encoding.hex
+import device
+
+class BleUUID:
+  data_/any
```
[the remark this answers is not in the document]
mikkeldamsgaard: It might. The two different implementations (NimBLE and macos) uses different representations, so I kept it native to the representation to save converting back and forth when calling primitives.

lib/ble.toit
```diff
@@ -78,78 +80,21 @@ class AdvertisementData:
...
 /**
 A remote device discovered by a scanning.
 */
-class RemoteDevice:
+class RemoteScannedDevice:
   /**
   The BLE address of the remote device.
   */
-  address/Address
+  address/any
```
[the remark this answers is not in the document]
mikkeldamsgaard: Gone.
It is not possible to set this from toit, it is provided by the lowlevel implementation. MacOS sets it to a temporary uuid and NimBLE sets it as the ble "mac address". Any suggestions as to expose this in more detail is welcome, but for now it is hidden.

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
+  service/RemoteService
+  uuid/BleUUID
+  properties/int
+  discovered_descriptors/List := []
   /**
-  Constructs a remote service from the given $client_, $service_id, and $handle_range_.
+  Constructs remote characteristic from the given $service, $handle, and $definition_handle.
   */
-  constructor .client_ .service_id .handle_range_:
+  constructor .service .uuid .properties characteristic:
```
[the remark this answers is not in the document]
mikkeldamsgaard: Good catch

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
-  Reads a remote characteristic on the remote service by looking up the handle of the given $characteristic_uuid.
+  Reads the value of the characteristic on the remote device.
+  */
+  value -> ByteArray?:
+    if properties & CHARACTERISTIC_PROPERTY_READ == 0:
+      throw "Characteristic does not support reads"
+
+    return request_read_
+
+  notification -> ByteArray?:
```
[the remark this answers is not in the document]
mikkeldamsgaard: Is there a cleverer name? 
receive_notification?
notification_read?
on_notification?

lib/ble.toit
mikkeldamsgaard: CentralManager.connect?

lib/ble.toit
```diff
@@ -517,134 +512,381 @@ class Device:
...
+class AdapterInformation:
+  identifier/string
+  address/ByteArray
+  supports_central_role/bool
+  supports_peripheral_role/bool
+  handle_/any
+  constructor .identifier .address .supports_central_role .supports_peripheral_role .handle_:
+
+class Adapter:
+  static adapters -> List/*<AdapterInfo>*/:
```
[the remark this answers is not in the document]
mikkeldamsgaard: There are some side effects, in that the platform actually initializes the adapter in the constructor. I believe that is why I kept them separate. So if you have multiple adapters, there might be other processes/programs that needs the adapter and this would cause issues. I will rename the name. Also, only relevant on windows and linux, that isn't implemented yet.

lib/ble.toit
```diff
@@ -517,134 +512,381 @@ class Device:
...
+
+  /**
+    Adds a new service to the peripheral identified by $uuid. The returned service should be configured with
+    the appropriate characteristics and then be deployed.
+  */
+  add_service uuid/BleUUID -> LocalService:
+    return LocalService this uuid
+
+
+class AdapterInformation:
```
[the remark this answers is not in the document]
mikkeldamsgaard: See above comment

lib/ble.toit
```diff
@@ -517,134 +512,381 @@ class Device:
...
+SERVICE_ADD_FAILED_EVENT_             ::= 1 << 19
+DATA_RECEIVED_EVENT_                  ::= 1 << 20
+
+
+
+order_attributes_ input/List/*<BleUUID>*/ output/List/*<Attribute>*/ -> List:
+  map := {:}
+  if input.is_empty: return output
+  output.do: | attribute/Attribute | map[attribute.uuid] = attribute
+  return input.map: | uuid/BleUUID | map.get uuid
```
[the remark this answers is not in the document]
mikkeldamsgaard: Not really, if the user asks to discover a UUID that is not present on the remote, it will not be there.

lib/ble.toit
```diff
@@ -171,339 +116,389 @@ class RemoteDevice:
...
-        : null
-    resource_group_ = ble_init_ server_configuration_resource_group
-    add_finalizer this:: this.close
-    ble_set_preferred_mtu_ preferred_mtu
-    try:
-      gap := ble_gap_ resource_group_
-      resource_state := ResourceState_ resource_group_ gap
-      state := resource_state.wait_for_state STARTED_EVENT_
-      resource_state_ = resource_state
+class CentralManager extends Resource_:
```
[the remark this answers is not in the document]
mikkeldamsgaard: Not just macos. Windows too: https://learn.microsoft.com/en-us/windows/uwp/devices-sensors/bluetooth-low-energy-overview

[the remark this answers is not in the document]
mikkeldamsgaard: The kind-of-accepted vocabulary does divide this into "Central" and "Peripheral". The "*Manager" part I agree, any name is better..

lib/ble.toit
```diff
@@ -517,134 +512,381 @@ class Device:
...
+    if state & VALUE_DATA_READ_FAILED_EVENT_ != 0: throw_error_
+    return ble_get_value_ resource_
+
+
+class LocalReadWriteElement_ extends Resource_:
+  constructor service/LocalService resource:
+    super service.peripheral_manager.adapter.resource_group_ resource
+
+  get_value_ -> ByteArray:
+    resource_state_.clear_state DATA_RECEIVED_EVENT_
```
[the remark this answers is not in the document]
mikkeldamsgaard: That should not be necessary. The loop should only loop at most 2 times. Second time we are guaranteed that the buf is not null

src/resources/ble_darwin.mm
mikkeldamsgaard: suggestion?

src/resources/ble_darwin.mm
mikkeldamsgaard: You should try it :)
The issue is that exactly for objective-c, they have their own BOOL macro....

src/resources/ble_darwin.mm
mikkeldamsgaard: It is not. The problem is my editor that does some "clever" annotation, so I dont always catch these missing spaces

src/event_sources/ble_host.h
```diff
@@ -0,0 +1,65 @@
...
+// directory of this repository.
+
+#pragma once
+
+#include "../resource.h"
+#include "ble.h"
+
+namespace toit {
+
+// RAII helper class to just lock the mutex from a non-toit thread
```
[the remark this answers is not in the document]
mikkeldamsgaard: I know. But the Locker class needs it to run from a toit thread, and that doesnt work on macos. The thread is external.

lib/ble.toit
floitsch: True. But it's not a "connect", "disconnect", "connect", dance, where you can do this multiple times.

Also, for sockets we do a connect, and then call `close` on the object.

lib/ble.toit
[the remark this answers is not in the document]
floitsch: Add a comment then.

lib/ble.toit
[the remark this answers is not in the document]
floitsch: Fine with "Central" and "Peripheral" then.

src/resources/ble_darwin.mm
floitsch: Maybe `unget`.
But Erik interpreted the name `put_back` as just returning it. I read it similar to `push_back` in C++...

src/resources/ble_darwin.mm
floitsch: Aah. you undefed the Toit one in line 24, and not the macos one...

src/resources/ble_darwin.mm
```diff
@@ -0,0 +1,1297 @@
...
+
+#include "../top.h"
+
+#include "../objects.h"
+#include "../objects_inline.h"
+#include "../event_sources/ble_host.h"
+
+#undef BOOL
+
+#import <CoreBluetooth/CoreBluetooth.h>
```
floitsch, line 26: Generally, we try to have system includes earlier than local ones. (Assuming that "import" is like an include).
I don't remember exactly why anymore, but there were some issues that appeared otherwise.

floitsch: That said. It probably doesn't matter, and with the `BOOL` undef it's easier this way.

mikkeldamsgaard: Which is a good idea, but, there are issues with #defines if it is done in any other order.

src/event_sources/ble_host.h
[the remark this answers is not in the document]
erikcorry: Perhaps Thread::current in os_posix.cc can be changed to create the Thread object on demand if it finds itself running on a thread that has no Thread object?

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
-  }
-
+  if (permissions & 0x4) flags |= BLE_GATT_CHR_F_READ_ENC;
+  if (permissions & 0x8) flags |= BLE_GATT_CHR_F_WRITE_ENC;
 
-  return Primitive::integer(gatt->result(), process);
-}
+  BLECharacteristicResource* characteristic =
+    service_resource->get_or_create_characteristics_resource(
+        ble_uuid, flags,0,0,true);
```
floitsch, line 1816:
```suggestion
      service_resource->get_or_create_characteristics_resource(ble_uuid, flags,0,0,true);
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
+  os_mbuf* om = null;
+  Object* error = object_to_mbuf(process, value, &om);
+  if (error) return error;
 
-  if (gatt->error() == BLE_HS_ENOENT) {
-    OUT_OF_RANGE;
-  } else if (gatt->error() != 0) {
+  uint32 flags = properties & 0x7F;
+  if (permissions & 0x1) flags |= BLE_GATT_CHR_F_READ;
+  if (permissions & 0x2 && !(properties & (BLE_GATT_CHR_PROP_WRITE | BLE_GATT_CHR_F_WRITE_NO_RSP)))
```
floitsch, line 1809: If the properties isn't a write, then it's an invalid argument.
But don't we need to set the flags, when the properties are write?

Also: don't we need to do the same properties test for the `write_enc` below?

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
-    return Primitive::os_error(err, process);
-  }
-
-  return process->program()->null_object();
-}
-
-PRIMITIVE(request_service) {
-  ARGS(GATTResource, gatt, Blob, uuid);
+  uint8 flags = 0;
+  if (properties & BLE_GATT_CHR_F_READ || permissions & 0x01) flags |= BLE_ATT_F_READ;
```
floitsch, line 1845: Confusing that the order in the condition is inverted (compared to add_characteristic). In the other one we first look at `permissions`.

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
+  if (properties & (BLE_GATT_CHR_F_WRITE & BLE_GATT_CHR_F_WRITE_NO_RSP) || permissions & 0x02) flags |= BLE_ATT_F_WRITE;
```
floitsch, line 1846: Shouldn't that be a logical or instead of `&` (for the `F_WRITE*`).

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
+  if (permissions & 0x04) flags |= BLE_ATT_F_READ_ENC;
```
floitsch, line 1847: Add a comment reminding the reader that "ENC" stands for "encrypted". Also in `add_characteristic`.
Maybe also, why the encrypted version behaves differently from the unencrypted one.

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
+  BLEDescriptorResource* descriptor =
+      characteristic->get_or_create_descriptor(ble_uuid, 0, flags, true);
+  if (!descriptor) {
+    if (om != null) os_mbuf_free(om);
+    MALLOC_FAILED;
+  }
 
-  gatt->set_result(UINT32_MAX);
+  if (om != null)
+    descriptor->set_mbuf_to_send(om);
```
floitsch, line 1858:
```suggestion
  if (om != null) descriptor->set_mbuf_to_send(om);
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
-  int err = ble_gattc_disc_svc_by_uuid(
-      gatt->handle(), &ble_uuid.u, BLEEventSource::on_gatt_service, gatt);
+  proxy->set_external_address(descriptor);
+  return proxy;
+}
 
-  if (err != ESP_OK) {
-    return Primitive::os_error(err, process);
+void clean_up_gatt_svr_chars(ble_gatt_chr_def* gatt_svr_chars, int count) {
+  for (int i=0;i<count;i++) {
```
floitsch, line 1865:
```suggestion
  for (int i = 0; i < count; i++) {
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
+    for (const auto &item: characteristic->descriptors()) {
+      USE(item);
+      descriptor_cnt++;
+    }
 
-PRIMITIVE(request_attribute) {
-  ARGS(GATTResource, gatt, uint16, handle);
+    if (descriptor_cnt>0) {
+      auto gatt_desc_defs = static_cast<ble_gatt_dsc_def*>(
+          calloc(descriptor_cnt+1,sizeof(ble_gatt_dsc_def)));
```
floitsch, line 1903:
```suggestion
      auto gatt_desc_defs = static_cast<ble_gatt_dsc_def*>(calloc(descriptor_cnt + 1, sizeof(ble_gatt_dsc_def)));
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
+  int characteristic_cnt = 0;
+  for (const auto &i: service_resource->characteristics()) {
+    USE(i);
+    characteristic_cnt++;
   }
 
-  return process->program()->null_object();
-}
+  auto gatt_svr_chars = static_cast<ble_gatt_chr_def*>(
+      calloc(characteristic_cnt + 1, sizeof(ble_gatt_chr_def)));
```
floitsch, line 1884:
```suggestion
  auto gatt_svr_chars = static_cast<ble_gatt_chr_def*>(calloc(characteristic_cnt + 1, sizeof(ble_gatt_chr_def)));
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
-/*
- *
- * Primitives for BLE server.
- *
- */
-PRIMITIVE(server_configuration_init) {
-  ByteArray* proxy = process->object_heap()->allocate_proxy();
-  if (!proxy) ALLOCATION_FAILED;
+      int descriptor_idx = 0;
+      for (const auto &descriptor: characteristic->descriptors()) {
```
floitsch, line 1912:
```suggestion
      for (const auto& descriptor : characteristic->descriptors()) {
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
+    free(gatt_services);
+    free(gatt_svr_chars);
+    return nimle_stack_error(process, rc);
+  }
 
-PRIMITIVE(server_configuration_dispose) {
-  ARGS(BLEServerConfigGroup, group);
-  group->tear_down();
+  // nimble does not do async service deployments, so
+  // simulate success event
```
floitsch, line 1944:
```suggestion
  // simulate success event.
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
-  proxy->set_external_address(characteristic);
-  return proxy;
+  return process->program()->null_object();
 }
 
-PRIMITIVE(set_characteristics_value) {
-  ARGS(BLEServerCharacteristicResource, resource, Object, value);
+PRIMITIVE(get_subscribed_clients) {
+  ARGS(BLECharacteristicResource, characteristic)
+  int cnt = 0;
```
floitsch, line 1968:
```suggestion
  int count = 0;
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
+  for (const auto &item: characteristic->subscriptions()) {
```
floitsch, line 1969:
```suggestion
  for (const auto& item : characteristic->subscriptions()) {
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
+    USE(item);
+    cnt++;
```
floitsch, line 1971:
```suggestion
    count++;
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
+  }
 
-  os_mbuf* om = null;
-  Object* error = object_to_mbuf(process, value, &om);
-  if (error) return error;
+  Array* array = process->object_heap()->allocate_array(cnt, process->program()->null_object());
```
floitsch, line 1974:
```suggestion
  Array* array = process->object_heap()->allocate_array(count, process->program()->null_object());
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
+  if (!array) ALLOCATION_FAILED;
 
-  resource->set_mbuf_to_send(om);
+  int idx = 0;
```
floitsch, line 1977:
```suggestion
  int index = 0;
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
+  for (const auto &sub: characteristic->subscriptions()) {
+    array->at_put(idx++, Smi::from(sub->conn_handle()));
```
floitsch, line 1979:
```suggestion
    array->at_put(index++, Smi::from(sub->conn_handle()));
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
 
-  os_mbuf* om = null;
-  Object* error = object_to_mbuf(process, value, &om);
-  if (error) return error;
+  Array* array = process->object_heap()->allocate_array(cnt, process->program()->null_object());
+  if (!array) ALLOCATION_FAILED;
 
-  resource->set_mbuf_to_send(om);
+  int idx = 0;
+  for (const auto &sub: characteristic->subscriptions()) {
```
floitsch, line 1978:
```suggestion
  for (const auto& sub : characteristic->subscriptions()) {
```

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
-  if (resource->is_notify_enabled() || resource->is_indicate_enabled()) {
-    os_mbuf* om = null;
-    Object* error = object_to_mbuf(process, value, &om);
-    if (error) return error;
-
-    int err = ESP_OK;
-    if (resource->is_notify_enabled()) {
-      err = ble_gattc_notify_custom(resource->conn_handle(), resource->nimble_value_handle(), om);
+  Subscription* subscription = null;
+  for (const auto &sub: characteristic->subscriptions()) {
```
floitsch, line 1989:
```suggestion
  for (const auto& sub : characteristic->subscriptions()) {
```

src/resources/ble_esp32.cc
```diff
@@ -18,721 +18,1682 @@
...
 
-  BLEServerServiceResource* add_service(ble_uuid_any_t uuid) {
-    BLEServerServiceResource* service = _new BLEServerServiceResource(this, uuid);
-    if (service != null) _services.prepend(service);
-    return service;
+typedef DoubleLinkedList<BLECharacteristicResource> CharacteristicResourceList;
+class BLECharacteristicResource :
+    public BLEReadWriteElement,
+    public CharacteristicResourceList::Element,
+    public DiscoverableResource {
```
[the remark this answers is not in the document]
mikkeldamsgaard: I tried your suggestion, it did not look right...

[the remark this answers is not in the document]
mikkeldamsgaard: I tried something else, looks better.

src/resources/ble_esp32.cc
```diff
@@ -18,721 +18,1682 @@
...
 
-  while (remove_next()) {}
+Object* nimble_error_code_to_string(Process* process, int error_code, bool host) {
+  static const size_t BUFFER_LEN = 400;
+  char buffer[BUFFER_LEN];
+  switch (error_code) { // NOLINT(hicpp-multiway-paths-covered)
+    case NO_CCCD_FOUND_FOR_CHARACTERISTIC:
+      break;
+    default:
+      const char* gist = "https://gist.github.com/mikkeldamsgaard/0857ce6a8b073a52d6f07973a441ad54";
```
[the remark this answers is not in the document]
mikkeldamsgaard: You are welcome to migrate both this and Erik's TLS. It would make a lot of sense.

src/resources/ble_esp32.cc
```diff
@@ -18,721 +18,1682 @@
...
+      for (const auto &current: _characteristics) {
+        if (!characteristic) {
+          if (dsc->handle <= start_handle()) return;
+        } else {
+          if (characteristic->definition_handle() <= dsc->handle && dsc->handle <= characteristic->handle()) return;
+          if (characteristic->handle() < dsc->handle && dsc->handle < current->definition_handle()) break;
+        }
+        characteristic = current;
+      }
+      if (dsc->handle <= characteristic->handle()) return;
```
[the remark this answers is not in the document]
mikkeldamsgaard: No. The way that the descriptors are identified is very painful.

src/resources/ble_esp32.cc
```diff
@@ -18,721 +18,1682 @@
...
 
-      if (fields.mfg_data_len > 0) {
-        ByteArray* custom_data = process->object_heap()->allocate_internal_byte_array(fields.mfg_data_len);
-        if (!custom_data) ALLOCATION_FAILED;
-        ByteArray::Bytes custom_data_bytes(custom_data);
-        memcpy(custom_data_bytes.address(), fields.mfg_data, fields.mfg_data_len);
-        array->at_put(4, custom_data);
-      }
-    }
+    descriptor->set_returned(true);
```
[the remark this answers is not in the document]
mikkeldamsgaard: good catch. Also for characteristics

src/resources/ble_esp32.cc
```diff
@@ -763,307 +1724,326 @@ PRIMITIVE(advertise_config) {
...
+    return nimle_stack_error(process,err);
   }
 
+  peripheral_manager->advertising_params().conn_mode = conn_mode;
+
+  // TODO(anders): Be able to tune this.
+  peripheral_manager->advertising_params().disc_mode = BLE_GAP_DISC_MODE_GEN;
+  peripheral_manager->advertising_params().itvl_min =
+      peripheral_manager->advertising_params().itvl_max =
+          interval_us / 625;
```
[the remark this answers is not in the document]
mikkeldamsgaard: It would indeed. I have no explanation. It was there already when I picked up the rework.

src/resources/ble_esp32.cc
mikkeldamsgaard: The reason this is here in the first place is due to the way that macos defines this. It could be argued that the mac implementation should adhere to the nimble, but in this case I turned it around. The code can not know from permissions&0x2 if it is a write with or without response and therefore it does not try to be clever.

src/resources/ble_esp32.cc
```diff
@@ -18,721 +18,1682 @@
...
 
-void BLEServerConfigGroup::set_subscription_status(uint16 attr_handle, uint16 conn_handle, bool indicate, bool notify) {
-  for (auto service: _services) {
-    for (auto characteristic: service->characteristics()) {
-      if (characteristic->nimble_value_handle() == attr_handle) {
-        characteristic->set_subscription_status(indicate, notify, conn_handle);
-        return;
-      }
-    }
+static bool uuid_equals(ble_uuid_any_t& uuid, ble_uuid_any_t& other) {
```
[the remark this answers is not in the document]
mikkeldamsgaard: Maybe.. Not even quite sure that "normal" implementors even consider that the 16 and 32 bit versions are abbreviated full uuids. So it might raise confusion.

examples/ble/heart_rate.toit
```diff
@@ -14,33 +14,37 @@ main:
...
 
   service := peripheral_manager.add_service
-      BleUUID GATT_IO_UUID
+      BleUuid GATT_IO_UUID
 
   /* Characteristic: SEND DATA */
-  heart_rate_send := service.add_notification_characteristic
-      BleUUID #[0x63, 0x4b, 0x3c, 0x6e, 0xac, 0x41, 0x40, 0x85,
+  heart_rate_send := service.add_characteristic
+      --notification
```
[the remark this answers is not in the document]
mikkeldamsgaard: No worries.
Reverted.

src/resources/ble_esp32.cc
```diff
@@ -76,22 +75,26 @@ class DiscoveredPeripheral : public DiscoveredPeripheralList::Element {
   uint8 _event_type;
 };
 
-
+// The thread on the BLEResourceGroup is responsible for running the nimble background thread. All events from the
+// nimble background thread is delivered through call backs to registered callback methods. The callback methods
```
floitsch, line 79:
```suggestion
// nimble background thread are delivered through call backs to registered callback methods. The callback methods
```

floitsch (approved):
Still LGTM!

floitsch:
The build failed. Do we need to install something on the bot? (and update the build instructions).
```
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_CBMutableCharacteristic", referenced from:
      objc-class-ref in libtoit_vm.a(ble_darwin.mm.o)
  "_OBJC_CLASS_$_NSArray", referenced from:
      objc-class-ref in libtoit_vm.a(ble_darwin.mm.o)
  "_OBJC_CLASS_$_CBMutableService", referenced from:
      objc-class-ref in libtoit_vm.a(ble_darwin.mm.o)
  "_OBJC_CLASS_$_NSUUID", referenced from:
      objc-class-ref in libtoit_vm.a(ble_darwin.mm.o)
  "_OBJC_CLASS_$_CBCentralManager", referenced from:
      objc-class-ref in libtoit_vm.a(ble_darwin.mm.o)
  "_OBJC_CLASS_$_NSMutableDictionary", referenced from:
      objc-class-ref in libtoit_vm.a(ble_darwin.mm.o)
  "_OBJC_CLASS_$_CBPeripheralManager", referenced from:
      objc-class-ref in libtoit_vm.a(ble_darwin.mm.o)
  "_OBJC_CLASS_$_NSMutableArray", referenced from:
      objc-class-ref in libtoit_vm.a(ble_darwin.mm.o)
  "_OBJC_CLASS_$_CBUUID", referenced from:
      objc-class-ref in libtoit_vm.a(ble_darwin.mm.o)
  "_OBJC_CLASS_$_NSString", referenced from:
      objc-class-ref in libtoit_vm.a(ble_darwin.mm.o)
  "_OBJC_CLASS_$_NSData", referenced from:
      objc-class-ref in libtoit_vm.a(ble_darwin.mm.o)
  "_OBJC_METACLASS_$_NSObject", referenced from:
      _OBJC_METACLASS_$_ToitPeripheralDelegate in libtoit_vm.a(ble_darwin.mm.o)
      _OBJC_METACLASS_$_TOITCentralManagerDelegate in libtoit_vm.a(ble_darwin.mm.o)
      _OBJC_METACLASS_$_TOITPeripheralManagerDelegate in libtoit_vm.a(ble_darwin.mm.o)
      _OBJC_METACLASS_$_BLEResourceHolder in libtoit_vm.a(ble_darwin.mm.o)
  "_OBJC_CLASS_$_NSObject", referenced from:
      _OBJC_CLASS_$_ToitPeripheralDelegate in libtoit_vm.a(ble_darwin.mm.o)
      _OBJC_CLASS_$_TOITCentralManagerDelegate in libtoit_vm.a(ble_darwin.mm.o)
      _OBJC_CLASS_$_TOITPeripheralManagerDelegate in libtoit_vm.a(ble_darwin.mm.o)
      _OBJC_CLASS_$_BLEResourceHolder in libtoit_vm.a(ble_darwin.mm.o)
  "__objc_empty_cache", referenced from:
      _OBJC_METACLASS_$_ToitPeripheralDelegate in libtoit_vm.a(ble_darwin.mm.o)
      _OBJC_CLASS_$_ToitPeripheralDelegate in libtoit_vm.a(ble_darwin.mm.o)
      _OBJC_METACLASS_$_TOITCentralManagerDelegate in libtoit_vm.a(ble_darwin.mm.o)
      _OBJC_CLASS_$_TOITCentralManagerDelegate in libtoit_vm.a(ble_darwin.mm.o)
      _OBJC_METACLASS_$_TOITPeripheralManagerDelegate in libtoit_vm.a(ble_darwin.mm.o)
      _OBJC_CLASS_$_TOITPeripheralManagerDelegate in libtoit_vm.a(ble_darwin.mm.o)
      _OBJC_METACLASS_$_BLEResourceHolder in libtoit_vm.a(ble_darwin.mm.o)
      ...
  "_NSLog", referenced from:
      -[TOITPeripheralManagerDelegate peripheralManagerDidStartAdvertising:error:] in libtoit_vm.a(ble_darwin.mm.o)
      -[TOITPeripheralManagerDelegate peripheralManager:didAddService:error:] in libtoit_vm.a(ble_darwin.mm.o)
  "_CBAdvertisementDataIsConnectable", referenced from:
      -[TOITCentralManagerDelegate centralManager:didDiscoverPeripheral:advertisementData:RSSI:] in libtoit_vm.a(ble_darwin.mm.o)
  "_objc_release", referenced from:
      toit::BLECharacteristicResource::set_error(NSError*) in libtoit_vm.a(ble_darwin.mm.o)
      -[TOITCentralManagerDelegate centralManager:didFailToConnectPeripheral:error:] in libtoit_vm.a(ble_darwin.mm.o)
      -[TOITCentralManagerDelegate centralManager:didDisconnectPeripheral:error:] in libtoit_vm.a(ble_darwin.mm.o)
      toit::DiscoveredPeripheral::~DiscoveredPeripheral() in libtoit_vm.a(ble_darwin.mm.o)
  "___CFConstantStringClassReference", referenced from:
      CFString in libtoit_vm.a(ble_darwin.mm.o)
  "_objc_msgSend", referenced from:
      -[ToitPeripheralDelegate initWithDevice:] in libtoit_vm.a(ble_darwin.mm.o)
      -[ToitPeripheralDelegate peripheral:didDiscoverServices:] in libtoit_vm.a(ble_darwin.mm.o)
      -[ToitPeripheralDelegate peripheral:didDiscoverCharacteristicsForService:error:] in libtoit_vm.a(ble_darwin.mm.o)
      -[ToitPeripheralDelegate peripheral:didUpdateValueForCharacteristic:error:] in libtoit_vm.a(ble_darwin.mm.o)
      toit::lookup_remote_characteristic_resource(CBPeripheral*, CBCharacteristic*) in libtoit_vm.a(ble_darwin.mm.o)
      -[ToitPeripheralDelegate peripheralIsReadyToSendWriteWithoutResponse:] in libtoit_vm.a(ble_darwin.mm.o)
      -[TOITCentralManagerDelegate initWithCentralManagerResource:] in libtoit_vm.a(ble_darwin.mm.o)
      ...
  "_objc_retain", referenced from:
      toit::BLECharacteristicResource::set_error(NSError*) in libtoit_vm.a(ble_darwin.mm.o)
      toit::BLECharacteristicResource::append_data(NSData*) in libtoit_vm.a(ble_darwin.mm.o)
      -[TOITPeripheralManagerDelegate peripheralManager:central:didSubscribeToCharacteristic:] in libtoit_vm.a(ble_darwin.mm.o)
      toit::DiscoveredPeripheral::DiscoveredPeripheral(CBPeripheral*, NSNumber*, NSArray*, NSNumber*, NSData*) in libtoit_vm.a(ble_darwin.mm.o)
  "_CBAdvertisementDataManufacturerDataKey", referenced from:
      -[TOITCentralManagerDelegate centralManager:didDiscoverPeripheral:advertisementData:RSSI:] in libtoit_vm.a(ble_darwin.mm.o)
  "_CBAdvertisementDataServiceUUIDsKey", referenced from:
      -[TOITCentralManagerDelegate centralManager:didDiscoverPeripheral:advertisementData:RSSI:] in libtoit_vm.a(ble_darwin.mm.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

lib/ble.toit
```diff
@@ -171,339 +152,422 @@ class RemoteDevice:
...
+        --properties=CHARACTERISTIC_PROPERTY_READ
+        --permissions=CHARACTERISTIC_PERMISSION_READ
+        --value=value
 
-  add_read_write_characteristic uuid --value=#[]-> ReadWriteCharacteristic:
-    char := ReadWriteCharacteristic this uuid value
-    characteristics_[uuid] = char
-    return char
+  /**
+  Convenience method to add a write-only characteristic with the given $uuid that can $require_response for each write.
```
floitsch, line 440:
```suggestion
  Convenience method to add a write-only characteristic with the given $uuid that can $requires_response for each write.
```

CMakeLists.txt
```diff
@@ -145,6 +145,7 @@ set(CMAKE_CXX_FLAGS_ASAN "${CMAKE_CXX_FLAGS_ASAN} -DTOIT_DEBUG -DTOIT_ASAN")
 if ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
     set(TOIT_LINK_GC_FLAGS "-Wl,-dead_strip")
     set(TOIT_LINK_GROUP_BEGIN_FLAGS "-Wl,-all_load")
+    list(APPEND TOIT_LINK_LIBS "-framework Foundation" "-framework CoreBluetooth" ObjC)
```
floitsch, line 148:
```suggestion
```

floitsch:
Following patch still needs to be applied. That should be enough to get the build green.
``` diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index ece2bc39..a257b627 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -166,6 +166,10 @@ else()
   set(TOIT_WINDOWS_LIBS )
 endif()

+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+  set(TOIT_BLUETOOTH_LIBS "-framework Foundation" "-framework CoreBluetooth" ObjC)
+endif()
+
 # Because of the `CACHE INTERNAL ""` at the end of the `set` we can
 #   use this variable outside of the directory.
 set(TOIT_LINK_LIBS
@@ -181,13 +185,10 @@ set(TOIT_LINK_LIBS
   ${CMAKE_DL_LIBS}
   ${TOIT_LINK_LIBS_LIBGCC}
   ${TOIT_LINK_SEGFAULT}
+  ${TOIT_BLUETOOTH_LIBS}
   CACHE INTERNAL ""
   )

-if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-  list(APPEND TOIT_LINK_LIBS "-framework Foundation" "-framework CoreBluetooth" ObjC)
-endif()
-
 target_link_libraries(
   toit_vm
   ${TOIT_WINDOWS_LIBS}
```

floitsch:
Could you merge master into this PR?
The TLS of microsoft changed, and this PR is currently failing on the buildbot because of it.

mikkeldamsgaard:
> Could you merge master into this PR? The TLS of microsoft changed, and this PR is currently failing on the buildbot because of it.

done

PR #1122, opened 2022-10-13
Add drom extension marker

kasperl:


src/embedded_data.cc
```diff
@@ -0,0 +1,135 @@
...
+  const uint32* header = reinterpret_cast<const uint32*>(this);
+  const uword* table = reinterpret_cast<const uword*>(&header[HEADER_WORDS]);
+  return EmbeddedImage{
+    .program = reinterpret_cast<const Program*>(table[n * 2]),
+    .size    = table[n * 2 + 1]
+  };
+}
+
+List<uint8> EmbeddedDataExtension::config() const {
+  const uint32* header = reinterpret_cast<const uint32*>(this);
```
erikcorry, line 48: This function would benefit from a block comment explaining the layout of the data we are
looking at.

src/embedded_data.h
```diff
@@ -0,0 +1,64 @@
...
+  uword offset(const Program* program) const;
+  const Program* program(uword offset) const;
+
+  static const EmbeddedDataExtension* cast(const void* pointer);
+
+ private:
+  static const uint32 HEADER_MARKER   = 0x98dfc301;
+  static const uint32 HEADER_CHECKSUM = 0xb3147ee9;
+
+  static const int HEADER_INDEX_MARKER      = 0;
```
erikcorry, line 56: Instead of this perhaps a Header struct that can be overlaid onto the memory?

src/embedded_data.cc
```diff
@@ -0,0 +1,135 @@
...
+  uint32 used = header[HEADER_INDEX_USED];
+  uint32 free = header[HEADER_INDEX_FREE];
+  if (free < sizeof(uint32)) return List<uint8>();
+  uword address = reinterpret_cast<uword>(header) + used;
+  uint32 size = *reinterpret_cast<uint32*>(address);
+  uint8* data = reinterpret_cast<uint8*>(address + sizeof(uint32));
+  return List<uint8>(data, Utils::min(size, free));
```
erikcorry, line 55: The two values in the `min` are not comparable.

erikcorry (approved):

src/embedded_data.cc
kasperl: Done.

src/embedded_data.cc
kasperl: Done.

PR #1131, opened 2022-10-15
Expose firmware.config

kasperl:


system/extensions/esp32/firmware.toit
```diff
@@ -97,6 +126,15 @@ class FirmwareWriter extends ServiceResource:
         written_ = ota_write_ buffer_
         fullness_ = 0
 
+  pad size/int value/int -> int:
+    return List.chunk_up 0 size (buffer_.size - fullness_) buffer_.size: | from to chunk |
+      end := fullness_ + chunk
+      buffer_.fill --from=fullness_ --to=end value
+      fullness_ = end
+      if fullness_ == buffer_.size:
```
floitsch, line 134: I don't understand the test:
The chunk-up method should make sure that the buffer is filled. The only case where this might not be true, is for the last chunk. Does that mean that we don't want to write the last chunk if it doesn't fully fill the buffer? if yes, add a comment.

system/extensions/esp32/wifi.toit
```diff
@@ -56,6 +57,8 @@ class WifiServiceDefinition extends NetworkServiceDefinitionBase:
     effective := config
     if not effective:
       catch --trace: effective = store_.get WIFI_CONFIG_STORE_KEY
+      if not effective:
```
floitsch, line 60: Feels  a bit inconsistent: the `if not effective` should either all be nested, or none of them.
I think it's ok to move the `if` from line 60 and 62 out.  The more efficient version would move the `if` from line 62 in.

floitsch (approved):
LGTM.

lib/system/firmware.toit
```diff
@@ -86,5 +98,22 @@ class FirmwareWriter extends ServiceResourceProxy:
...
+
   commit --checksum/ByteArray?=null -> none:
     _client_.firmware_writer_commit handle_ checksum
+
+class FirmwareConfig_:
+  /**
+  Returns the configuration entry for the given $key, or
+    null if the $key isn't present in the configuration.
+  */
+  operator [] key/string -> any:
```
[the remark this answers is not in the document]
kasperl: Let's discuss the API next week!

system/extensions/esp32/firmware.toit
```diff
@@ -97,6 +126,15 @@ class FirmwareWriter extends ServiceResource:
         written_ = ota_write_ buffer_
         fullness_ = 0
 
+  pad size/int value/int -> int:
+    return List.chunk_up 0 size (buffer_.size - fullness_) buffer_.size: | from to chunk |
+      end := fullness_ + chunk
```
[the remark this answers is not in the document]
kasperl: Will refactor this a bit, so it is obvious that this is shared with the `write` method.

system/extensions/esp32/firmware.toit
kasperl: Will add comment in follow up PR.

PR #1146, opened 2022-10-25
Add espnow primitive and example

donghengqaz:


lib/espnow.toit
```diff
@@ -0,0 +1,75 @@
...
+    check_mac addr
+    return espnow_send_ addr data wait
+
+  receive -> any:
+    array := espnow_receive_ (Array_ 2)
+    if not array:
+      return null
+    return Datagram array[0] array[1]
+
+  add_peer --addr/ByteArray --channel/int --encrypted/bool --key/ByteArray -> bool:
```
floitsch, line 56: The peer is identified by its address, as such it's the main argument and probably doesn't need any named argument.
```suggestion
  add_peer address/ByteArray --channel/int --encrypted/bool --key/ByteArray -> bool:
```

I would combine this and the next function:
```suggestion
  add_peer address/ByteArray --channel/int --key/ByteArray?=null --encrypted/bool=(key != null) -> bool:
    if not key: key = #[]
```

From what I know, the espnow protocol has a standard key it can use. If that's a good default, then I would make `--encrypted` by default equal to true. This is really a question of common use and how most people do/should work with espnow.

lib/espnow.toit
```diff
@@ -0,0 +1,75 @@
+// Copyright (C) 2022 Toitware ApS. All rights reserved.
```
floitsch, line 1: espnow is an esp32 library.
I would suggest the following:
- create an 'esp32' folder inside the lib directory.
- move `lib/esp32.toit` into `lib/esp32`.  This way `import esp32` continues to work.
- move `lib/espnow.toit` into `lib/esp32`. This way `import esp32.espnow` works.

floitsch (commented):
Thanks.
I'm very happy about the code. There are a lot of comments from me, but many are just nitpicks and things that are very Toit specific.

I will let Erik review the C++ code.
I can already say now, that we will want to listen to state changes in the Toit code, so that `receive` actually waits until there is a message before it attempts to read it. This is a bit more complicated, so hopefully Erik will explain it nicely in his review.

lib/espnow.toit
```diff
@@ -0,0 +1,75 @@
...
+STATION ::= 0 // Wi-Fi station mode
+SOFTAP  ::= 1 // Not support yet
+
+BROADCAST_MAC ::= #[0xff, 0xff, 0xff, 0xff, 0xff, 0xff]
+
+check_mac address/ByteArray:
+  if address.size != 6:
+      throw "ESP-Now MAC address length must be 6 bytes"
+
+check_key key/ByteArray:
```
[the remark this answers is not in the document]
donghengqaz: OK.

lib/espnow.toit
```diff
@@ -0,0 +1,75 @@
...
+class Datagram:
+  addr/ByteArray := ?
+  data/ByteArray := ?
+
+  constructor .addr .data:
+
+class Service:
+  _group ::= ?
+
+  constructor --mode/int=STATION --pmk/ByteArray:
```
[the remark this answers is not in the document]
donghengqaz: Yes, it isn't. If not set, internally wifi can use default one.

lib/espnow.toit
donghengqaz: This key is used to encrypt LMK(Local Master Keys), description is [here](https://docs.espressif.com/projects/esp-idf/en/v4.4.2/esp32/api-reference/network/esp_now.html#security): LMK of the paired device is used to encrypt the vendor-specific action frame with the CCMP method. The maximum number of different LMKs is six. If the LMK of the paired device is not set, the vendor-specific action frame will not be encrypted.
So I think for safety we should make it true and use a default key in this example.

lib/espnow.toit
donghengqaz: OK, it's esp32's function.

lib/espnow.toit
donghengqaz: Why here add `if not key: key = #[]`, can it pass `key=null` to primitive ?

src/resources/espnow.cc
```diff
@@ -0,0 +1,237 @@
...
+  tx_status = status;
+  portBASE_TYPE ret = xSemaphoreGive(tx_sem);
+  if (ret != pdTRUE) {
+    // ESP_LOGE("ESPNow", "Failed to give TX semaphore");
+  }
+}
+
+static void espnow_recv_cb(const uint8_t *mac_addr, const uint8_t *data, int data_len)
+{
+  struct DataGram *datagram = (struct DataGram *)malloc(sizeof(struct DataGram) + data_len);
```
erikcorry, line 94: I have written a short style guide because we have so many unwritten rules.
Most of them align with the Google C++ style guide.
One of them is that we don't allow C-style casts.
See https://github.com/toitlang/toit/tree/master/src#casts

src/resources/espnow.cc
```diff
@@ -0,0 +1,237 @@
...
+
+  proxy->set_external_address(group);
+
+  return proxy;
+}
+
+PRIMITIVE(send) {
+  ARGS(Blob, addr, Blob, data, bool, wait);
+
+  xSemaphoreTake(tx_sem, 0);
```
erikcorry, line 151: If you are taking a semaphore with a timeout of zero should you not check the return value?

src/resources/espnow.cc
```diff
@@ -0,0 +1,237 @@
...
+  ASSERT(rx_queue);
+}
+
+ESPNowResourceGroup::~ESPNowResourceGroup() {
+  tx_sem = NULL;
+}
+
+MODULE_IMPLEMENTATION(espnow, MODULE_ESPNOW)
+
+static void espnow_send_cb(const uint8_t *mac_addr, esp_now_send_status_t status)
```
erikcorry, line 83: Please put the { on the same line as the function in C++.

lib/espnow.toit
floitsch: Yes. You can pass null to primitives as well.
You would need to pay attention with the typing of the C++ code, as you wouldn't be able to take a `Blob` (or whatever type you expect) anymore.
I'm fine either way. I believe we canonicalize the #[], so it wouldn't actually cost any allocation.

donghengqaz: OK, it looks like better to using #[].

src/resources/espnow.cc
```diff
@@ -0,0 +1,227 @@
...
+//
+// The license can be found in the file `LICENSE` in the top level
+// directory of this repository.
+
+#include "../top.h"
+
+// #define TOIT_FREERTOS 1
+#define CONFIG_TOIT_ENABLE_ESPNOW 1
+
+#if defined(TOIT_FREERTOS) && defined(CONFIG_TOIT_ENABLE_ESPNOW)
```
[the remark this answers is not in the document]
donghengqaz: Yes, I will put this as option in `toolchains/idf/components/toit/Kconfig`.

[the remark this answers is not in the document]
donghengqaz: Yes, `UNIMPLEMENTED_PRIMITIVE` is important, but I see it is not used widely, so maybe we can add this later for all needed primitive ones.

src/resources/espnow.cc
donghengqaz: OK, I will read this document and modify the code.

src/resources/espnow.cc
donghengqaz: It takes semaphore here, mainly to `reset` the semaphore for next sending.

src/resources/espnow.cc
```diff
@@ -0,0 +1,237 @@
...
+  if (ret != pdTRUE) {
+    // ESP_LOGE("ESPNow", "Failed to give TX semaphore");
+  }
+}
+
+static void espnow_recv_cb(const uint8_t *mac_addr, const uint8_t *data, int data_len)
+{
+  struct DataGram *datagram = (struct DataGram *)malloc(sizeof(struct DataGram) + data_len);
+  if (!datagram) {
+    // ESP_LOGE("ESPNow", "Failed to malloc datagram");
```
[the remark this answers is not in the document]
donghengqaz: OK, I will try to allocate buffer for espnow RX.

src/resources/espnow.cc
```diff
@@ -0,0 +1,237 @@
...
+    ALLOCATION_FAILED;
+  }
+
+  ByteArray* array = process->allocate_byte_array(datagram->len);
+  if (array == null) {
+    free(datagram);
+    ALLOCATION_FAILED;
+  }
+
+  Array* out = Array::cast(output);
```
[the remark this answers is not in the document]
donghengqaz: This `assert` is to check input object contains 2 elements to, I refer to the UDP code: https://github.com/toitlang/toit/blob/master/src/resources/udp_esp32.cc#L294.

src/resources/espnow.cc
```diff
@@ -0,0 +1,237 @@
...
+  ARGS(Object, output);
+
+  struct DataGram *datagram;
+  portBASE_TYPE ret = xQueueReceive(rx_queue, &datagram, 0);
+  if (ret != pdTRUE) {
+    return process->program()->null_object();
+  }
+
+  ByteArray* address = null;
+  address = process->allocate_byte_array(6);
```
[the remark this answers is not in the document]
donghengqaz: Does the ByteArray allocated by `process->allocate_byte_array` not need to free when failed in some case, the function can return directly ?

src/resources/espnow.cc
```diff
@@ -0,0 +1,237 @@
...
+  if (address == null) {
+    free(datagram);
+    ALLOCATION_FAILED;
+  }
+
+  ByteArray* array = process->allocate_byte_array(datagram->len);
+  if (array == null) {
```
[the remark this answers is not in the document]
donghengqaz: OK, I see some drive like I2C/I2S write like this.

src/resources/espnow.cc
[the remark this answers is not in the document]
erikcorry: That's right, it doesn't need to be freed.  In fact there is no call available to free it.  On the next garbage collection there is no reference to it, and it is collected automatically.

lib/espnow.toit
```diff
@@ -0,0 +1,75 @@
...
+// found in the lib/LICENSE file.
+
+import bytes
+
+STATION ::= 0 // Wi-Fi station mode
+SOFTAP  ::= 1 // Not support yet
+
+BROADCAST_MAC ::= #[0xff, 0xff, 0xff, 0xff, 0xff, 0xff]
+
+check_mac address/ByteArray:
```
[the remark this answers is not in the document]
donghengqaz: Add class `Address` is OK, but some function calling will like:
```
send data/ByteArray address/Address ...
```
then
```
send data (Address address(ByteArrary))
```
Is there other writing style to simple this? do it like:
```
send data  address(ByteArrary)
```

src/resources/espnow.cc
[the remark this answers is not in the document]
erikcorry: Yes, that's not good.  I'll fix that: https://github.com/toitlang/toit/pull/1155

lib/espnow.toit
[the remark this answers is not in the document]
floitsch: This is the trade-off with the class.

Another writing style is to put the address-argument in the next line, so it doesn't need the parenthesis:
```
send data
    Address address_bytes
```
Generally preferred is to have a local a line earlier, as it costs the same amount of lines:
```
address := Address address_bytes
send data address
```

As mentioned above: this is a trade-off.
If you think that users will have lots of byte-array addresses, then adding the class would not be useful.
If the addresses come from the system (for example with a "scan" like in Bluetooth), then the users would already get the object.
If they are constants, then users would probably write:
```
PEER ::= Address #[0x12, 0x34, 0x56, ... ]
```

I can live with both approaches. Note, however, that I would definitely prefer the address to be a named parameter if it is still a byte-array (to make sure it's not accidentally switched with the data).

floitsch:
If possible, please don't force-push when doing updates to PRs, as this breaks Github's "see changes since last review".
We are going to squash the commits at the end anyway, so there is no harm in having additional commits during the review.

lib/esp32/espnow.toit
```diff
@@ -0,0 +1,75 @@
...
+  if address.size != 6:
+      throw "ESP-Now MAC address length must be 6 bytes"
+
+check_key_ key/ByteArray:
+  if key.size != 16:
+      throw "ESP-Now key length must be 16 bytes"
+
+check_channel_ channel/int:
+  if channel > 14 or channel < 0:
+      throw "ESP-Now channel range must be 0~14"
```
floitsch, line 20:
```suggestion
  if not 0 <= channel <= 14:
      throw "ESP-Now channel range must be 0-14"
```

lib/esp32/espnow.toit
```diff
@@ -0,0 +1,75 @@
...
+
+class Address:
+  address/ByteArray
```
floitsch, line 23: Maybe call this "mac" ?
This way, we don't have the confusion of which "address" we are talking about.
When using "mac" we mean the byte array, when saying "address" we mean the Address class.

Clearly this is a made-up distinction, but it should help when documenting.

lib/esp32/espnow.toit
```diff
@@ -0,0 +1,75 @@
...
+
+  constructor .address:
+    check_mac_ address
```
floitsch, line 26: The `check_mac_` should be inlined here, and all other methods should use the `Address` class.

lib/esp32/espnow.toit
```diff
@@ -0,0 +1,75 @@
...
+class Datagram:
+  address/Address
+  data/ByteArray
+
+  constructor .address .data:
+
+class Service:
+  _group ::= ?
+
+  constructor --mode/int=STATION --pmk/ByteArray?=null:
```
floitsch, line 40: Any reason you didn't switch to a named constructor, and changed the name of the key parameter?

lib/esp32/espnow.toit
```diff
@@ -0,0 +1,75 @@
...
+    if not pmk:
+      pmk = #[]
+    else:
+      check_key_ pmk
+    _group = espnow_init_ mode pmk
+
+  send data/ByteArray --address/ByteArray --wait/bool=true -> int:
```
floitsch, line 47: This should now take an Address:
```suggestion
  send data/ByteArray --address/Address --wait/bool=true -> int:
```

lib/esp32/espnow.toit
```diff
@@ -0,0 +1,75 @@
...
+    check_mac_ address
+    return espnow_send_ address data wait
+
+  receive -> Datagram?:
+    array := espnow_receive_ (Array_ 2)
+    if not array:
+      return null
```
floitsch, line 54:
```suggestion
    if not array: return null
```

floitsch (commented):
Still a few nits and such on the Toit code, but getting there.

lib/espnow.toit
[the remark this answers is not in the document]
donghengqaz: OK, I get it and I will add class of `Key` like this.

lib/esp32/espnow.toit
```diff
@@ -0,0 +1,75 @@
...
+    return Datagram (Address array[0]) array[1]
+
+  add_peer address/ByteArray --channel/int --lmk/ByteArray?=null --encrypted/bool=(lmk != null) -> bool:
```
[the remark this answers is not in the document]
donghengqaz: Yes, keep using `key` maybe more readable than `lmk`.

lib/esp32/espnow.toit
donghengqaz: No reason, mainly I don't catch why using named constructor is better, but I can try this.

donghengqaz:
@floitsch @erikcorry I have modified by your advice and some toit rule, PTAL.

lib/esp32/espnow.toit
```diff
@@ -0,0 +1,79 @@
...
+  constructor .address .data:
+
+class Service:
+  _group ::= ?
+
+  constructor --mode/int=STATION:
+    _group = espnow_init_ mode #[]
+  
+  constructor.with_key --mode/int=STATION --key/Key:
+    _group = espnow_init_ mode key.data
```
floitsch, line 51: I think this is more convenient (and rename the `STATION` to `STATION_` in line 5).
```suggestion
  constructor.station --key/Key?:
    key_data = key ? key.data : #[]
    _group = espnow_init_ STATION_ key_data
```

Basically, the call would then be:
```
service := espnow.Service.station
// or
service := espnow.Service.station --key=(espnow.Key #[1, 2, 3])
```

If the `station` mode is used significantly more often, then we might not need a named constructor for the station mode:
```suggestion
  constructor --key/Key?:
    key_data = key ? key.data : #[]
    _group = espnow_init_ STATION_ key_data
```

Once we support SoftAP, we can then add a named constructor:
```
  constructor.soft_ap ...:
```

Alternatively, if you think the mode flag is just better, we can add the mode flag at that point.

lib/esp32/espnow.toit
```diff
@@ -0,0 +1,79 @@
...
+
+  send data/ByteArray --address/Address --wait/bool=true -> int:
+    return espnow_send_ address.mac data wait
```
floitsch, line 54: Looking through the `send` code, it seems like the function never returns anything else than 0.
-> I would change this to:
```suggestion
  send data/ByteArray --address/Address --wait/bool=true -> none:
    espnow_send_ address.mac data wait
```
If there is an error, it would yield an exception.

src/resources/espnow.cc
```diff
@@ -0,0 +1,277 @@
...
+  if (ret != pdTRUE) {
+    free_datagram(datagram);
+    // ESP_LOGE("ESPNow", "Failed to send datagram to rx_queue");
+  }
+}
+
+PRIMITIVE(init) {
+  ARGS(int, mode, Blob, pmk);
+
+  ByteArray* proxy = process->object_heap()->allocate_proxy();
```
floitsch, line 156: Add a check that the key is of the right size. We do that in Toit before, but it's still good practice to ensure your invariants in C++ again.

src/resources/espnow.cc
```diff
@@ -0,0 +1,277 @@
...
+struct DataGram {
+  bool used;
+  int len;
+  uint8_t mac[6];
+  uint8_t buffer[ESPNOW_RX_DATAGRAM_LEN_MAX];
+};
+
+const int kInvalidESPNow = -1;
+
+// Only allow one instance of provisioning running.
```
floitsch, line 49:
```suggestion
// Only allow one instance to use espnow.
```

lib/esp32/espnow.toit
```diff
@@ -0,0 +1,79 @@
+// Copyright (C) 2022 Toitware ApS. All rights reserved.
+// Use of this source code is governed by an MIT-style license that can be
+// found in the lib/LICENSE file.
+
+STATION ::= 0 // Wi-Fi station mode
+SOFTAP  ::= 1 // Not support yet
+
+check_mac_ address/ByteArray:
```
floitsch, line 8: Inline this function.

lib/esp32/espnow.toit
```diff
@@ -0,0 +1,79 @@
...
+  if address.size != 6:
+      throw "ESP-Now MAC address length must be 6 bytes"
+
+check_key_ key/ByteArray:
```
floitsch, line 12: Inline this function.

lib/esp32/espnow.toit
```diff
@@ -0,0 +1,79 @@
...
+  if key.size != 16:
+      throw "ESP-Now key length must be 16 bytes"
+
+check_channel_ channel/int:
```
floitsch, line 16: Remove this unused function.
(Probably add this check to `add_peer` ?)

lib/esp32/espnow.toit
```diff
@@ -0,0 +1,79 @@
...
+espnow_send_ mac data wait:
+  #primitive.espnow.send
+
+espnow_receive_ output:
+  #primitive.espnow.receive
+
+espnow_add_peer_ mac channel key:
+  #primitive.espnow.add_peer
+
+BROADCAST_ADDRESS ::= Address #[0xff, 0xff, 0xff, 0xff, 0xff, 0xff]
```
floitsch, line 79: Nit: I would move this constant closer to the top of the file.

floitsch (commented):
LGTM for the Toit part (with a few more comments).

I will let Erik do the review of the C++ code.

lib/esp32/espnow.toit
donghengqaz: Yes, using this method will look more direct.

lib/esp32/espnow.toit
donghengqaz: In the send primitive side, the function `send` may return error code, like `ETIMEDOUT` and `EIO`.

src/resources/espnow.cc
donghengqaz: Of source.

lib/esp32/espnow.toit
donghengqaz: Moving this to `add_peer` is better, I would like to check this in toit side.

lib/esp32/espnow.toit
floitsch: The `Primitive::os_error(ETIMEDOUT, process)` is actually nicer than that: it builds an exception with the error message. As such, the primitive will never send anything else than 0.

In fact, it should then change to return the null-object instead, but fundamentally that's not as important.

floitsch (commented):
Toit code now LGTM! (once the `send` function has return type `none`).

I think the API of the library is now nice. Will be happy to use it.

src/resources/espnow.cc
```diff
@@ -0,0 +1,279 @@
...
+
+class ESPNowResourceGroup : public ResourceGroup {
+ public:
+  TAG(ESPNowResourceGroup);
+
+  ESPNowResourceGroup(Process* process);
+  ~ESPNowResourceGroup();
+};
+
+ESPNowResourceGroup::ESPNowResourceGroup(Process* process)
```
erikcorry, line 68: The ResourceGroup is not created at boot time, it is created when the Toit program starts using ESPnow.  This means we have to guard against allocation failures.

It's probably better to have a trivial constructor in the ESPNowResourceGroup, then do the difficult work in the init primitive. That way, because you are in a primitive, you can return an error when allocations fail.

donghengqaz: Yes. Can the ESPNowResourceGroup's constructor function throw out an error when some errors trigger ?

donghengqaz: I know to implement according to discussion yesterday.

src/resources/espnow.cc
```diff
@@ -0,0 +1,307 @@
...
+static struct DataGram* rx_datagrams;
+
+class ESPNowResourceGroup : public ResourceGroup {
+ public:
+  TAG(ESPNowResourceGroup);
+
+  ESPNowResourceGroup(Process* process) : ResourceGroup(process) {}
+  ~ESPNowResourceGroup();
+
+  bool Init();
```
floitsch, line 67:
```suggestion
  bool init();
```

src/resources/espnow.cc
```diff
@@ -0,0 +1,307 @@
...
+  rx_queue = NULL;
+
+  vSemaphoreDelete(rx_datagrams_mutex);
+  rx_datagrams_mutex = NULL;
+
+  free(rx_datagrams);
+  rx_datagrams = NULL;
+}
+
+bool ESPNowResourceGroup::Init(void) {
```
floitsch, line 84:
```suggestion
bool ESPNowResourceGroup::init(void) {
```

src/resources/espnow.cc
```diff
@@ -0,0 +1,307 @@
...
+  int id = espnow_pool.any();
+  if (id == kInvalidESPNow) ALREADY_IN_USE;
+
+  ESPNowResourceGroup* group = _new ESPNowResourceGroup(process);
+  if (!group) {
+    espnow_pool.put(id);
+    MALLOC_FAILED;
+  }
+
+  if (!group->Init()) {
```
floitsch, line 193:
```suggestion
  if (!group->init()) {
```

floitsch (commented):
DBC. (drive-by-comment).

donghengqaz:
> DBC. (drive-by-comment).

The function name style of primitive class  is a little stranger...

floitsch (commented):
DBC.

erikcorry:
Please merge up to newest origin/master

donghengqaz:
> Please merge up to newest origin/master

It's done.

src/resources/espnow.cc
```diff
@@ -0,0 +1,307 @@
...
+
+  proxy->set_external_address(group);
+
+  return proxy;
+}
+
+PRIMITIVE(send) {
+  ARGS(Blob, mac, Blob, data, bool, wait);
+
+  // // Reset the value of semaphore(max value is 1) to 0, so no need to check the result.
```
floitsch, line 222:
```suggestion
  // Reset the value of semaphore(max value is 1) to 0, so no need to check the result.
```

src/resources/espnow.cc
```diff
@@ -0,0 +1,307 @@
...
+static void espnow_recv_cb(const uint8_t *mac_addr, const uint8_t *data, int data_len) {
+  if (data_len > ESPNOW_RX_DATAGRAM_LEN_MAX) {
+    ESP_LOGE("ESPNow", "Receive datagram length=%d is larger than max=%d", data_len, ESPNOW_RX_DATAGRAM_LEN_MAX);
+    return ;
+  }
+
+  struct DataGram* datagram = alloc_datagram();
+  if (!datagram) {
+    // ESP_LOGE("ESPNow", "Failed to malloc datagram");
+    return ;
```
floitsch, line 162:
```suggestion
    return;
```

floitsch (approved):
LGTM.

PR #1148, opened 2022-10-25
Parity

erikcorry:


erikcorry:
We have `is_nan` and `is_power_of_two`, so perhaps they should be called `is_even_parity` but it sounds a little strange to me so I went with `has_even_parity`.

tests/bit_count_test.toit
```diff
@@ -70,3 +71,14 @@ test_popcount:
...
+test_parity:
+  expect
+    0x34.has_odd_parity
+  expect
+    0x3c.has_even_parity
+
+  expect_equals 1
+    0x23.parity
+  expect_equals 0
+    0xff.parity
```
floitsch, line 84: Add the test for -1.

floitsch (approved):
LGTM.

PR #1150, opened 2022-10-25
A C++ style guide for Toit

erikcorry:


src/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Language features
+
+For performance reasons, we compile with `-fno-exceptions` and so C++ exception
+handling is not available.
+
+We also compile with `-fno-rtti`, though this might change at some point.
+
+We do not allow use of the C++ standard library on the device, for reasons of
+code size.  In particular this means no use of `std::string`.
```
floitsch, line 20: Not just. `std::string` uses `malloc` which is a more important reason not to use it.

src/README.md
```diff
@@ -0,0 +1,224 @@
...
+  int method_a();  // Yes.
+  int MethodB();   // No, use lower case.
+  int methodC();   // No, use underscores.
+};
+```
+
+Local variables and arguments can be short, but not too short.  Avoid needless abbreviations.
+For example, `address` is usually better than `addr`.  Single-letter loop variables like `i`
+and `j` are OK for very small loops, but it probably makes sense to give a better name for
+loops of 10 lines or more.
```
floitsch, line 168: It should be ok to use `i`, if the first thing I do is `auto x = foo[i]`.
```suggestion
and `j` are OK if their uses are in proximity to the `for` keyword.
```

src/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Readability
+
+Try to avoid unnamed bool arguments.
+
+```
+  my_function(42, true);    // What does this mean?
+
+  bool force = true;
+  my_function(42, force);  // More readable.
```
floitsch, line 180: ?
```suggestion
  my_function(42, force);  // More readable.

  bool force;
  my_function(42, force=false);  
```

src/README.md
```diff
@@ -0,0 +1,224 @@
...
+third party code that uses `new` you will hit a fatal assertion.  This
+is because such code generally cannot cope with memory exhaustion without
+crashing.
+
+Some uses of C++ closures will cause the compiler to generate
+calls to `new` that will crash on memory shortage.  Avoid capturing
+too many variables to avoid this.  Otherwise you will hit an assert.
+
+Always check the return value of `_new` for a null pointer.  In this
+case the constructor has not been called.
```
floitsch, line 211:
```suggestion
Always check the return value of `_new` for a null pointer.  In this
case the constructor has not been called. This is not required in the compiler, which can assume
that there is infinite memory.
```

floitsch (approved):
I'm sure we will add much more over time, but a good start.
LGTM.

PR #1154, opened 2022-10-26
Uart windows

mikkeldamsgaard:
Adding most functionality of the uart for windows.  (read/write/set control). Some of the write options are not enabled.

Incidentally this PR also adds a win64arm port. To my surprise, the compiler also uses mbedtls and I had to add some porting code to make win64arm compile. So the toit_porting lib is added to the compiler aswell.

clang14 was used to cross compile, binaries downloaded from: https://github.com/mstorsjo/llvm-mingw/releases
The ucrt version for mac was used during development and test.


mikkeldamsgaard:
No reason to spend time on reviewing this PR, as the async handling is below par.

PR #1155, opened 2022-10-27
Replace some asserts in primitives with actual checks

erikcorry:


src/primitive.cc
```diff
@@ -61,6 +61,7 @@ Object* Primitive::allocate_large_integer(int64 value, Process* process) {
 
 Object* Primitive::allocate_array(int length, Object* filler, Process* process) {
   ASSERT(length <= Array::max_length_in_process());
```
kasperl, line 63: Drop this assert now?

src/resources/x509.cc
```diff
@@ -101,7 +101,9 @@ PRIMITIVE(parse) {
     String* str = String::cast(input);
     data = reinterpret_cast<const uint8_t*>(str->as_cstr());
     length = str->length() + 1;
-    ASSERT(data[length - 1] == '\0');
+    // Should not happen since Toit strings are null terminated.
```
kasperl, line 104: I think this should remain an ASSERT if it shouldn't happen and the VM is supposed to guarantee it.

kasperl (approved):
Looks good to me if you avoid the update to third_party/esp-idf.

src/primitive.cc
erikcorry: Returning null is safe and will cause an OOM instead of a vulnerability, but I still would prefer the assert if this happens while I am developing - the check should be in the calling primitive where we can return an appropriate exception.

PR #1157, opened 2022-10-27
Add generic CRC algorithm

erikcorry:
Replaces the nibble-at-a-time CRC32 algorithm with a byte-at-a-time generic algorithm.
Replaces the bit-at-a-time CRC16 algorithm with a byte-at-a-time generic algorithm.

### DIFF
--- a/lib/crypto/crc.toit
+++ b/lib/crypto/crc.toit
@@ -13,7 +13,7 @@ Cyclic Redundancy Check (CRC).
 CRC is an error-detection code.
 
 See https://en.wikipedia.org/wiki/Cyclic_redundancy_check
-and https://crccalc.com/
+  and https://crccalc.com/
 */
 
 class Crc extends Checksum:
--- a/lib/crypto/crc.toit
+++ b/lib/crypto/crc.toit
@@ -1,6 +1,9 @@
-// Copyright (C) 2020 Toitware ApS. All rights reserved.
+// Copyright (C) 2022 Toitware ApS. All rights reserved.
 // Use of this source code is governed by an MIT-style license that can be
 // found in the lib/LICENSE file.
+
+import binary show LITTLE_ENDIAN
+import expect show *
 
 import .checksum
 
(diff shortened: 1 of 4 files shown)


lib/crypto/crc16.toit
```diff
@@ -17,40 +18,11 @@ The $data must be a string or byte array.
...
-  /**
-  See $super.
-
-  Returns the CRC16 checksum as a 2 element byte array in little-endian order.
-  */
-  get -> ByteArray:
-    checksum := sum_
-    return ByteArray 2: (checksum >> (8 * it)) & 0xff
+class Crc16 extends Crc:
+  constructor:
```
floitsch, line 27: I would make the polynomial a default value of a named parameter.

Maybe even deprecate a constructor without any polynomial.

Not completely sure how to do it, but the Modem Crc16 is not *that* common to make it the default.

lib/crypto/crc32.toit
```diff
@@ -14,26 +14,17 @@ The $data must be a string or byte array.
 Returns the checksum as a 4 element byte array in little-endian order.
 */
 crc32 data from/int=0 to/int=data.size -> ByteArray:
```
floitsch, line 16: I would make the initial-state and xor_result be optional parameters. (unless nobody uses any else).

For crc32, the function might be good enough, as there aren't many different implementations.

lib/crypto/crc32.toit
```diff
@@ -14,26 +14,17 @@ The $data must be a string or byte array.
...
-
-  /**
-  See $super.
-
-  Returns the CRC32 checksum as a 4 element byte array in little-endian order.
-  */
-  get -> ByteArray:
-    checksum := sum_ ^ 0xffffffff
-    return ByteArray 4: (checksum >> (8 * it)) & 0xff
+    super.little_endian 32
```
floitsch, line 27: At least for the constructor I would probably allow different polynomials, initial_state and/or xor_results.

floitsch (approved):
LGTM.

lib/crypto/crc.toit
```diff
@@ -1,43 +1,223 @@
...
+    (reversed) ordering for the polynomial.
+
+  # Example
+  ```
+  // The popular CRC-32 used in PNG.
+  crc := Crc.little_endian 32 --polynomial=0xEDB88320 --initial_state=0xffff_ffff --xor_result=0xffff_ffff
+  ```
+  */
+  constructor.little_endian .width/int --.polynomial/int --initial_state/int=0 --.xor_result/int=0:
+    if polynomial > 1 << width: throw "Polynomial and width don't match"
```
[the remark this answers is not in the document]
erikcorry: I think the () are noise here.

erikcorry:
I think I want to deprecate Crc16 completely and replace it with classes and methods that use the name from crccalc.com, which is CRC-16/XMODEM, which would be Crc16Xmodem.

For Crc32 that is the name used at crccalc.com, so I would keep it, but not add config options, since using those would make it a different algorithm that did not match the name any more.

But I'll do this in a new change.

lib/crypto/crc.toit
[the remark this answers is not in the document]
floitsch: I disagree. I misread it the first time.
The `>` and `<<` look too similar in this context.

PR #1158, opened 2022-10-28
Add support for mapping current firmware into memory

kasperl:


src/primitive_core.cc
```diff
@@ -2332,4 +2334,104 @@ PRIMITIVE(word_size) {
...
+  int size = Smi::cast(receiver->at(2))->value();
+  if (index < 0 || index >= size) OUT_OF_BOUNDS;
+
+  Blob input;
+  if (!receiver->at(0)->byte_content(process->program(), &input, STRINGS_OR_BYTE_ARRAYS)) {
+    WRONG_TYPE;
+  }
+
+  // Firmware is potentially mapped into memory that only allow word
+  // access. We read the full word before masking and shifting.
```
erikcorry, line 2408: Add a comment that mentions endianism so we have something to search for if we ever port to a big-endian platform.

src/primitive_core.cc
```diff
@@ -2332,4 +2334,104 @@ PRIMITIVE(word_size) {
...
+  index += offset;
+  const uint32* words = reinterpret_cast<const uint32*>(input.address());
+  uint32 shifted = words[index >> 2] >> ((index & 3) << 3);
```
erikcorry, line 2411: This is not always enough, and a 'helpful' compiler may generate a byte operation here when optimizing.

In cmpctmalloc I used this to avoid that:

```
// Prevents some optimizations by doing a null operation on a register variable
// that the compiler cannot see.  This is used to prevent the compiler
// converting 32 bit load instructions into 16 bit load instructions that do
// not work on IRAM on the ESP32.
#define OBFUSCATE(variable)    \
  __asm__ (""                  \
          :"=r"(variable)      \
          :"0"(variable)       \
          :                    \
  )                            \
```

lib/system/firmware.toit
```diff
@@ -101,19 +114,93 @@ class FirmwareWriter extends ServiceResourceProxy:
...
+    else:
+      block_to = block_from
+    copy_range_ block_to to into cursor
+
+  copy_range_ from/int to/int into/ByteArray index/int -> int:
+    while from < to: into[index++] = this[from++]
+    return index
+
+  copy_block_ from/int to/int into/ByteArray index/int -> int:
+    #primitive.core.firmware_mapping_copy
```
erikcorry, line 200: You can just call ByteArray.replace instead of doing this, which means you don't need a new primitive.

erikcorry (approved):

lib/system/firmware.toit
kasperl: Simplified ByteArray.replace instead.

src/primitive_core.cc
kasperl: Done.

src/primitive_core.cc
kasperl: Given that the index is a non-constant here, I don't think we're (as) likely to run into the issue.

PR #1159, opened 2022-10-28
Standard crc algorithms

erikcorry:


lib/crypto/crc.toit
```diff
@@ -68,12 +68,17 @@ class Crc extends Checksum:
...
-  constructor.little_endian width/int --normal_polynomial/int --initial_state/int=0 --xor_result/int=0:
-    polynomial := 0
+  constructor.little_endian .width/int --normal_polynomial/int --initial_state/int=0 --.xor_result/int=0:
+    poly := 0
     for i := 0; i < width; i++:
       if (normal_polynomial >> i) & 1 == 1:
-        polynomial |= 1 << (width - 1 - i)
-    return Crc.little_endian width --polynomial=polynomial --initial_state=initial_state --xor_result=xor_result
+        poly |= 1 << (width - 1 - i)
+    if poly > 1 << width: throw "Polynomial and width don't match"
```
floitsch, line 76: I'm still needing two attempts to read this correctly...
```suggestion
    if poly > (1 << width): throw "Polynomial and width don't match"
```

lib/crypto/crc.toit
```diff
@@ -210,8 +215,730 @@ class Crc extends Checksum:
...
+      binary.BIG_ENDIAN.put_uint result result.size 0 checksum
+    return result
+
+/**
+Computes the CRC-16/CCITT-FALSE checksum of the given $data.
+
+The $data must be a string or byte array.
+Returns the checksum as a 16-bit integer.
+*/
+crc_16_ccitt_false data from/int=0 to/int=data.size -> int:
```
floitsch, line 236: I would drop the `from` and `to` here and for all the others. We have slices for that.

If kept, they should be named arguments. `crc_16_ccitt_false data 3` wouldn't read nicely or convey what it does.

floitsch, line 236: I would prefer the name `crc16_ccitt_false` (same for all the others).

floitsch (approved):
LGTM.

PR #1162, opened 2022-10-29
Accelerate crc calculations

erikcorry:


lib/crypto/crc.toit
```diff
@@ -187,30 +187,37 @@ class Crc extends Checksum:
 
   /** See $super. */
   add data from/int to/int -> none:
-    sum := sum_
     if little_endian:
+      sum_ = calculcate_crc_little_endian_ sum_ 0 data from to table_
+    else:
+      sum_ = calculcate_crc_big_endian_ sum_ width data from to table_
+
+  static calculcate_crc_little_endian_ sum/int zero/int data from/int to/int table -> int:
```
floitsch, line 195: Add a comment, indicating that the primitive uses the `zero/width` to know which endianness is used.

floitsch (approved):
LGTM.

PR #1165, opened 2022-10-31
Fix 64 bit CRC calculations and remove --from from top level CRC functions

erikcorry:


lib/crypto/crc.toit
```diff
@@ -49,7 +49,7 @@ class Crc extends Checksum:
   ```
   */
   constructor.little_endian .width/int --.polynomial/int --initial_state/int=0 --.xor_result/int=0:
-    if polynomial > (1 << width): throw "Polynomial and width don't match"
+    if width < 64 and polynomial > (1 << width): throw "Polynomial and width don't match"
```
floitsch, line 52: Add check for width > 64.

floitsch (approved):
LGTM.

floitsch (approved):
Still LGTM.

PR #1166, opened 2022-10-31
Add protobuf library to external tests.

floitsch:


floitsch:
Dropping the PR.
The tests need some setup, and it's not worth changing the Toit repo for that.

PR #1170, opened 2022-11-01
Add CRC-64 from XZ format.

erikcorry:
This is mainly to make it clearer that the CRC package works for 64 bit CRCs too.


floitsch (approved):
LGTM.

PR #1173, opened 2022-11-02
Use a special type for stop_bits of uart.

erikcorry:
It's confusing that --stop_bits=2 runs just fine,
but sets the number of stop bits to 1.5.  To avoid this, make a special type.  Should be backwards
compatible except in the case where someone used
--stop_bits=1, which coincidentally worked fine.
However this was also the default, so there would
be no need to do this explicitly.


floitsch (approved):
LGTM.
An alternative would be to set the constants to:
```
static STOP_BITS_1 ::= 1
static STOP_BITS_1_5 ::= -1
static STOP_BITS_2 ::= 2
```

This way a `--stop_bits=1` or `--stop_bits=2` would work, and one would need to use the constant for 1.5.

PR #1179, opened 2022-11-02
Add arm-linux-gnueabi compilation target.

floitsch:


toolchains/arm-linux-gnueabi.cmake
```diff
@@ -0,0 +1,23 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# The license can be found in the file `LICENSE` in the top level
+# directory of this repository.
+
+set(ARM_TARGET "arm-linux-gnueabi")
+
+set(ARM_CPU_FLAGS "-marm -march=armv5te -mtune=arm926ej-s -msoft-float -mfloat-abi=soft")
```
kasperl, line 18: It feels a bit weird that arm-linux-gnueabi implies arvm5te and specific tuning.

Makefile
```diff
@@ -185,6 +185,32 @@ build/$(PI_CROSS_ARCH)/sysroot/usr: check-env-sysroot
 pi: pi-sysroot
 	$(MAKE) CROSS_ARCH=raspberry_pi SYSROOT="$(CURDIR)/build/$(PI_CROSS_ARCH)/sysroot" all-cross
 
+ARM_LINUX_GNUEABI_CROSS_ARCH := arm-linux-gnueabi
```
kasperl, line 188: I wonder if we should find a way to move some of these out of the main makefile.

kasperl (approved):
Looks good to me.

Makefile
floitsch: I will try to do that in a separate PR.

toolchains/arm-linux-gnueabi.cmake
floitsch: Agree.
That said, we had to make a similar decision for the raspberry pi.

I think users will need to adapt these flags anyway.
Open for suggestions, but would leave it as is for now.

PR #1180, opened 2022-11-03
Unify style for empty constructors

erikcorry:


src/ar.h
```diff
@@ -41,14 +41,14 @@ class File {
     , name_disposal_(name_disposal)
     , content_(content_arg)
     , content_disposal_(content_disposal)
-    , byte_size(byte_size_arg) { }
+    , byte_size(byte_size_arg) {}
```
kasperl, line 44: Have you looked for cases where there is a newline involved?

```
    , byte_size(byte_size_arg) {
}
```

Also, why is byte_size public and not const?

floitsch (approved):
LGTM.

PR #1181, opened 2022-11-03
Minor changes for the ESP32S2

floitsch:
The biggest change is in rtc_memory_esp32.cc. We now call esp_timer_impl_early_init for non-ESP32 devices.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,6 @@ tedi_test.go
 *.swp
 *.swo
 *.swn
-/toolchains/esp32s3/sdkconfig
-/toolchains/esp32s2/sdkconfig
-/toolchains/esp32c3/sdkconfig
-/toolchains/esp32/sdkconfig
+
+/toolchains/*/sdkconfig
+/toolchains/*/sdkconfig.old
--- a/.gitignore
+++ b/.gitignore
@@ -27,5 +27,5 @@ tedi_test.go
 *.swo
 *.swn
 
-/toolchains/*/sdkconfig
-/toolchains/*/sdkconfig.old
+/toolchains/esp32*/sdkconfig
+/toolchains/esp32*/sdkconfig.old
--- a/src/os_esp32.cc
+++ b/src/os_esp32.cc
@@ -200,7 +200,7 @@ class ConditionVariable {
 
     mutex_->unlock();
 
-#ifdef CONFIG_IDF_TARGET_ESP32C3
+#ifndef CONFIG_IDF_TARGET_ESP32
     uint32_t value = 0;
 #else
     uint32 value = 0;
--- a/src/os_esp32.cc
+++ b/src/os_esp32.cc
@@ -200,11 +200,7 @@ class ConditionVariable {
 
     mutex_->unlock();
 
-#ifndef CONFIG_IDF_TARGET_ESP32
     uint32_t value = 0;
-#else
-    uint32 value = 0;
-#endif
     bool success = xTaskNotifyWait(0x00, 0xffffffff, &value, ticks) == pdTRUE;
 
     mutex_->lock();
--- a/src/primitive_esp32.cc
+++ b/src/primitive_esp32.cc
@@ -305,7 +305,7 @@ PRIMITIVE(total_deep_sleep_time) {
 }
 
 PRIMITIVE(enable_external_wakeup) {
-#ifdef CONFIG_IDF_TARGET_ESP32
+#ifndef CONFIG_IDF_TARGET_ESP32C3
   ARGS(int64, pin_mask, bool, on_any_high);
   esp_err_t err = esp_sleep_enable_ext1_wakeup(pin_mask, on_any_high ? ESP_EXT1_WAKEUP_ANY_HIGH : ESP_EXT1_WAKEUP_ALL_LOW);
   if (err != ESP_OK) {
@@ -317,7 +317,7 @@ PRIMITIVE(enable_external_wakeup) {
 }
 
 PRIMITIVE(enable_touchpad_wakeup) {
-#ifdef CONFIG_IDF_TARGET_ESP32
+#ifndef CONFIG_IDF_TARGET_ESP32C3
   esp_err_t err = esp_sleep_enable_touchpad_wakeup();
   if (err != ESP_OK) {
     ESP_LOGE("Toit", "Failed: sleep_enable_touchpad_wakeup");
--- a/src/primitive_esp32.cc
+++ b/src/primitive_esp32.cc
@@ -305,7 +305,7 @@ PRIMITIVE(total_deep_sleep_time) {
 }
 
 PRIMITIVE(enable_external_wakeup) {
-#ifndef CONFIG_IDF_TARGET_ESP32C3
+#ifdef CONFIG_IDF_TARGET_ESP32
   ARGS(int64, pin_mask, bool, on_any_high);
   esp_err_t err = esp_sleep_enable_ext1_wakeup(pin_mask, on_any_high ? ESP_EXT1_WAKEUP_ANY_HIGH : ESP_EXT1_WAKEUP_ALL_LOW);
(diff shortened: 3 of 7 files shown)


src/os_esp32.cc
```diff
@@ -200,7 +200,7 @@ class ConditionVariable {
 
     mutex_->unlock();
 
-#ifdef CONFIG_IDF_TARGET_ESP32C3
+#ifndef CONFIG_IDF_TARGET_ESP32
```
kasperl, line 203: I don't understand this. Do we not have uint32 defined?

src/primitive_esp32.cc
```diff
@@ -305,7 +305,7 @@ PRIMITIVE(total_deep_sleep_time) {
 }
 
 PRIMITIVE(enable_external_wakeup) {
-#ifndef CONFIG_IDF_TARGET_ESP32C3
+#ifdef CONFIG_IDF_TARGET_ESP32
```
kasperl, line 308: This doesn't work on the S3?

src/primitive_esp32.cc
```diff
@@ -317,7 +317,7 @@ PRIMITIVE(enable_external_wakeup) {
 }
 
 PRIMITIVE(enable_touchpad_wakeup) {
-#ifndef CONFIG_IDF_TARGET_ESP32C3
+#ifdef CONFIG_IDF_TARGET_ESP32
```
kasperl, line 320: This doesn't work on the S3?

src/resources/ethernet_esp32.cc
```diff
@@ -15,7 +15,7 @@
 
 #include "../top.h"
 
-#if defined(TOIT_FREERTOS) && defined(CONFIG_TOIT_ENABLE_ETHERNET)
+#if defined(TOIT_FREERTOS) && defined(CONFIG_IDF_TARGET_ESP32) && defined(CONFIG_TOIT_ENABLE_ETHERNET)
```
kasperl, line 18: This doesn't work on the S3 and instead of failing from within init_esp32, it now is just a bunch of unimplemented primitives?

src/rtc_memory_esp32.cc
```diff
@@ -128,6 +134,11 @@ static void reset_rtc(const char* reason) {
   memset(&rtc, 0, sizeof(rtc));
   // We only clear RTC on boot, so this must be exactly 1.
   rtc.boot_count = 1;
+
+#ifndef CONFIG_IDF_TARGET_ESP32
```
kasperl, line 138: So we want this on the S3 and C3 too?

src/resources/touch_esp32.cc
```diff
@@ -274,13 +272,12 @@ PRIMITIVE(read) {
...
 
-#if CONFIG_IDF_TARGET_ESP32S3 || CONFIG_IDF_TARGET_ESP32S2
-  uint32_t val;
-  esp_err_t err = touch_pad_read_raw_data(pad, &val);
-#else
+#ifdef CONFIG_IDF_TARGET_ESP32
   uint16_t val;
-  esp_err_t err = touch_pad_read(pad, &val);
+#else
+  uint32_t val;
```
kasperl, line 278: C3 used to use uint16, no?

src/resources/touch_esp32.cc
```diff
@@ -274,13 +272,12 @@ PRIMITIVE(read) {
...
 #endif
+  esp_err_t err = touch_pad_read_raw_data(pad, &val);
```
kasperl, line 280: Is the change to always use raw okay?

src/os_esp32.cc
donghengqaz: I think using following judgement maybe more direct:
```
#if defined(CONFIG_IDF_TARGET_ESP32C3) || (CONFIG_IDF_TARGET_ESP32S2) || (CONFIG_IDF_TARGET_ESP32S3)
```
This can help make it easy to add more SoC support to toit. By a way, why not use `uint32_t` directly, this is defined in <stdint.h>.

src/resources/ethernet_esp32.cc
donghengqaz: Maybe we can add dependence(IDF_TARGET_ESP32) for `TOIT_ENABLE_ETHERNET` in Kconfig.

src/rtc_memory_esp32.cc
donghengqaz: Add this because we should initialize hardware system timer before calling trying to get time, you can see here: https://github.com/toitware/esp-idf/blob/67fa2950f6bed9cc8e2e89a8ffac1ed77f087214/components/esp_timer/Kconfig#L54.

src/resources/touch_esp32.cc
donghengqaz: Only ESP32/S2/S3 supports touch pad sensor, and data type is uint32_t.

src/primitive_esp32.cc
donghengqaz: This is supported by ESP32/S2/S3, but not C3.

src/primitive_esp32.cc
donghengqaz: This is supported by ESP32/S2/S3, but not C3.

src/os_esp32.cc
floitsch: Changed to `uint32_t` and removed the `#ifdef`.

src/primitive_esp32.cc
floitsch: reverted my change.

src/primitive_esp32.cc
floitsch: reverted my change.

src/resources/ethernet_esp32.cc
floitsch: reverted.

src/resources/touch_esp32.cc
floitsch: The C3 is already excluded by the big #ifdef at the top.

src/resources/touch_esp32.cc
floitsch: Yes. It just avoids the filter. Should be ok.

src/rtc_memory_esp32.cc
floitsch: Added a comment.

floitsch: @donghengqaz: is the call to esp_timer_impl_early_init documented somewhere?

src/rtc_memory_esp32.cc
```diff
@@ -128,6 +134,13 @@ static void reset_rtc(const char* reason) {
...
+  // Non ESP32 targets should use the SYSTIMER which needs a call to early init.
```
kasperl, line 139:
```suggestion
  // Non-ESP32 targets use the SYSTIMER which needs a call to early init.
```

src/resources/touch_esp32.cc
kasperl: SG!

.gitignore
```diff
@@ -26,7 +26,6 @@ tedi_test.go
 *.swp
 *.swo
 *.swn
-/toolchains/esp32s3/sdkconfig
-/toolchains/esp32s2/sdkconfig
-/toolchains/esp32c3/sdkconfig
-/toolchains/esp32/sdkconfig
+
+/toolchains/*/sdkconfig
```
kasperl, line 30:
```suggestion
/toolchains/esp32*/sdkconfig
```

.gitignore
```diff
@@ -26,7 +26,6 @@ tedi_test.go
...
+/toolchains/*/sdkconfig.old
```
kasperl, line 31:
```suggestion
/toolchains/esp32*/sdkconfig.old
```

kasperl (approved):
Looks good to me.

PR #1182, opened 2022-11-04
Implement FlashRegistry for Windows hosts

kasperl:


src/flash_registry_win.cc
```diff
@@ -22,40 +22,76 @@
...
+static const int ENCRYPTION_WRITE_SIZE = 16;
+
 const char* FlashRegistry::allocations_memory_ = null;
+static void* allocations_malloced = null;
 
 void FlashRegistry::set_up() {
+  ASSERT(allocations_malloced == null);
+  ASSERT(allocations_memory() == null);
+
+  allocations_malloced = malloc(ALLOCATION_SIZE + FLASH_PAGE_SIZE);
```
mikkeldamsgaard, line 35: Check malloc error?

kasperl: Yeah, maybe. In general, we do not necessarily check all mallocs on the host platforms, because we expect to crash there if we run out. A check here would have to be fatal.

src/flash_registry_win.cc
```diff
@@ -22,40 +22,76 @@
...
+  allocations_memory_ = Utils::round_up(unvoid_cast<char*>(allocations_malloced), FLASH_PAGE_SIZE);
```
floitsch, line 36: ?
```suggestion
  // Align the memory to FLASH_PAGE_SIZE.
  // Note that we allocated FLASH_PAGE_SIZE more than necessary, so we could do this.
  allocations_memory_ = Utils::round_up(unvoid_cast<char*>(allocations_malloced), FLASH_PAGE_SIZE);
```

src/flash_registry_win.cc
```diff
@@ -22,40 +22,76 @@
 
 namespace toit {
 
+static const int ALLOCATION_SIZE = 2 * MB;
+static const int ENCRYPTION_WRITE_SIZE = 16;
+
```
floitsch, line 27:
```suggestion

// An aligned (FLASH_BASED_SIZE) view into the allocations_malloced.
```

floitsch (approved):
LGTM.

PR #1183, opened 2022-11-06
C++ style: Remove newline in empty functions

erikcorry:


kasperl (approved):
Not sure I'm totally digging the look of these lines, but I really like the consistency.

PR #1188, opened 2022-11-11
Buffered reader cursor

erikcorry:


lib/reader.toit
```diff
@@ -124,6 +131,12 @@ class BufferedReader implements Reader:
   buffered -> int:
     return arrays_.size_in_bytes - first_array_position_
 
+  /**
+  The number of bytes that have been consumed from the BufferedReader.
+  */
+  cursor -> int:
```
kasperl, line 137: Maybe consumed is a better name given the description? I could see the name 'cursor' being used to delimit both the consumed part but also the buffered part.

Is it easy to add some tests of this functionality?

lib/expect.toit
```diff
@@ -117,11 +117,12 @@ expect_bytes_equal expected/ByteArray actual/ByteArray:
 /** Expects $throw_block to throw an object equal to the $expected value. */
 expect_throw expected [throw_block]:
   try:
-     throw_block.call
-  finally: | is_exception exception |
-    if not is_exception:
+    error := catch --trace=(: expected != it) --unwind=(: expected != it):
+      throw_block.call
+    if expected != error:
```
floitsch, line 122: This doesn't make sense to me.

If the caught exception is not equal to the `expected` we have a message that says that we got `null`?

I think this test was correctly testing whether `error` was null`. That said: given that an exception can be a boolean, making it an explicit `error == null` is clearly better.

Also: I wouldn't use the `--message` thingy anymore. Just `throw "Expected exception, but executed normally"` or so.

lib/expect.toit
```diff
@@ -117,11 +117,12 @@ expect_bytes_equal expected/ByteArray actual/ByteArray:
...
       expect false --message="Expected throw, got <null>"
-    if expected != exception.value:
+  finally: | is_exception exception |
```
floitsch, line 124: Now we would change the message from "Expected throw, got <null>" to "Expected throw with <...>, but was Expected throw, got <null>".

I think the `finally` clause should disappear.

floitsch (approved):
LGTM.

floitsch (changes requested):
The last commit introduced changes to `tls/session.toit`.
That was probably accidental.

floitsch (approved):
LGTM with the changes to session.toit reverted.

PR #1192, opened 2022-11-15
Add bignum lib and test.

donghengqaz:


erikcorry:
Hi

Thanks for this.  However I think we might want a different approach here.

We would like to keep long lived objects on the GCed Toit heap where the compacting GC can counteract fragmentation.  By having them on the malloc heap we increase the amount of data that is not movable and can cause memory issues.

Unfortunately I think the library you are using does not allow bignums to be backed by memory that it does not manage.  So we would have to copy the data into native bignums, do the operation, then copy it back.  I think the overhead of this is not too bad.  we have a fast malloc and memcpy is also fast.

Another idea I had is that there are currently a lot of primitives, and they do almost the same.  This is going to take more code than it has to, since a primitive has a minimum size overhead.  You don't necessarily see it all, because it's hidden away in the ARGS macro, but it causes code to be generated, which can cost us on footprint.  So I would prefer something like:

```
typedef int two_input_function_t(mbedtls_mpi *X, const mbedtls_mpi *A, const mbedtls_mpi *B);

static const two_input_function_t* TWO_INPUT_FUNCTIONS[] = {
    mbedtls_mpi_add_abs,
    mbedtls_mpi_sub_abs,
    mbedtls_mpi_add_mpi,
    mbedtls_mpi_sub_mpi,
    mbedtls_mpi_add_int,
    mbedtls_mpi_sub_int,
    mbedtls_mpi_mul_mpi,
    mbedtls_mpi_mul_int
};

static const int NUMBER_OF_TWO_INPUT_FUNCTIONS = sizeof(TWO_INPUT_FUNCTIONS) / sizeof(two_input_function_t*);

      // When i has been checked, call with:
      TWO_INPUT_FUNCTIONS[i](arg1, arg2, arg3)
```

donghengqaz:
> Hi
> 
> Thanks for this. However I think we might want a different approach here.
> 
> We would like to keep long lived objects on the GCed Toit heap where the compacting GC can counteract fragmentation. By having them on the malloc heap we increase the amount of data that is not movable and can cause memory issues.
> 
> Unfortunately I think the library you are using does not allow bignums to be backed by memory that it does not manage. So we would have to copy the data into native bignums, do the operation, then copy it back. I think the overhead of this is not too bad. we have a fast malloc and memcpy is also fast.
> 
> Another idea I had is that there are currently a lot of primitives, and they do almost the same. This is going to take more code than it has to, since a primitive has a minimum size overhead. You don't necessarily see it all, because it's hidden away in the ARGS macro, but it causes code to be generated, which can cost us on footprint. So I would prefer something like:
> 
> ```
> typedef int two_input_function_t(mbedtls_mpi *X, const mbedtls_mpi *A, const mbedtls_mpi *B);
> 
> static const two_input_function_t* TWO_INPUT_FUNCTIONS[] = {
>     mbedtls_mpi_add_abs,
>     mbedtls_mpi_sub_abs,
>     mbedtls_mpi_add_mpi,
>     mbedtls_mpi_sub_mpi,
>     mbedtls_mpi_mul_mpi
> };
> 
> static const int NUMBER_OF_TWO_INPUT_FUNCTIONS = sizeof(TWO_INPUT_FUNCTIONS) / sizeof(two_input_function_t*);
> 
>       // When i has been checked, call with:
>       TWO_INPUT_FUNCTIONS[i](arg1, arg2, arg3)
> ```

Hi @erikcorry  your advice is much meaning, I get your 2nd point: combine alike process to be one abstract process ? But I don't really understand your 1st point, do you mean like this ?
```
bignum = generat_by_mbedtls()
primitive_bignum = malloc(bignum.total_size);
primitive_bignum = bignum
```

erikcorry:
My suggestion is to have a Toit-side Bignum class:

```
BIGNUM_ADD_ ::= 0
BIGNUM_SUB_ ::= 1

class Bignum:
  sign /int
  limbs /ByteArray
  
  constructor .sign .limbs:
  
  operator + other -> Bignum:
    result := ?
    if other is int:
      result = bignum_operation_int_ BIGNUM_ADD_ sign limbs other
    else:
      other_bignum := other as Bignum
      result = bignum_operation_ BIGNUM_ADD_ sign limbs other_bignum.sign other_bignum.limbs
    return Bignum result[0] result[1]

bignum_operation_int_ operation/int sign/int limbs/ByteArray other/int -> List:
  #primitive.bignum.operation_int
  
bignum_ operation/int sign/int limbs/ByteArray other_sign/int other_limbs/ByteArray -> List:
  #primitive.bignum_operation
 ```
 
 Then in the primitive, you
 1) Allocate a two-element array for the result
 2) Create two `mbedtls_mpi` structs from the arguments
 3) Call the operation identified by the `operation` integer
 4) Put the sign of the result in the first element of the array
 5) Put a fresh ByteArray in the second element of the array with the limbs of the result
 6) Free all three mbedtls_mpi structures (you also need to do this if any allocation fails).
 7) Return the 2-element array

There are some issues around the conversion from a ByteArray to the MPI structure that uses a large int type.  It may be necessary to pad with zeros.

This is a bit heavy in terms of the allocation that takes place.  For mul, div, and mod I think the basic operation is so heavy that the allocation work is not important.  For simpler operations like add and sub we may want to make an operation that modifies the Bignum.  I don't think we have to be strict about them being immutable objects if performance becomes an issue.  But that would be a later change.

**Comment by donghengqaz** [comment]
Thanks for your detailed explanation, let me read and try this method.

**Comment by donghengqaz** [comment]
@erikcorry I refactor the code and put more calculation functions and transform functions in toit side.

File: lib/bignum.toit [hunk 1]
```diff
@@ -0,0 +1,86 @@
...
+import encoding.hex
+
+BIGNUM_ADD_ ::= 0
+BIGNUM_SUB_ ::= 1
+BIGNUM_MUL_ ::= 2
+BIGNUM_DIV_ ::= 3
+BIGNUM_MOD_ ::= 4
+
+class Bignum:
+  sign_/bool := ?
```
erikcorry, line 14: It's unclear what the mapping is between true-false and positive-negative.  I would suggest renaming this to `negative_`.

lib/bignum.toit
```diff
@@ -0,0 +1,86 @@
...
+  constructor .sign_/bool .limbs_/ByteArray:
+
+  constructor.from_string data/string:
+    if data[0] == '-':
+      sign_ = true
+      data = data[1..]
+    else:
+      sign_ = false
+    
+    if data.size & 0x1 !=0:
```
erikcorry, line 26:
```suggestion
    if data.size & 0x1 != 0:
```

src/bignum.cc
```diff
@@ -0,0 +1,165 @@
...
+
+PRIMITIVE(operator) {
+  ARGS(int, operator_id, bool, a_sign, Blob, a_limbs, bool, b_sign, Blob, b_limbs);
+
+  int ret;
+  mbedtls_mpi a_mpi;
+  mbedtls_mpi b_mpi;
+  mbedtls_mpi x_mpi;
+
+  if (operator_id >= MPI_BASIC_FUNCS_NUM) INVALID_ARGUMENT;
```
erikcorry, line 60: You must check for an id < 0 too.

lib/bignum.toit
```diff
@@ -0,0 +1,86 @@
...
+      data = "0" + data
+    
+    limbs_ = hex.decode data
+
+  int_to_arry_ i/int -> ByteArray:
+    return #[(i >>  0) & 0xff, (i >>  8) & 0xff,
+             (i >> 16) & 0xff, (i >> 24) & 0xff]
```
erikcorry, line 33: Should be called `array` or `byte_array`, not `arry`.

This fails for ints that are outside the 32 bit range.  Toit has 64 bit signed integers.

Also it uses little-endian order, but it appears that MbedTLS uses big-endian order for import and export.  So it's unclear to me how this works.

I suggest `import binary show BIG_ENDIAN` and then:

```suggestion
  int_to_arry_ i/int -> ByteArray:
    result := ByteArray 8
    BIG_ENDIAN.put_int64 result 0 i
    return result
```

lib/bignum.toit
```diff
@@ -0,0 +1,86 @@
...
+BIGNUM_SUB_ ::= 1
+BIGNUM_MUL_ ::= 2
+BIGNUM_DIV_ ::= 3
+BIGNUM_MOD_ ::= 4
+
+class Bignum:
+  sign_/bool := ?
+  limbs_/ByteArray := ?
+
+  constructor .sign_/bool .limbs_/ByteArray:
```
erikcorry, line 17: Perhaps this constructor should be named and private, and `trans_other_` should be the unnamed constructor.

lib/bignum.toit
```diff
@@ -0,0 +1,86 @@
...
+    if other is int:
+      other = Bignum
+          other < 0
+          int_to_arry_ other.abs
+    else if other is not Bignum:
+      throw "$(other.type) is not supported"
+    
+    return other
+
+  basic_operator_ operation/int sign/bool limbs/ByteArray other/any:
```
erikcorry, line 45: Since this method is not static it doesn't need sign and limbs as arguments, it can get them from `this`.

lib/bignum.toit
```diff
@@ -0,0 +1,86 @@
...
+  operator == other -> bool:
+    other = trans_other_ other
+    
+    if sign_ != other.sign_:
+      return false
+
+    return limbs_ == other.limbs_
+  
+  stringify -> string:
+    s := sign_ ? "-" : ""
```
erikcorry, line 74:
```suggestion
    s := sign_ ? "-0x" : "0x" 
```

src/bignum.cc
```diff
@@ -0,0 +1,165 @@
...
+#include "process.h"
+#include "resource.h"
+#include "resource_pool.h"
+#include "vm.h"
+
+#include "mbedtls/bignum.h"
+
+namespace toit {
+
+typedef int (*mpi_basic_func_t)(mbedtls_mpi *X, const mbedtls_mpi *A, const mbedtls_mpi *B);
```
erikcorry, line 27: Please use lower case for argument names throughout.

Asterisk is placed with the type, not with the variable throughout.

src/bignum.cc
```diff
@@ -0,0 +1,165 @@
...
+    mbedtls_mpi_div_mpi_no_r,
+    mbedtls_mpi_mod_mpi,
+};
+static const int MPI_BASIC_FUNCS_NUM = sizeof(MPI_BASIC_FUNCS) / sizeof(MPI_BASIC_FUNCS[0]);
+
+static int mbedtls_mpi_div_mpi_no_r(mbedtls_mpi *X,
+                                    const mbedtls_mpi *A,
+                                    const mbedtls_mpi *B)
+{
+  return mbedtls_mpi_div_mpi(X, NULL, A, B);
```
erikcorry, line 46: We use `null` instead of `NULL` in our code.

src/bignum.cc
```diff
@@ -0,0 +1,165 @@
...
+  ByteArray* limbs = process->allocate_byte_array(n);
+  if (limbs == null) {
+    mbedtls_mpi_free(&x_mpi);
+    ALLOCATION_FAILED;
+  }
+
+  bool sign = x_mpi.s == -1 ? true : false;
+  ret = mbedtls_mpi_write_binary(&x_mpi, ByteArray::Bytes(limbs).address(), n);
+  mbedtls_mpi_free(&x_mpi);
+  if (ret) INVALID_ARGUMENT;
```
erikcorry, line 96: You should return division by zero here when applicable.  See PRIMITIVE(smi_divide).

erikcorry (approved):

lib/bignum.toit
donghengqaz: Although MbedTLS use big-endian format to store data in MPI buffer, but this is raw data which read by MPI APIs, so we should store it as little-endian.

lib/bignum.toit
donghengqaz: Perhaps the `string` one should be unnamed constructor, because it use by user and SSL certification directly.

lib/bignum.toit
```diff
@@ -0,0 +1,87 @@
...
+
+  constructor.from_string data/string:
+    if data[0] == '-':
+      negative_ = true
+      data = data[1..]
+    else:
+      negative_ = false
+    
+    if data.size & 0x1 != 0:
+      data = "0" + data
```
erikcorry, line 28: I fixed this today so you can remove this code now.

lib/bignum.toit
```diff
@@ -0,0 +1,87 @@
...
+BIGNUM_DIV_ ::= 3
+BIGNUM_MOD_ ::= 4
+
+class Bignum:
+  negative_/bool := ?
+  limbs_/ByteArray := ?
+
+  constructor .negative_/bool .limbs_/ByteArray:
+
+  constructor.from_string data/string:
```
erikcorry, line 20: I agree, let's make this the unnamed constructor.

lib/bignum.toit
```diff
@@ -0,0 +1,87 @@
...
+      data = data[1..]
+    else:
+      negative_ = false
+    
+    if data.size & 0x1 != 0:
+      data = "0" + data
+    
+    limbs_ = hex.decode data
+
+  int_to_arry_ i/int -> ByteArray:
```
erikcorry, line 32: This still has a spelling mistake and should be static.

lib/bignum.toit
```diff
@@ -0,0 +1,87 @@
...
+    result := ByteArray 8
+    BIG_ENDIAN.put_int64 result 0 i
+    return result
+
+  trans_other_ other/any -> Bignum:
+    if other is int:
+      other = Bignum
+          other < 0
+          int_to_arry_ other.abs
```
erikcorry, line 41: The code for negative int is untested and does not work.

lib/bignum.toit
```diff
@@ -0,0 +1,87 @@
...
+    else if other is not Bignum:
+      throw "$(other.type) is not supported"
```
erikcorry, line 43:
```suggestion
      throw "WRONG_OBJECT_TYPE"
```

lib/bignum.toit
donghengqaz: I add negative test case, and main calculation is well, the `mod` and `exp_mod` don't support `dividend` is negative.

erikcorry:
Thanks!

donghengqaz:
@erikcorry Thanks for your help and direction.

PR #1193, opened 2022-11-15
Improve documentation of deque.

floitsch:
We want searches for "stack" or "queue" to find the deque class.

### DIFF
--- a/lib/core/collections.toit
+++ b/lib/core/collections.toit
@@ -2941,8 +2941,13 @@ class IdentityMap extends Map:
     return identical key key_or_probe
 
 /**
-A collection where you can add to the end, and remove items from
-  either end efficiently.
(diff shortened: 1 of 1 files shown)


lib/core/collections.toit
```diff
@@ -2941,8 +2941,13 @@ class IdentityMap extends Map:
     return identical key key_or_probe
 
 /**
-A collection where you can add to the end, and remove items from
-  either end efficiently.
+A double-ended queue.
+
+Supports adding and removing at the beginning and the end. Both operations
```
erikcorry, line 2946: You can only add at one end at the moment though.

floitsch: done.

erikcorry (approved):

PR #1194, opened 2022-11-16
Fix Set bug that would appear if the hash_do intrinsic always failed

erikcorry:


lib/core/collections.toit
```diff
@@ -2147,8 +2147,8 @@ abstract class HashedInsertionOrderedCollection_:
       // is very rare because the intrinsic will generally return a
       // progress-indicating integer rather than failing.
       return reversed
```
kasperl, line 2149: Maybe check for backing_ first?

kasperl (approved):
Looks good to me.

PR #1195, opened 2022-11-17
Allow an odd number of hex digits in encoding.hex.decode

erikcorry:


lib/encoding/hex.toit
```diff
@@ -39,22 +39,22 @@ create_decoding_map_ -> ByteArray:
 
 decode str/string -> ByteArray:
```
floitsch, line 40: Add toitdoc with examples.

floitsch (approved):
LGTM.

PR #1199, opened 2022-11-20
Ble darwin fixes

mikkeldamsgaard:
Two things. 

1) It seems that when scanning, remote devices can have larger advertising data and the size check in AdvertisementData needs to be disabled.

2) When not nicely closing BLE resources on darwin, a segfault is thrown on exit. Updated the code to be aware if it is tearing down, to exit gracefully.


mikkeldamsgaard:
Included in more recent PR

PR #1200, opened 2022-11-22
Add AES functions for GCM mode, used by TLS.

erikcorry:


src/primitive_crypto.cc
```diff
@@ -118,6 +120,206 @@ PRIMITIVE(siphash_get) {
...
+      GcmContext::BLOCK_SIZE - context->buffered_bytes(),
+      data_length);
+  memcpy(context->buffered_data() + context->buffered_bytes(), data_address, to_copy);
+  data_address += to_copy;
+  data_length -= to_copy;
+  const int buffered = context->buffered_bytes() + to_copy;
+  if (buffered < GcmContext::BLOCK_SIZE && remains != 0) {
+    // Success.  We copied all the data into the internal buffer, and so the
+    // output byte array is inevitably big enough.
+    // Update the context and the number of bytes of new output.
```
floitsch, line 224:
```suggestion
    // Success.  We copied all the data into the internal buffer.
    // Since we haven't filled up a full block yet, we can't run the crypt function yet
    // Update the context and let the caller know that we successfully handled the data.
```

src/primitive_crypto.cc
```diff
@@ -118,6 +120,206 @@ PRIMITIVE(siphash_get) {
...
+
+  // Some data is to be encrypted/decrypted.
+  const int to_process_after_internal_buffer = remains == 0
+      ? data_length
+      : Utils::round_down(data_length, GcmContext::BLOCK_SIZE);
+  if (buffered + to_process_after_internal_buffer > result.length()) {
+    // Output byte array not big enough.  At this point we have not yet
+    // modified the context.  Return null to indicate the problem.
+    return process->program()->null_object();
+  }
```
floitsch, line 238: This check can be done before doing any memcpy.

src/primitive_crypto.cc
```diff
@@ -118,6 +120,206 @@ PRIMITIVE(siphash_get) {
...
+PRIMITIVE(gcm_add) {
+  ARGS(GcmContext, context, Blob, data, MutableBlob, result);
+  const uint8* data_address = data.address();
+  int data_length = data.length();
+  if (data_length == 0) return Smi::from(0);
+  int remains = context->remaining_length_in_current_message();
+  if (remains < data_length) OUT_OF_BOUNDS;
+  remains -= data_length;
+
+  // Start by copying into the temporary buffer in the context.
```
floitsch, line 213: I think I would structure this differently:
1. the old and new data are smaller than a block.
  -> add the new data to the buffered. and return
2. compute the output size (a multiple of the block size). At this stage I would ignore whether this is the last block or not.  If it is the last block, then we might buffer something, which we should then handle in the `finish`/`verify` function.
3. if the result buffer isn't big enough -> return null
4. if buffer is not empty, fill it up and encrypt it.  (There is a good chance the chunks that come in here are aligned by 16 bytes)
5. encrypt the rest that can be handled this time.


Change `finish`/`verify` to crypt the last buffered chunk (if there is any).

src/primitive_crypto.cc
erikcorry: I prefer to update data_address and data_length when I do the memcpy, and this check uses the updated version of data_length, so I prefer it this way.

tests/crypto_test.toit
```diff
@@ -178,6 +181,19 @@ hamming_test:
...
+  initialization_vector := ByteArray 12: random 256
+
+  encrypted := (AesGcm.encryptor key initialization_vector).encrypt DREAM
+
+  expect_equals DREAM.size + 16 encrypted.size
+  expect_not_equals DREAM[..DREAM.size].to_byte_array DREAM
+
+  round_trip := (AesGcm.decryptor key initialization_vector).decrypt encrypted
+
+  expect_equals DREAM.to_byte_array round_trip
```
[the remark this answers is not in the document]
erikcorry: This is tested in the vector test.

src/primitive_crypto.cc
erikcorry: Not knowing the size up front means that PRIMITIVE(gcm_add) has been rewritten.  I think it's clearer now.

The encrypt method on an instance of the class follows the other Aes classes. :-/

lib/crypto/aes.toit
```diff
@@ -269,27 +276,34 @@ class AesGcm:
...
+        round_up data.size BLOCK_SIZE_
+    number_of_bytes /int := gcm_add_ gcm_aes_ data result
+    if not buffer_ or data.size > buffer_.size: buffer_ = data
+    return result[..number_of_bytes]
 
   /**
   Finishes encrypting.
   Can be called after $start and $add.
-  Returns the 16 byte verification tag.
+  Returns a two-element list with the last encrypted bytes and the verification
```
floitsch, line 296: Would it make sense to just return the ByteArray, and stating that the last 16 bytes are the tag?
If they are always concatenated afterwards anyway, we might as well keep them together.

src/primitive_crypto.cc
```diff
@@ -169,102 +222,102 @@ PRIMITIVE(gcm_close) {
...
-      context->gcm_context(),
-      buffered,
-      context->buffered_data(),
-      result_address);
+  int buffered = context->number_of_buffered_bytes();
+  // We cache buffered above because the next line changes the result of
+  // context->number_of_buffered_bytes().
+  context->increment_length(in.length());
+
+  if (buffered != 0) {
```
floitsch, line 285:
```suggestion
  if (buffered != 0) {
    // We have data buffered.
    // Fill the block and crypt it separately.
```

floitsch (approved):
Still LGTM.
I think the 'add' is now easier to understand.

PR #1202, opened 2022-11-22
Ble extra params

mikkeldamsgaard:
Added support for manipulating and reading advertising flags on ESP32. (Ignored on Darwin)

Darwin's lowlevel scan returns data that can not be length checked, so disabling length check when reading advertised data.

Fix a bug in darwin that caused a segfault when exiting the vm without properly closing ble resources.

### DIFF
--- a/lib/ble.toit
+++ b/lib/ble.toit
@@ -115,13 +115,13 @@ class AdvertisementData:
   */
   connectable/bool
 
-  constructor --.name=null --.service_classes=[] --.manufacturer_data=#[] --.connectable=false:
+  constructor --.name=null --.service_classes=[] --.manufacturer_data=#[] --.connectable=false --check_size=true:
     size := 0
     if name: size += 2 + name.size
     service_classes.do: | uuid/BleUuid |
       size += 2 + uuid.to_byte_array.size
     if not manufacturer_data.is_empty: size += 2 + manufacturer_data.size
-    if size > 31: throw "PACKET_SIZE_EXCEEDED"
+    if size > 31 and check_size: throw "PACKET_SIZE_EXCEEDED"
 
 /**
 A remote device discovered by a scanning.
@@ -607,6 +607,7 @@ class Central extends Resource_:
             --service_classes=service_classes
             --manufacturer_data=(next[4]?next[4]:#[])
             --connectable=next[5]
+            --check_size=false
         block.call discovery
     finally:
       ble_scan_stop_ resource_
--- a/lib/ble.toit
+++ b/lib/ble.toit
@@ -82,11 +82,15 @@ Conceptually, attributes are on the server, and can be accessed (read and/or wri
 interface Attribute:
   uuid -> BleUuid
 
-BLE_CONNECT_MODE_NONE          ::= 0
-BLE_CONNECT_MODE_DIRECTIONAL   ::= 1
-BLE_CONNECT_MODE_UNDIRECTIONAL ::= 2
+BLE_CONNECT_MODE_NONE                  ::= 0
+BLE_CONNECT_MODE_DIRECTIONAL           ::= 1
+BLE_CONNECT_MODE_UNDIRECTIONAL         ::= 2
 
-BLE_DEFAULT_PREFERRED_MTU_     ::= 23
+BLE_ADVERTISE_FLAGS_LIMITIED_DISCOVERY ::= 0x01
+BLE_ADVERTISE_FLAGS_GENERAL_DISCOVERY  ::= 0x02
+BLE_ADVERTISE_FLAGS_BREDR_UNSUPPORTED  ::= 0x04
+
+BLE_DEFAULT_PREFERRED_MTU_             ::= 23
 
 /**
 Advertisement data as either sent by advertising or received through scanning.
@@ -115,7 +119,14 @@ class AdvertisementData:
   */
   connectable/bool
 
-  constructor --.name=null --.service_classes=[] --.manufacturer_data=#[] --.connectable=false --check_size=true:
+  /**
+  Advertise flags. This must be a bitwise or of the BLE_ADVERTISE_FLAG_* constants
+    (see $BLE_ADVERTISE_FLAGS_GENERAL_DISCOVERY and similar).
+  */
+  flags/int
+
+  constructor --.name=null --.service_classes=[] --.manufacturer_data=#[]
+              --.connectable=false --.flags=0 --check_size=true:
     size := 0
     if name: size += 2 + name.size
     service_classes.do: | uuid/BleUuid |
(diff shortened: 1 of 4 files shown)


lib/ble.toit
```diff
@@ -82,11 +82,15 @@ Conceptually, attributes are on the server, and can be accessed (read and/or wri
...
 
-BLE_CONNECT_MODE_NONE          ::= 0
-BLE_CONNECT_MODE_DIRECTIONAL   ::= 1
-BLE_CONNECT_MODE_UNDIRECTIONAL ::= 2
+BLE_CONNECT_MODE_NONE                  ::= 0
+BLE_CONNECT_MODE_DIRECTIONAL           ::= 1
+BLE_CONNECT_MODE_UNDIRECTIONAL         ::= 2
 
-BLE_DEFAULT_PREFERRED_MTU_     ::= 23
+BLE_ADVERTISE_FLAGS_LIMITIED_DISCOVERY ::= 0x01
```
kasperl, line 89:
```suggestion
BLE_ADVERTISE_FLAGS_LIMITED_DISCOVERY ::= 0x01
```

lib/ble.toit
```diff
@@ -115,13 +119,20 @@ class AdvertisementData:
   */
   connectable/bool
 
-  constructor --.name=null --.service_classes=[] --.manufacturer_data=#[] --.connectable=false:
+  /**
+  Advertise flags. This must be a bitwise or of the BLE_ADVERTISE_FLAG_* constants
```
kasperl, line 123:
```suggestion
  Advertise flags. This must be a bitwise 'or' of the BLE_ADVERTISE_FLAG_* constants
```

lib/ble.toit
```diff
@@ -115,13 +119,20 @@ class AdvertisementData:
...
+    (see $BLE_ADVERTISE_FLAGS_GENERAL_DISCOVERY and similar).
+  */
+  flags/int
+
+  constructor --.name=null --.service_classes=[] --.manufacturer_data=#[]
+              --.connectable=false --.flags=0 --check_size=true:
```
kasperl, line 129: Nit: This is generally applicable. I'm not entirely sure about this, but I don't think we infer parameter types from their default value just yet. 

```suggestion
              --.connectable=false --.flags=0 --check_size/bool=true:
```

lib/ble.toit
```diff
@@ -606,7 +617,9 @@ class Central extends Resource_:
             --name=next[2]
             --service_classes=service_classes
             --manufacturer_data=(next[4]?next[4]:#[])
-            --connectable=next[5]
+            --flags=next[5]
+            --connectable=next[6]
+            --check_size=false
```
kasperl, line 622:
```suggestion
            --no-check_size
```

src/resources/ble_darwin.mm
```diff
@@ -38,12 +38,22 @@ explicit BLEResourceGroup(Process* process)
       : ResourceGroup(process, HostBLEEventSource::instance()) {
   }
 
+  void tear_down() override {
+    tearing_down_ = true;
+    ResourceGroup::tear_down();
+  }
+
+  bool is_tearing_down() { return tearing_down_; }
```
kasperl, line 46:
```suggestion
  bool is_tearing_down() const { return tearing_down_; }
```

src/resources/ble_darwin.mm
```diff
@@ -854,8 +868,10 @@ - (id)initWithResource:(toit::BLEResource*)resource {
     array->at_put(4, custom_data);
   }
 
+  array->at_put(5, Smi::from(0)); // flags not available on Darwin
```
kasperl, line 871:
```suggestion
  array->at_put(5, Smi::from(0)); // Flags are not available on Darwin.
```

src/resources/ble_darwin.mm
```diff
@@ -854,8 +868,10 @@ - (id)initWithResource:(toit::BLEResource*)resource {
...
+
   NSNumber* is_connectable = peripheral->connectable();
-  array->at_put(5, (is_connectable != nil && is_connectable.boolValue == YES)
+  array->at_put(6, (is_connectable != nil && is_connectable.boolValue == YES)
```
kasperl, line 874:
```suggestion
  array->at_put(6, BOOL(is_connectable != nil && is_connectable.boolValue == YES));
```

src/resources/ble_darwin.mm
```diff
@@ -650,9 +660,11 @@ - (id)initWithResource:(toit::BLEResource*)resource {
 
 template<typename T>
 void ServiceContainer<T>::make_deletable() {
-  NSArray<BLEResourceHolder*>* services = [_service_resource_index allValues];
-  for (int i = 0; i < [services count]; i++) {
-    group()->unregister_resource(services[i].resource);
+  if (!group()->is_tearing_down()) {
```
kasperl, line 663:
```suggestion
  // Tearing down the resource group will also delete the services (resources) that
  // this service container holds on to. We don't want to do that twice.
  if (!group()->is_tearing_down()) {
```

src/resources/ble_darwin.mm
```diff
@@ -671,9 +683,11 @@ - (id)initWithResource:(toit::BLEResource*)resource {
 }
 
 void BLEServiceResource::make_deletable() {
-  NSArray<BLEResourceHolder*>* characteristics = [_characteristics_resource_index allValues];
-  for (int i = 0; i < [characteristics count]; i++) {
-    group()->unregister_resource(characteristics[i].resource);
+  if (!group()->is_tearing_down()) {
```
kasperl, line 686:
```suggestion
  // Tearing down the resource group will also delete the characteristics (resources) 
  // that this BLE service holds on to. We don't want to do that twice.
  if (!group()->is_tearing_down()) {
```

src/resources/ble_darwin.mm
```diff
@@ -38,12 +38,22 @@ explicit BLEResourceGroup(Process* process)
...
+
  protected:
   uint32_t on_event(Resource* resource, word data, uint32_t state) override {
     USE(resource);
     state |= data;
     return state;
   }
+
+ private:
+  bool tearing_down_ = false;
```
kasperl, line 56: It almost feels like we should have better support for these nested resources trees with sub-resources that are 'owned' or at least 'held on to' by another resource in the group.

kasperl (approved):
Looks good to me.

lib/ble.toit
kasperl: Done.

lib/ble.toit
kasperl: Done.

src/resources/ble_darwin.mm
kasperl: Done.

src/resources/ble_darwin.mm
kasperl: Done.

src/resources/ble_darwin.mm
kasperl: Done.

src/resources/ble_darwin.mm
kasperl: Done.

kasperl:
As agreed with @mikkeldamsgaard, I am addressing the feedback here https://github.com/toitlang/toit/pull/1205.

PR #1203, opened 2022-11-22
Add IDENTICAL bytecode

kasperl:


lib/core/utils.toit
```diff
@@ -39,8 +39,8 @@ For floats, two NaN's (not-a-number) are identical when they have the same
   `==` where NaN's are never equal, so `float.NAN == float.NAN` is always
   false.
 */
-identical x y:
-  #primitive.core.identical
+//identical x y:
```
kasperl, line 42: Not quite sure what to do with this. In particular, it would be nice to keep the documentation around somehow. Maybe keep it around and let it throw? I'll have to change the builtin lookup rules for that to work (they don't find the builtin, if the identifier resolves in the scope).

kasperl: Fixed.

src/compiler/resolver_method.cc
```diff
@@ -2306,7 +2307,13 @@ ir::Node* MethodResolver::_resolve_call_target(ast::Node* target_node,
       }
       if (method_node->is_static()) {
         check_sdk_protection(name, target_node->range(), method_node->range());
-        return _new ir::ReferenceMethod(method_node, range);
+        // We special case the identical top-level method, because we want to
+        // turn calls to that into a special bytecode.
+        if (name == Symbols::identical) {
```
floitsch, line 2312: We only want to replace `identical` calls to the core method.
(This is different to `assert` which is a keyword like function).
The holder check is probably not necessary, but this way we ensure that we don't refer to a static `Foo.identical` inside a class.
```suggestion
        if (name == Symbols::identical && method_node->is_runtime_method() && method_node->holder == null) {
```

floitsch (approved):
LGTM.

floitsch:
Please add a test where the user added an `identical` function.

src/compiler/resolver_method.cc
floitsch: We could also check for the shape here. However, there really should only be one `identical` in the core libraries, so we should be safe.

kasperl: Done.

PR #1207, opened 2022-11-23
Regularize use of names for (un)encrypted data.

erikcorry:
cleartext: Never encrypted.
plaintext: Not encrypted yet/any more (one word).
ciphertext: Encrypted data (one word).


floitsch (approved):
LGTM.

PR #1209, opened 2022-11-24
Move hash_find out of the interpreter itself

erikcorry:
The other two methods with the INTERPRETER_CORE (IRAM) attribute are fast_at and compare_numbers.


kasperl (approved):
Looks good to me.

PR #1218, opened 2022-11-24
Disable DTLS support in sdkconfig.defaults

erikcorry:
We have no bindings for DTLS so it's not useful.
This reduces the size of the bin by about 3.5k


kasperl (approved):
Looks good to me.

PR #1219, opened 2022-11-25
Remove check whether port is character device on windows.

floitsch:


kasperl (approved):
Looks good to me. Thanks!

PR #1223, opened 2022-11-28
Add ESP32 Wi-Fi scan function.

donghengqaz:


system/extensions/esp32/wifi.toit
```diff
@@ -71,10 +80,15 @@ class WifiServiceDefinition extends NetworkServiceDefinitionBase:
     if not ssid or ssid.is_empty: throw "wifi ssid not provided"
     password/string := effective.get wifi.CONFIG_PASSWORD --if_absent=: ""
 
-    module ::= (state_.up: WifiModule.sta this ssid password) as WifiModule
-    if module.ap:
+    if module_ != null:
```
kasperl, line 83: This doesn't seem right. The NetworkState object is supposed to keep the network up as long as there are users.

system/extensions/esp32/wifi.toit
```diff
@@ -99,14 +113,19 @@ class WifiServiceDefinition extends NetworkServiceDefinitionBase:
       throw "wifi channel must be between 1 and 13"
     broadcast/bool := config.get wifi.CONFIG_BROADCAST --if_absent=: true
 
-    module ::= (state_.up: WifiModule.ap this ssid password broadcast channel) as WifiModule
-    if not module.ap:
+    if module_ != null:
```
kasperl, line 116: This doesn't seem right. The NetworkState object is supposed to keep the network up as long as there are users.

donghengqaz: Yes, the module_ may be closed here, I modify to check `state_.module`.

system/extensions/esp32/wifi.toit
```diff
@@ -118,6 +137,18 @@ class WifiServiceDefinition extends NetworkServiceDefinitionBase:
...
+    if module_:
+      if module_.ap:
+        throw "wifi already established in AP mode"
+    else:
+      module_ = WifiModule.sta this "" ""
+
+    channel := config.get wifi.CONFIG_SCAN_CHANNEL
+    passive := config.get wifi.CONFIG_SCAN_PASSIVE
+    period  := config.get wifi.CONFIG_SCAN_PERIOD
+    return module_.scan channel passive period
```
kasperl, line 151: It feels like this is the problem with the `module` handling. You want to support multiple calls to scan without calling the `wifi_init` primitive multiple times? If that isn't necessary then I would call `state_.up` and `state_.down` from in here, but probably check that `state_.module` is null before doing that if scanning cannot work while we're connected.

```suggestion
    if state_.module:
      throw "wifi already connected or scanning in process"
    module := (state_.up: WifiModule.sta this "" "") as WifiModule
    try:
      channel := config.get wifi.CONFIG_SCAN_CHANNEL
      passive := config.get wifi.CONFIG_SCAN_PASSIVE
      period  := config.get wifi.CONFIG_SCAN_PERIOD
      return module.scan channel passive period
    finally:
      state_.down
```

kasperl (commented):
Thank you for working on this @donghengqaz! Here are my comments:

examples/wifi/scan.toit
```diff
@@ -0,0 +1,39 @@
...
+
+// The standard BLE peripheral demo for simulating a heart rate monitor.
+
+import net.wifi
+import encoding.hex
+
+SCAN_CHANNELS := #[1, 2, 3, 4, 5, 6, 7]
+
+main:
+  ap_list := wifi.scan
```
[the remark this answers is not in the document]
donghengqaz: No, I intend to flow workflow:

1. initialize Wi-Fi to be station mode, and this is background program, so users should not do this
2. users can scan multi timers if needs
3. user connect to target scanned AP

examples/wifi/scan.toit
```diff
@@ -0,0 +1,39 @@
...
+  if ap_list == null:
+    throw "Scan done, but no AP is found"
+
+  print """
+      $(%-32s "SSID") $(%-18s "BSSID") \
+      $(%-6s "RSSI") $(%-8s "Channel") \
+      $(%-8s "Author")\n"""
+
+  ap_list.do:
+    ssid := it[wifi.SCAN_AP_SSID]
```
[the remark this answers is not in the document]
donghengqaz: Yes, this is a much better solution.

examples/wifi/scan.toit
```diff
@@ -0,0 +1,39 @@
+// Copyright (C) 2022 Toitware ApS.
+// Use of this source code is governed by a Zero-Clause BSD license that can
+// be found in the examples/LICENSE file.
+
+// The standard BLE peripheral demo for simulating a heart rate monitor.
+
+import net.wifi
+import encoding.hex
```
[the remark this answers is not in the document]
donghengqaz: This is used in older testing code, I will remove this.

system/extensions/esp32/wifi.toit
```diff
@@ -235,6 +267,30 @@ class WifiModule implements NetworkModule:
...
+    ap_list := List
+    ap_array.do:
+      ap_info := Map
+      ap_info[wifi.SCAN_AP_SSID] = it[WIFI_SCAN_SSID_]
+      ap_info[wifi.SCAN_AP_BSSID] = it[WIFI_SCAN_BSSID_]
+      ap_info[wifi.SCAN_AP_RSSI] = it[WIFI_SCAN_RSSI_]
+      ap_info[wifi.SCAN_AP_AUTHMODE] = it[WIFI_SCAN_AUTHMODE_]
+      ap_info[wifi.SCAN_AP_CHANNEL] = it[WIFI_SCAN_CHANNEL_]
+      ap_list.add ap_info
+    return ap_list
```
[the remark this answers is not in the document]
donghengqaz: OK, I will try this.

examples/wifi/scan.toit
```diff
@@ -0,0 +1,39 @@
...
+
+SCAN_CHANNELS := #[1, 2, 3, 4, 5, 6, 7]
+
+main:
+  ap_list := wifi.scan
+      SCAN_CHANNELS
+      --period_per_channel=1000
```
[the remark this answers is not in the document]
donghengqaz: Using `period_per_channel_us` is easier.

lib/system/api/wifi.toit
```diff
@@ -33,3 +36,6 @@ class WifiServiceClient extends NetworkServiceClient implements WifiService:
 
   rssi handle/int -> int?:
     return invoke_ WifiService.RSSI_INDEX handle
+
+  scan config/Map -> List?:
+    return invoke_ WifiService.SCAN_INDEX config
```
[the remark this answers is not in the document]
donghengqaz: Could we map this in `scan` function of  `lib/net/wifi.toit` ? This may look clearer to me.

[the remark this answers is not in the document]
kasperl: Yeah, that would probably be even better!

examples/wifi/scan.toit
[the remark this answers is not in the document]
floitsch: But what should it do, when a user doesn't follow that order?
Should it throw?

system/extensions/esp32/wifi.toit
donghengqaz: Yes, I would like to scan multi times without repeating to initialize Wi-Fi, because it costs much time. Here using `state_.up` will trigger a connection with ssid="" and password="", so this is not what I expect. What I would like to is that deinitialize Wi-Fi and free resource owned by scan when somewhere calls "connect" or "establish".

kasperl: In that case, I think we should move towards a model where we should go for a init/use*/close model like the network in AP and STA mode. This probably means that we should be returning a resource proxy for the network in the 'ready-to-scan' state and react with `state_.down` when it closes.

Do you need help figuring out how to do that?

examples/wifi/scan.toit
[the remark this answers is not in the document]
donghengqaz: 1. if run "scan" after "connect" or "establish", it should throw error
2. if run "connect" or "establish" after "scan", it should close "scan" before operation

system/extensions/esp32/wifi.toit
donghengqaz: Yes, I want to know how to implement this.

kasperl: A good starting point might be the "Service resources and proxies" section in https://github.com/toitlang/toit/discussions/869. We would need to define a new subclass of `ServiceResource` that represents the network in the scanning state and return a handle for that resource through the `wifi.scanning` API call. That handle can then be turned into a subclass of `ServiceResourceProxy` on the client side and this is where you would add the `scan` method. The `scan` method will pass the handle back through the API and the WiFi service will lookup up the resources (using the handle as key) and act on that.

donghengqaz: I study this, but it is much complex to adding scan function by this method, and event it beyond to what I am able do ... so I decrease some performance to the scan, just one API that scan a set of channels, I think this can cover much use case, because in most applications users may scan all channels and determine to connect which one, if they don't find AP what need, they will also scan all channels, the time cost of initializing Wi-Fi is much less than scan all channels.

In my first opinion, I try to add APIs to scan one channel, and then one by one in sequence, so I have to keep the Wi-Fi is open, but if scan a set of channels one time, we may not need to keep the Wi-Fi open.

src/resources/wifi_esp32.cc
```diff
@@ -166,81 +201,100 @@ class WifiIpEvents : public SystemResource {
...
-        case WIFI_REASON_ASSOC_LEAVE:
-        case WIFI_REASON_ASSOC_EXPIRE:
-        case WIFI_REASON_AUTH_EXPIRE:
-          state |= WIFI_RETRY;
-          break;
-        default:
-          state |= WIFI_DISCONNECTED;
-          break;
+
+  if (system_event->base == WIFI_EVENT) {
```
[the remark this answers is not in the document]
donghengqaz: Something strange, I can' this comment in `Conversation` menu...

system/extensions/esp32/wifi.toit
```diff
@@ -235,6 +267,30 @@ class WifiModule implements NetworkModule:
   rssi -> int?:
     return wifi_get_rssi_ resource_group_
 
+  scan channel/int passive/bool period/int -> List?:
+    if ap or not resource_group_: return null
```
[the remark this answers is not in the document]
donghengqaz: Yes, throwing an exception is better, I think it will not happen, we if wifi state is error, it will throw exception in upper wifi service, not here.

system/extensions/esp32/wifi.toit
```diff
@@ -235,6 +267,30 @@ class WifiModule implements NetworkModule:
...
+
+    resource := wifi_start_scan_ resource_group_ channel passive period
+    scan_events := monitor.ResourceState_ resource_group_ resource
+    state := scan_events.wait
+    if (state & WIFI_SCAN_DONE) == 0: throw "WIFI_SCAN_ERROR"
+    scan_events.dispose
```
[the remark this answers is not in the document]
donghengqaz: Yes, this is better. And when I test, the `scan_events.dispose` can't free event resource directly.

src/resources/wifi_esp32.cc
```diff
@@ -166,81 +201,100 @@ class WifiIpEvents : public SystemResource {
...
+    switch (system_event->id) {
```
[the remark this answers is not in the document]
donghengqaz: Yes, this will be better.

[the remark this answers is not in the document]
donghengqaz: Besides this, I think maybe using 2 event: Wi-Fi event and IP event is better, this is clearer.

system/extensions/esp32/wifi.toit
```diff
@@ -71,10 +80,15 @@ class WifiServiceDefinition extends NetworkServiceDefinitionBase:
     if not ssid or ssid.is_empty: throw "wifi ssid not provided"
     password/string := effective.get wifi.CONFIG_PASSWORD --if_absent=: ""
 
-    module ::= (state_.up: WifiModule.sta this ssid password) as WifiModule
-    if module.ap:
+    if module_ != null and state_.module == null:
```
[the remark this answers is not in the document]
donghengqaz: Scan is refactored to keep origin process.

lib/net/wifi.toit
```diff
@@ -13,8 +13,56 @@ CONFIG_PASSWORD  /string ::= "wifi.password"
...
+    2: "WPA PSK",
+    3: "WPA2 PSK",
+    4: "WPA/WPA2 PSK",
+    5: "WPA2 Enterprise",
+    6: "WPA3 PSK",
+    7: "WPA2/WPA3 PSK",
+    8: "WAPI PSK",
+  }
+
+  constructor .ssid/string .bssid/ByteArray .rssi/int .authmode/int .channel/int:
```
kasperl, line 56: It might be nice to use named arguments for this (and the types are inferred from the fields):

```suggestion
  constructor --.ssid --.bssid --.rssi --.authmode --.channel:
```

lib/net/wifi.toit
```diff
@@ -53,6 +101,33 @@ establish config/Map? -> Interface:
...
+    offset := it * WIFI_SCAN_ELEMENT_COUNT_
+    ap := AccessPoint
+        ap_array[offset + WIFI_SCAN_SSID_]
+        ap_array[offset + WIFI_SCAN_BSSID_]
+        ap_array[offset + WIFI_SCAN_RSSI_]
+        ap_array[offset + WIFI_SCAN_AUTHMODE_]
+        ap_array[offset + WIFI_SCAN_CHANNEL_]
+    ap_list.add ap
+
+  return ap_list.size > 0 ? ap_list : null
```
kasperl, line 129:
```suggestion
  config ::= {
    CONFIG_SCAN_PASSIVE: passive,
    CONFIG_SCAN_CHANNELS: channels,
    CONFIG_SCAN_PERIOD: period_per_channel_ms,
  }

  array := service.scan config
  return List array.size / WIFI_SCAN_ELEMENT_COUNT_:
    offset := it * WIFI_SCAN_ELEMENT_COUNT_
    AccessPoint
        --ssid=array[offset + WIFI_SCAN_SSID_]
        --bssid=array[offset + WIFI_SCAN_BSSID_]
        --rssi=array[offset + WIFI_SCAN_RSSI_]
        --authmode=array[offset + WIFI_SCAN_AUTHMODE_]
        --channel=array[offset + WIFI_SCAN_CHANNEL_]
```

system/extensions/esp32/wifi.toit
```diff
@@ -235,6 +249,26 @@ class WifiModule implements NetworkModule:
   rssi -> int?:
     return wifi_get_rssi_ resource_group_
 
+  scan channels/ByteArray passive/bool period/int -> Array_:
+    if ap or not resource_group_:
+      throw "Wi-Fi is AP mode or not initialized"
```
kasperl, line 254: For consistency:

```suggestion
      throw "wifi is AP mode or not initialized"
```

system/extensions/esp32/wifi.toit
```diff
@@ -235,6 +249,26 @@ class WifiModule implements NetworkModule:
...
+
+    resource := wifi_init_scan_ resource_group_
+    scan_events := monitor.ResourceState_ resource_group_ resource
+    ap_array := Array_ 0
```
kasperl, line 258:
```suggestion
   result := []
```

system/extensions/esp32/wifi.toit
```diff
@@ -235,6 +249,26 @@ class WifiModule implements NetworkModule:
...
+    try:
+      channels.do:
+        wifi_start_scan_ resource_group_ it passive period
+        state := scan_events.wait
+        if (state & WIFI_SCAN_DONE) == 0: throw "WIFI_SCAN_ERROR"
+        scan_events.clear_state WIFI_SCAN_DONE
+        array := wifi_read_scan_ resource_group_
+        ap_array += array
```
kasperl, line 266:
```suggestion
        result.add_all array
```

system/extensions/esp32/wifi.toit
```diff
@@ -235,6 +249,26 @@ class WifiModule implements NetworkModule:
   rssi -> int?:
     return wifi_get_rssi_ resource_group_
 
+  scan channels/ByteArray passive/bool period/int -> Array_:
```
kasperl, line 252:
```suggestion
  scan channels/ByteArray passive/bool period/int -> List:
```

system/extensions/esp32/wifi.toit
```diff
@@ -235,6 +249,26 @@ class WifiModule implements NetworkModule:
...
+        wifi_start_scan_ resource_group_ it passive period
+        state := scan_events.wait
+        if (state & WIFI_SCAN_DONE) == 0: throw "WIFI_SCAN_ERROR"
+        scan_events.clear_state WIFI_SCAN_DONE
+        array := wifi_read_scan_ resource_group_
+        ap_array += array
+    finally:
+      scan_events.dispose
+
+    return ap_array
```
kasperl, line 270:
```suggestion
    return result
```

system/extensions/esp32/wifi.toit
```diff
@@ -118,6 +120,17 @@ class WifiServiceDefinition extends NetworkServiceDefinitionBase:
...
+  scan config/Map -> Array_:
+    if state_.module:
+      throw "wifi already connected or established"
+    module := WifiModule.sta this "" ""
+    channels := config.get wifi.CONFIG_SCAN_CHANNELS
+    passive := config.get wifi.CONFIG_SCAN_PASSIVE
+    period  := config.get wifi.CONFIG_SCAN_PERIOD
+    ap_array := module.scan channels passive period
+    module.disconnect
+    return ap_array
```
kasperl, line 133:
```suggestion
    module := WifiModule.sta this "" ""
    try:
      channels := config.get wifi.CONFIG_SCAN_CHANNELS
      passive := config.get wifi.CONFIG_SCAN_PASSIVE
      period  := config.get wifi.CONFIG_SCAN_PERIOD
      return module.scan channels passive period
    finally:
      module.disconnect
```

kasperl (commented):
Great progress. Here is my final round of comments. I think we're pretty close to being able to land this!

lib/net/wifi.toit
```diff
@@ -13,8 +13,56 @@ CONFIG_PASSWORD  /string ::= "wifi.password"
...
+SCAN_AP_CHANNEL  /string ::= "scan.ap.channel"
+
+WIFI_SCAN_SSID_     ::= 0
+WIFI_SCAN_BSSID_    ::= 1
+WIFI_SCAN_RSSI_     ::= 2
+WIFI_SCAN_AUTHMODE_ ::= 3
+WIFI_SCAN_CHANNEL_  ::= 4
+WIFI_SCAN_ELEMENT_COUNT_ ::= 5
+
+SCAN_TIMEOUT_MS_/int := 1000 /* microseconds */
```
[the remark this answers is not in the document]
donghengqaz: This is milliseconds, and I drop the wrong comment.

lib/net/wifi.toit
```diff
@@ -13,8 +13,56 @@ CONFIG_PASSWORD  /string ::= "wifi.password"
...
 service_/WifiServiceClient? ::= (WifiServiceClient --no-open).open
 
+class AccessPoint:
+  ssid/string := ?
+  bssid/ByteArray := ?
+  rssi/int := ?
+  authmode/int := ?
+  channel/int := ?
+
+  static WIFI_AUTHMODE_NAME_/Map ::= {
```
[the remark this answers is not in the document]
donghengqaz: I think you have more experience, turning this into a list may be better, I will do this.

examples/wifi/scan.toit
```diff
@@ -0,0 +1,28 @@
...
+  if ap_list.size == 0:
+    throw "Scan done, but no AP is found"
+
+  print """
+      $(%-32s "SSID") $(%-18s "BSSID") \
+      $(%-6s "RSSI") $(%-8s "Channel") \
+      $(%-8s "Author")\n"""
+
+  ap_list.do:
+    ap := it as wifi.AccessPoint
```
kasperl, line 24:
```suggestion
  ap_list.do: | ap/wifi.AccessPoint |
```

examples/wifi/scan.toit
```diff
@@ -0,0 +1,28 @@
...
+import net.wifi
+
+SCAN_CHANNELS := #[1, 2, 3, 4, 5, 6, 7]
+
+main:
+  ap_list := wifi.scan
+      SCAN_CHANNELS
+      --period_per_channel_ms=120
+  if ap_list.size == 0:
+    throw "Scan done, but no AP is found"
```
kasperl, line 16: Should this throw or just print?

lib/net/wifi.toit
```diff
@@ -53,6 +101,33 @@ establish config/Map? -> Interface:
...
+    offset := it * WIFI_SCAN_ELEMENT_COUNT_
+    ap := AccessPoint
+        --ssid=data_list[offset + WIFI_SCAN_SSID_]
+        --bssid=data_list[offset + WIFI_SCAN_BSSID_]
+        --rssi=data_list[offset + WIFI_SCAN_RSSI_]
+        --authmode=data_list[offset + WIFI_SCAN_AUTHMODE_]
+        --channel=data_list[offset + WIFI_SCAN_CHANNEL_]
+    ap_list.add ap
+
+  return ap_list
```
kasperl, line 129: You didn't like my suggestion for allocating the resulting list with the right size and returning it?

```suggestion
  data_list := service.scan config
  ap_count := data_list.size / WIFI_SCAN_ELEMENT_COUNT_
  return List ap_count:
    offset := it * WIFI_SCAN_ELEMENT_COUNT_
    AccessPoint
        --ssid=data_list[offset + WIFI_SCAN_SSID_]
        --bssid=data_list[offset + WIFI_SCAN_BSSID_]
        --rssi=data_list[offset + WIFI_SCAN_RSSI_]
        --authmode=data_list[offset + WIFI_SCAN_AUTHMODE_]
        --channel=data_list[offset + WIFI_SCAN_CHANNEL_]
```

system/extensions/esp32/wifi.toit
```diff
@@ -118,6 +120,18 @@ class WifiServiceDefinition extends NetworkServiceDefinitionBase:
...
     return (state_.module as WifiModule).rssi
+  
+  scan config/Map -> List:
+    if state_.module:
+      throw "wifi already connected or established"
+    module := WifiModule.sta this "" ""
+    try:
+      channels := config.get wifi.CONFIG_SCAN_CHANNELS
+      passive := config.get wifi.CONFIG_SCAN_PASSIVE
+      period  := config.get wifi.CONFIG_SCAN_PERIOD
```
kasperl, line 131:
```suggestion
      period := config.get wifi.CONFIG_SCAN_PERIOD
```

kasperl (approved):
Looks good to me!

examples/wifi/scan.toit
donghengqaz: Maybe printing is better.

kasperl:
Fixed the compilation issues here: https://github.com/toitlang/toit/pull/1252.

kasperl:
Landed in f5e210d9eca56757152e044346eb4afa412d0771.

PR #1227, opened 2022-11-28
Added null-check to the send Command helper function for at.Session

### LINKED ISSUE #1226: Wrong parameter type for some at.session.Session calls
The at.session.Session class has helper methods to do different at commands.
They are: action, read, test, and set. 
Each of these helper methods create a corresponding at.session.Command object. The at.session.Command objects take a Duration as an input parameter, but the helper classes take a Duration? as a parameter and passes it along without a null check. This can lead to a type-check failure at runtime.
The erroneous code:
```
class Command:
...
  timeout/Duration

  constructor.action .name --.timeout=DEFAULT_TIMEOUT:
    type = ACTION_
...
class Session:
...
  /**
  Executes an `action` command with the $command_name.
  */
  action command_name/string --timeout/Duration?=null -> Result:
    return send
      Command.action command_name --timeout=timeout
...
```

MikAtCosesy:
Added null-check to the send Command helper function for at.Session to fix #1226.

### DIFF
--- a/lib/at/session.toit
+++ b/lib/at/session.toit
@@ -226,40 +226,60 @@ class Session:
   Executes a `read` command with the $command_name.
   */
   read command_name/string --timeout/Duration?=null -> Result:
-    return send
-      Command.read command_name --timeout=timeout
+    cmd ::= ?
+    if timeout == null:
+      cmd = Command.read command_name
+    else:
+      cmd = Command.read command_name --timeout=timeout
+    return send cmd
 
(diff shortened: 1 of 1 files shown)


MikAtCosesy:
@cla-bot check

kasperl:
We have deprecated the AT library in the toitlang/toit repository and we will be maintaining this as part of the https://github.com/toitware/cellular repository going forward.

PR #1230, opened 2022-11-29
Fix registering of externals in message encoding

kasperl:


src/messaging.cc
```diff
@@ -283,10 +283,7 @@ bool MessageEncoder::encode_byte_array(ByteArray* object) {
     too_many_externals_ = true;
     return false;
   }
-  if (encoding_for_size()) {
-    externals_[externals_count_] = object;
-  }
-  externals_count_++;
+  externals_[externals_count_++] = object;
```
kasperl, line 286: Alternatively we could store null in there if we're encoding for size.

```suggestion
  externals_[externals_count_++] = encoding_for_size() ? null : object;
```

floitsch, line 286: If I understand correctly, the bug is that the condition is not inverted. correct?
It should have been `if (!encoding_for_size())`

floitsch (approved):
LGTM.
Either works for me. I probably have a small preference for the `null` one, as it indicates that we are not using the object when `encoding_for_size()`.

PR #1232, opened 2022-11-29
GC compaction: Don't fix pointers from dead objects.

erikcorry:
Previously we fixed pointers in dead new-space objects that seemed to point into old space after compaction. Now we skip those objects.  They will never be touched by the system again:
* The mutator can't reach them.
* Remembered-set object scans only happen in old space.
* Mark stack overflow object scans only look at gray objects.
* Scavenges don't look at dead objects.
* Finalizers only determine that objects are dead, they don't look inside them at their pointers.

### DIFF
--- a/src/third_party/dartino/object_memory.cc
+++ b/src/third_party/dartino/object_memory.cc
@@ -132,7 +132,15 @@ void Space::iterate_objects(HeapObjectVisitor* visitor, LivenessOracle* filter) 
         ASSERT(size > 0);
         current += size;
       } else {
-        current += object->size(program_);
+        word size = object->size(program_);
+#ifdef DEBUG
+        // Zapping words after the header should be harmless in new-space.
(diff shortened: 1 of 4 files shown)


kasperl (approved):
Looks good to me.

Maybe try to initialize objects with freaky values?

src/third_party/dartino/object_memory.cc
```diff
@@ -132,7 +132,15 @@ void Space::iterate_objects(HeapObjectVisitor* visitor, LivenessOracle* filter)
         ASSERT(size > 0);
         current += size;
       } else {
-        current += object->size(program_);
+        word size = object->size(program_);
+#ifdef DEBUG
```
kasperl, line 136: Should we stop eagerly initializing arrays and instances?

erikcorry: Let's let this cook for a week, then remove the eager initialization.

PR #1233, opened 2022-11-29
AesGcm: Concatenate the rest of the ciphertext and the tag concatenated.

erikcorry:
It just turns out this is a nicer API when finishing an AesGcm encryption most of the time.

### DIFF
--- a/src/third_party/dartino/object_memory.h
+++ b/src/third_party/dartino/object_memory.h
@@ -276,8 +276,6 @@ class SemiSpace : public Space {
 
   virtual bool is_flushed();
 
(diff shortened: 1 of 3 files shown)


src/third_party/dartino/object_memory.h
```diff
@@ -276,8 +276,6 @@ class SemiSpace : public Space {
 
   virtual bool is_flushed();
 
-  void trigger_gc_soon() { limit_ = top_ + SENTINEL_SIZE; }
```
floitsch, line 279: unrelated change?

floitsch (approved):
LGTM.

PR #1235, opened 2022-11-30
Fix stack check overflows (primarily related to preemption)

kasperl:


src/objects.h
```diff
@@ -680,12 +680,139 @@ class FrameCallback {
 };
 
 
+class Method {
```
kasperl, line 683: Moved to make it usable from the Stack implementation.

src/interpreter_run.cc
```diff
@@ -249,16 +249,27 @@ Interpreter::Result Interpreter::run() {
...
+    sp = load_stack(&pending);
+    static_assert(FRAME_SIZE == 2, "Unexpected frame size");
     Object* frame_marker = POP();
     ASSERT(frame_marker == program->frame_marker());
+    bcp = reinterpret_cast<uint8*>(POP());
+    // When we're preempted right when a stack needs to grow, we remember
+    // that we haven't done the correct stack overflow check and taken
+    // care of any necessary stack growth. We do that now that we're
+    // resuming interpretation. This is also another preemption check
+    // so we risk making no progress if we keep getting preempted.
```
floitsch, line 268: ?
```suggestion
    // When we are preempted at a call-site, we haven't done the
    // correct stack overflow check yet. We do the check now,
    // using the remembered 'pending' target method.
    // This is also another preemption check so we risk making no
    // progress if we keep getting preempted.
```

src/objects.h
```diff
@@ -716,10 +850,30 @@ class Stack : public HeapObject {
...
+  static const int GUARD_ZONE_WORDS = 4;
+#endif
+  static const int GUARD_ZONE_SIZE = GUARD_ZONE_WORDS * WORD_SIZE;
+
   static const int LENGTH_OFFSET = HeapObject::SIZE + WORD_SIZE;
   static const int TOP_OFFSET = LENGTH_OFFSET + WORD_SIZE;
   static const int TRY_TOP_OFFSET = TOP_OFFSET + WORD_SIZE;
-  static const int HEADER_SIZE = TRY_TOP_OFFSET + WORD_SIZE;
+  static const int PENDING_STACK_CHECK_METHOD_OFFSET = TRY_TOP_OFFSET + WORD_SIZE;
+  static const int GUARD_ZONE_OFFSET = PENDING_STACK_CHECK_METHOD_OFFSET + WORD_SIZE;
```
floitsch, line 875: Add a comment that the stack grows down.

floitsch (approved):
LGTM.

src/objects.h
kasperl: Done.

PR #1237, opened 2022-11-30
Fewer leaks around process spawn

erikcorry:


src/messaging.cc
```diff
@@ -67,30 +67,47 @@ class NestingTracker {
...
+MessageEncoder::~MessageEncoder() {
   for (unsigned i = 0; i < copied_count_; i++) {
     free(copied_[i]);
   }
+  if (take_ownership_of_buffer_) free(buffer_);
 }
 
-void MessageEncoder::neuter_externals() {
+uint8* MessageEncoder::take_buffer() {
+  uint8* result = buffer_;
```
kasperl, line 109: Move this down to just before setting `buffer_` to null.

src/messaging.h
```diff
@@ -128,22 +130,42 @@ class ObjectNotifyMessage : public Message {
...
+    been encountered, this should be called.  It neuters the external byte
+    arrays and forgets the allocated external buffers, which must now be freed
+    by the receiver.
+  Also takes ownership of the buffer away.
+  */
+  uint8* take_buffer();
 
   bool encode(Object* object) { ASSERT(!encoding_tison()); return encode_any(object); }
-  bool encode_byte_array_external(void* data, int length);
+  bool encode_bytes_external(void* data, int length, bool free_on_failure=true);
```
kasperl, line 168:
```suggestion
  bool encode_bytes_external(void* data, int length, bool free_on_failure = true);
```

src/process.h
```diff
@@ -304,38 +304,41 @@ class AllocationManager {
...
+        : true;
     if (!ok) {
       return null;
     }
     // Don't change this to use C++ array 'new' because that isn't compatible
     // with realloc.
     ptr_ = malloc(length);
     if (ptr_ == null) {
-      process_->object_heap()->set_last_allocation_result(ObjectHeap::ALLOCATION_OUT_OF_MEMORY);
+      if (process_) process_->object_heap()->set_last_allocation_result(ObjectHeap::ALLOCATION_OUT_OF_MEMORY);
```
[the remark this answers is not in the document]
kasperl: I think process might always be non-null. That would make this better again :)

kasperl (approved):
Looks good to me.

PR #1238, opened 2022-11-30
Fix pulse counter test.

floitsch:
We were missing some units due to an undefined precompiler variable.


kasperl (approved):
Looks good to me.

PR #1242, opened 2022-12-03
Copy type stacks lazily

kasperl:


floitsch (approved):
LGTM.
Normally I would have preferred not to use a pointer bit for the 'lazy' bit, but since we already obfuscate pointers with the type-set, it's probably ok.

floitsch:
That said: I think the type pointers aren't "obfuscated".

PR #1248, opened 2022-12-05
Add ChaCha20-Poly1305 support to MbedTLS and primitives.

erikcorry:
This has better performance than AES-GCM and may also be useful for WireGuard (we may have to add more
primitives for that).

This increases the flash size on the device by 4672 bytes.

Adding MbedTLS support for this algorithm aligns our host and device builds, since the host builds always had this support.


kasperl (approved):
Looks good to me.

PR #1250, opened 2022-12-06
Propagate types through run_global_initializer

kasperl:


src/compiler/propagation/type_propagator.cc
```diff
@@ -623,8 +648,14 @@ static void process(TypeScope* scope, uint8* bcp, std::vector<Worklist*>& workli
   OPCODE_END();
 
   OPCODE_BEGIN_WITH_WIDE(LOAD_GLOBAL_VAR_LAZY, index);
+    // The interpreter calls 'run_global_initializer_' to ensure
+    // that we deal with reentrant initialization correctly. This
+    // can throw.
+    propagator->ensure_run_global_initializer();
+    scope->throw_maybe();
+    // Analyze a call to the initialier method.
```
floitsch, line 656:
```suggestion
    // Analyze a call to the initializer method.
    // The type of the global does not depend on the `run_global_initializer_` function,
    // and only on the actual initializer expression.
```

src/compiler/propagation/type_scope.cc
```diff
@@ -21,6 +21,8 @@
 namespace toit {
 namespace compiler {
 
+static const int EXTRA = 1;
```
floitsch, line 24: Add a comment, what this constant is for.

kasperl: Done.

src/compiler/propagation/type_propagator.cc
kasperl: Done.

PR #1252, opened 2022-12-06
Land donghengqaz's wifi scanning support

kasperl:


donghengqaz:
@kasperl thank you. These days I can't compile and test because I can't access the toit/go related network point without VPN.

PR #1255, opened 2022-12-06
Analyze remaining entry points

kasperl:


src/compiler/propagation/type_primitive_core.cc
```diff
@@ -253,5 +253,13 @@ TYPE_PRIMITIVE_ANY(firmware_unmap)
 TYPE_PRIMITIVE_ANY(firmware_mapping_at)
 TYPE_PRIMITIVE_ANY(firmware_mapping_copy)
 
+bool TypePrimitive::uses_entry_task(unsigned module, unsigned index) {
+  return module == 0 && index == (183 - 74);
+}
+
+bool TypePrimitive::uses_entry_spawn(unsigned module, unsigned index) {
+  return module == 0 && index == (171 - 74);
```
kasperl, line 261: Will try to find a better way for this, so we're not depending on the numeric value of the indexes (repeated).

lib/core/exceptions.toit
```diff
@@ -150,10 +150,10 @@ class Exception_:
 
 // Spontaneous entry points invoked by the interpreter when failure occurs at runtime.
 // We use 'rethrow' to avoid extra frame in trace.
-lookup_failure_ receiver selector_or_selector_offset:
+lookup_failure_ receiver selector_offset:
```
floitsch, line 153: When we have a method call for a selector that doesn't exist in the whole system, then we call `lookup_failure_` directly (a static call). In that case we pass the name of the selector as a string.

lib/core/exceptions.toit
```diff
@@ -169,14 +169,13 @@ program_failure_ bci:
 
 /**
 Signals an 'as' failure.
-$id might be either:
-- a class name of type $string
-- an absolute BCI of the 'as' check.
+The absolute bytecode index of the 'as' check is
```
floitsch, line 172: Same here: when there is no instantiation of the class, then we call the `as_check_failure_` method directly with the name of the class (since there isn't any class-id to check against).

src/compiler/propagation/type_primitive_core.cc
```diff
@@ -253,5 +253,13 @@ TYPE_PRIMITIVE_ANY(firmware_unmap)
 TYPE_PRIMITIVE_ANY(firmware_mapping_at)
 TYPE_PRIMITIVE_ANY(firmware_mapping_copy)
 
+bool TypePrimitive::uses_entry_task(unsigned module, unsigned index) {
+  return module == 0 && index == (183 - 74);
```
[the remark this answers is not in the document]
floitsch: as discussed: try to use the resolver_primitive.h

src/compiler/propagation/type_propagator.cc
```diff
@@ -84,6 +84,82 @@ void TypePropagator::ensure_entry_main() {
...
+  call_static(null, &scope, null, target);
+  has_entry_task_ = true;
+}
+
+void TypePropagator::ensure_lookup_failure() {
+  if (has_lookup_failure_) return;
+  TypeScope scope(2, words_per_type());
+  TypeStack* stack = scope.top();
+  stack->push_any();           // receiver
+  stack->push_smi(program());  // selector_offset
```
floitsch, line 112: Maybe add a comment explaining that implicit lookup failures always have the integer selector offset.
Explicit ones may have a string but are evaluated normally anyway.

src/compiler/propagation/type_propagator.cc
```diff
@@ -84,6 +84,82 @@ void TypePropagator::ensure_entry_main() {
...
+  call_static(null, &scope, null, target);
+  has_lookup_failure_ = true;
+}
+
+void TypePropagator::ensure_as_check_failure() {
+  if (has_as_check_failure_) return;
+  TypeScope scope(2, words_per_type());
+  TypeStack* stack = scope.top();
+  stack->push_any();           // receiver
+  stack->push_smi(program());  // bci
```
floitsch, line 123: Same. Explicit ones can be a string.

floitsch (approved):
LGTM.

lib/core/exceptions.toit
kasperl: Reverted my change.

src/compiler/propagation/type_propagator.cc
kasperl: Fixed.

src/compiler/propagation/type_propagator.cc
kasperl: Fixed.

PR #1256, opened 2022-12-07
Fix type propagation through finally with parameters

kasperl:


tests/type_propagation/gold/richards_test.gold
```diff
@@ -191,12 +191,12 @@ Scheduler.current_tcb= tests/type_propagation/richards_test.toit
 
 Scheduler.current_id tests/type_propagation/richards_test.toit
  - argument 0: {Scheduler}
-  0[009] - load field local 82 // {Null_|SmallInteger_|ByteArray_}
+  0[009] - load field local 82 // {Null_|LargeInteger_|SmallInteger_|ByteArray_}
```
kasperl, line 194: This is a weird side-effect from calling `id` (getter) on the any type. There is now a path through the service code that handles exceptions correctly and knows that `id` can be a large integer. This is the same reason ByteArray is in here in the first place, but that stems from id on a different class.

floitsch (approved):
LGTM.

PR #1258, opened 2022-12-07
Call --on_error for 'int.parse ""'.

floitsch:
Also make it a parsing error and not an out-of-bounds error.


kasperl (approved):
Looks good to me.

PR #1264, opened 2022-12-08
Allow json.decode_stream to take a BufferedReader.

erikcorry:
By using a BufferedReader instead of a plain Reader this means you can stream more than one JSON object from a source.


floitsch (approved):
LGTM.

PR #1266, opened 2022-12-09
Avoid infinite block analysis recursion

kasperl:


src/compiler/propagation/type_propagator.cc
```diff
@@ -382,7 +400,61 @@ void TypePropagator::call_virtual(MethodTemplate* caller, TypeScope* scope, uint
...
+      // propagating types through them.
+      block->argument(i)->merge(this, argument);
+    }
+  }
+
+  // Drop the arguments from the stack.
+  for (int i = 0; i < arity; i++) stack->pop();
+
+  // If the return type of this block changes, we enqueue the
+  // surrounding method again.
```
floitsch, line 426: Do we analyze the whole method again?
Can we not continue at the bytecode after the block call?

src/compiler/propagation/type_propagator.cc
```diff
@@ -394,7 +466,9 @@ void TypePropagator::propagate_through_lambda(Method method) {
     // anything even though that isn't great.
     arguments.push_back(ConcreteType::any());
   }
-  find(method, arguments);
+  MethodTemplate* callee = find_method(method, arguments);
+  scope->top()->push_smi(program());
```
floitsch, line 470: What's the smi?
Add a comment.

src/compiler/propagation/type_propagator.cc
```diff
@@ -490,33 +564,31 @@ void TypePropagator::add_site(uint8* site, TypeVariable* result) {
...
-      MethodTemplate* candidate = templates[i];
-      if (candidate->matches(target, arguments)) {
-        return candidate;
-      }
-    }
-    MethodTemplate* result = instantiate(target, arguments);
-    templates.push_back(result);
-    result->propagate();
-    return result;
+MethodTemplate* TypePropagator::find_method(Method target, std::vector<ConcreteType> arguments) {
```
floitsch, line 567: maybe rename to `find_or_create_method`?
(same for `find_block`).

src/compiler/propagation/type_propagator.h
```diff
@@ -84,41 +86,45 @@ class TypePropagator {
...
+      , propagator_(propagator)
       , method_(method)
       , arguments_(arguments)
       , result_(propagator->words_per_type()) {}
 
+  ~MethodTemplate() {
+    delete link_;
+  }
+
+  MethodTemplate* link() const { return link_; }
```
floitsch, line 116: I find `link` to be a bad name. Makes me think of an action (the verb "link").
Since this is a linked list, I would just go with `next`.

I would also add a comment on the class, explaining that it has a field for a linked-list. You could mention where it is used for, but I would be ok with just mentioning the fact that it is a linked-list element.

floitsch (approved):
LGTM.

PR #1268, opened 2022-12-09
Use a wrapper class for the TLS record header bytes

erikcorry:


lib/tls/session.toit
```diff
@@ -483,8 +492,8 @@ class SymmetricSession_:
     // don't send too large records.  This size is intended to fit in two MTUs on
     // Ethernet.
     List.chunk_up from to 2800: | from2/int to2/int length2 |
-      record_header := #[APPLICATION_DATA_, 3, 3, 0, 0]
-      BIG_ENDIAN.put_uint16 record_header 3 length2
+      record_header := RecordHeader_ #[APPLICATION_DATA_, 3, 3, 0, 0]
```
floitsch, line 495: could be called `size` or `size2`.

floitsch (approved):
LGTM.

PR #1270, opened 2022-12-10
Propagate types through non-local returns that use `continue.` syntax

kasperl:


src/compiler/propagation/type_propagator.cc
```diff
@@ -1106,15 +1106,27 @@ static TypeScope* process(TypeScope* scope, uint8* bcp, std::vector<Worklist*>&
   OPCODE_END();
 
   OPCODE_BEGIN(NON_LOCAL_RETURN);
+    int level = stack->local(0).block()->level();
     stack->pop();  // Pop block.
-    method->ret(propagator, stack);
+    if (level == 0) {
+      method->ret(propagator, stack);
+    } else {
+      BlockTemplate* block = worklists[level]->block();
```
floitsch, line 1114: Add a comment, reminding readers that the block-level and the work-list level are synchronized.

src/compiler/propagation/type_propagator.cc
```diff
@@ -1106,15 +1106,27 @@ static TypeScope* process(TypeScope* scope, uint8* bcp, std::vector<Worklist*>&
   OPCODE_END();
 
   OPCODE_BEGIN(NON_LOCAL_RETURN);
+    int level = stack->local(0).block()->level();
```
floitsch, line 1109:
```suggestion
    // For 'continue.label' the pushed block is the block from which the
    // non-local return should return.
    int level = stack->local(0).block()->level();
```

src/compiler/propagation/worklist.h
```diff
@@ -29,15 +29,19 @@ class Worklist {
...
 
+  BlockTemplate* block() const { return block_; }
+
   bool has_next() const { return !unprocessed_.empty(); }
   Item next();
 
   TypeScope* add(uint8* bcp, TypeScope* scope, bool split);
 
  private:
+  BlockTemplate* const block_;
```
floitsch, line 43:
```suggestion
  // If this worklist is for a block, contains the value of it. Null otherwise.
  BlockTemplate* const block_;
```

floitsch (approved):
LGTM.

src/compiler/propagation/type_propagator.cc
kasperl, line 1114:
```suggestion
      // The worklists keep track of the blocks they correspond
      // to. The outermost worklist has a null block because it 
      // corresponds to a method.
      BlockTemplate* block = worklists[level]->block();
```

kasperl: Done.

PR #1272, opened 2022-12-10
Factor out utility to set a range of bits in a bit field.

erikcorry:


kasperl (approved):
Looks good to me. I think we should add a test case in tests/ctest/, but I'll try to do that for TypeSet in another PR.

PR #1273, opened 2022-12-11
Re-enable virtual call optimizations

kasperl:


tests/type_propagation/gold/null_equals_test.gold
```diff
@@ -70,7 +70,7 @@ A tests/type_propagation/null_equals_test.toit
 
 A.== tests/type_propagation/null_equals_test.toit
  - argument 0: {A}
- - argument 1: {String_|Null_|SmallInteger_|A}
+ - argument 1: {*}
```
kasperl, line 73: This got worse. I'll need to understand why before submitting this.

kasperl: I now understand what's happening. Inside a lambda method, we don't know what the type of 'this' is in the type propagation phase. When we inline the 'load field' bytecode that loads from 'this' into the lambda method -- instead of calling a named getter -- we get broader types, because stores to all fields with the same index (not the same name anymore) flow into it. 

This can be addressed by dealing better with captured variables (in particular final ones) for lambdas. All for another PR.

src/compiler/propagation/type_propagator.cc
```diff
@@ -388,9 +388,45 @@ void TypePropagator::call_method(MethodTemplate* caller,
...
+  } else {
+    Program* program = this->program();
+    TypeSet receiver = stack->local(arity);
+    TypeSet::Iterator it(receiver, words_per_type_);
+    int variants = 0;
+    while (it.has_next()) {
+      unsigned id = it.next();
+      int entry_index = id + offset;
+      int entry_id = program->dispatch_table[entry_index];
+      if (entry_id < 0) continue;
```
floitsch, line 419: How is that possible?

src/compiler/propagation/type_propagator.cc
```diff
@@ -388,9 +388,45 @@ void TypePropagator::call_method(MethodTemplate* caller,
...
+    // super calls where we cannot find the (virtual)
+    // super method in the dispatch table entries for
+    // the receiver because they have been overridden.
+    is_static = receiver.size(words_per_type_) <= 1;
+  }
+
+  if (is_static) {
+    call_method(caller, scope, site, target, arguments);
+  } else {
+    Program* program = this->program();
```
floitsch, line 411: Wouldn't it be easier to just call `call_virtual` from here?
Actually, from line 407, and returning, so that you can move the `stack->push_empty from line 394 to after the `call_virtual(...); return`.

If we are doing something different than `call_virtual` add a comment explaining why that's not what we are doing.

floitsch (approved):
LGTM.

PR #1277, opened 2022-12-12
Register types for loads of outer blocks

kasperl:


src/compiler/propagation/type_propagator.cc
```diff
@@ -254,7 +254,7 @@ void TypePropagator::propagate(TypeDatabase* types) {
   sites_.for_each([&](uint8* site, Set<TypeVariable*>& results) {
     type.clear(words_per_type());
     for (auto it = results.begin(); it != results.end(); it++) {
-      type.add_all((*it)->type(), words_per_type());
+      type.add_all_even_blocks((*it)->type(), words_per_type());
```
floitsch, line 257: It took me way too long to realize this is not talking about "even blocks", where "even" means not odd.
```suggestion
      type.add_all_also_blocks((*it)->type(), words_per_type());
```

src/compiler/propagation/type_set.h
```diff
@@ -53,7 +53,7 @@ class TypeSet {
     unsigned next() {
       ASSERT(has_next());
       uword bits = bits_;
-      // Compute the index of the next bit by counting 
+      // Compute the index of the next bit by counting
```
floitsch, line 56: not important, but the comment should fit on one line. no?

floitsch (approved):
LGTM.

src/compiler/propagation/type_propagator.cc
kasperl: Done.

PR #1284, opened 2022-12-13
Remember to replace receiver in FieldStore

kasperl:


src/compiler/optimizations/utils.cc
```diff
@@ -63,7 +63,7 @@ Type compute_guaranteed_type(Expression* node, Class* holder, Method* method) {
     return field->type();
   } else if (node->is_Typecheck()) {
     auto check = node->as_Typecheck();
-    return check->type();
+    if (check->is_as_check()) return check->type();
```
floitsch, line 66: I theory we know that the type is 'bool' otherwise, but I don't see an easy way to get to that type here.

floitsch (approved):
LGTM.
oops.

src/compiler/optimizations/utils.cc
kasperl: Yeah, I thought about that. The type checker knows this (and more).

PR #1285, opened 2022-12-13
Check propagated types in interpreter

kasperl:


src/compiler/propagation/type_database.cc
```diff
@@ -36,10 +41,97 @@ TypeDatabase::~TypeDatabase() {
...
+
+void TypeDatabase::check_return(uint8* bcp, Object* value) const {
+  // TODO(kasper): This isn't super nice, but we have to avoid
+  // getting hung up over the intrinsic bytecodes. We sometimes
+  // return from a block and restart at the intrinsic bytecode,
+  // but we don't care about that for now. We could make the
+  // propagator allow any value as the top stack element here,
+  // but it would achieve the same things as this check.
+  uint8 opcode = *bcp;
+  if (opcode > HALT) return;
```
floitsch, line 73: Might be easier to read if you replace the `opcode > HALT` with:
`opcode >= INTRINSIC_SMI_REPEAT`.
You could have a static assert that `HALT + 1 == INTRINSIC_SMI_REPEAT` and that `INTRINSIC_HASH_DO + 1 == ILLEGAL_END`, and maybe even that there are only 4 of them.
That should make sure that we don't accidentally hit this `if` for the wrong bytecodes if we reshuffle them.

src/compiler/propagation/type_database.cc
```diff
@@ -36,10 +41,97 @@ TypeDatabase::~TypeDatabase() {
...
+    }
+    Smi* class_id = is_smi(argument)
+        ? program_->smi_class_id()
+        : HeapObject::cast(argument)->class_id();
+    if (type.contains(class_id->value())) continue;
+    FATAL("method has wrong argument type %d @ %d: %d", class_id->value(), i, position);
+  }
+}
+
 TypeDatabase* TypeDatabase::compute(Program* program) {
```
floitsch, line 120: Add a comment that it returns an object that should be freed with `delete`.

floitsch (approved):
LGTM.

PR #1288, opened 2022-12-16
Allow more TLS certificate data to stay in flash

erikcorry:


src/resources/x509.cc
```diff
@@ -27,7 +27,24 @@
...
+  return c == ' ' or c == '\n' or c == '\r';
+}
+
+bool X509ResourceGroup::is_pem_format(const uint8* data, size_t length) {
+  const char HEADER[] = "-----BEGIN ";
+  while (length > 0 && is_white_space(data[0])) {
+    length--;
+    data++;
+  }
+  if (length < sizeof(HEADER)) return false;
```
kasperl, line 42: Pick sizeof(HEADER) or strlen(HEADER) -- maybe put it in a local const -- and use that consistently.

src/resources/x509.cc
```diff
@@ -49,7 +66,16 @@ Object* X509ResourceGroup::parse(Process* process, const uint8_t* encoded, size_
...
+  int ret;
+  if (is_pem_format(encoded, encoded_size)) {
+    ret = mbedtls_x509_crt_parse(cert->cert(), encoded, encoded_size);
+  } else {
+    if (in_flash) {
+      ret = mbedtls_x509_crt_parse_der_nocopy(cert->cert(), encoded, encoded_size);
+    } else {
+      ret = mbedtls_x509_crt_parse_der(cert->cert(), encoded, encoded_size);
+    }
+  }
```
kasperl, line 78:
```suggestion
  } else if (in_flash) {
    ret = mbedtls_x509_crt_parse_der_nocopy(cert->cert(), encoded, encoded_size);
  } else {
    ret = mbedtls_x509_crt_parse_der(cert->cert(), encoded, encoded_size);
  }
```

kasperl (approved):
Looks good to me.

PR #1292, opened 2022-12-19
Improve types for is checks and report types for is/as checks

kasperl:


src/compiler/propagation/type_propagator.cc
```diff
@@ -1408,6 +1430,17 @@ void BlockTemplate::invoke_from_try_block() {
 }
 
 void BlockTemplate::propagate(TypeScope* scope, std::vector<Worklist*>& worklists, bool linked) {
+  // To avoid having empty types on the stack while we analyze
+  // block methods, we bail out early if there are any argument
+  // types that are empty. This happens before we've seen the
+  // first invocation of the block.
+  const int words_per_type = scope->words_per_type();
+  for (int i = 1; i < arity(); i++) {
+    if (argument(i)->type().is_empty(words_per_type)) {
```
erikcorry, line 1439: It's a bit strange that a TypeSet does not now its own width.

erikcorry (approved):

src/compiler/propagation/type_propagator.cc
kasperl: Yeah, that might be worth looking into as things mature.

PR #1293, opened 2022-12-19
Stop following impossible branches during type propagation

kasperl:


src/compiler/propagation/type_propagator.cc
```diff
@@ -1067,17 +1067,27 @@ static TypeScope* process(TypeScope* scope, uint8* bcp, std::vector<Worklist*>&
   OPCODE_END();
 
   OPCODE_BEGIN(BRANCH_IF_TRUE);
+    TypeSet top = stack->local(0);
+    bool can_be_falsy = top.contains_null(program) || top.contains_false(program);
```
erikcorry, line 1071: I think you could do the same for truthy.  If the only types present are in the set {null, false} then it can't be truthy.

erikcorry (approved):

PR #1295, opened 2022-12-21
Introduce type oracle that maps from optimized IR nodes to unoptimized ones

kasperl:


src/compiler/compiler.cc
```diff
@@ -1515,9 +1516,17 @@ toit::Program* construct_program(ir::Program* ir_program,
 
   ASSERT(_sorted_by_inheritance(ir_program->classes()));
 
-  if (run_optimizations) optimize(ir_program, propagated_types);
+  if (run_optimizations) optimize(ir_program, oracle);
   tree_shake(ir_program);
 
+  if (propagated_types) {
```
erikcorry, line 1522: Needs a comment on how ordering is preserved between the two runs.

erikcorry (approved):

src/compiler/compiler.cc
kasperl: Done.

PR #1298, opened 2022-12-21
Move ESP32 TCP buffers on-heap

erikcorry:
This also makes them smaller.  The old size of about 1500 bytes packed very poorly in 4k pages, often wasting about a kilobyte of memory for every two packets.


src/resources/tcp_esp32.cc
```diff
@@ -408,26 +384,44 @@ PRIMITIVE(read)  {
...
     }
 
-    ByteArray* array = allocate_read_buffer(process, p);
+    if (p->tot_len == 0) return Smi::from(-1);  // Nothing to read.
+
+    int offset = socket->read_offset();
+    // Wifi MTU is 1500 bytes, subtract a 20 byte TCP header and we have 1480.
+    const int MAX_SIZE = 496;  // Three nicely-packable byte arrays per 1480 MTU.
+    int tot_len = p->tot_len;
+    ByteArray* array = process->allocate_byte_array(Utils::min(MAX_SIZE, tot_len - offset), false);  // On-heap byte array.
```
kasperl, line 393: I suggest computing the length here and using that instead of `bytes.length()` further down.

src/resources/tcp_esp32.cc
```diff
@@ -408,26 +384,44 @@ PRIMITIVE(read)  {
       return Smi::from(-1);
     }
 
-    ByteArray* array = allocate_read_buffer(process, p);
+    if (p->tot_len == 0) return Smi::from(-1);  // Nothing to read.
+
+    int offset = socket->read_offset();
+    // Wifi MTU is 1500 bytes, subtract a 20 byte TCP header and we have 1480.
+    const int MAX_SIZE = 496;  // Three nicely-packable byte arrays per 1480 MTU.
+    int tot_len = p->tot_len;
```
kasperl, line 392: Move this up before the `p->tot_len == 0` check and maybe give `tot_len` a better local name?

src/resources/tcp_esp32.cc
```diff
@@ -408,26 +384,44 @@ PRIMITIVE(read)  {
...
-
-      pbuf* n = p->next;
-      // Free the first part of the chain.
-      if (n != null) pbuf_ref(n);
-      pbuf_free(p);
-      p = n;
+    int bytes_to_ack = 0;
+    int copied = 0;
+    while (copied < bytes.length()) {
+      int to_copy = Utils::min(p->len - offset, bytes.length() - copied);
```
kasperl, line 401: Maybe cache `p->len` in a local?

src/resources/tcp_esp32.cc
```diff
@@ -408,26 +384,44 @@ PRIMITIVE(read)  {
       return Smi::from(-1);
     }
 
-    ByteArray* array = allocate_read_buffer(process, p);
+    if (p->tot_len == 0) return Smi::from(-1);  // Nothing to read.
+
+    int offset = socket->read_offset();
```
kasperl, line 389: `socket->read_offset()` was just unused before? I might be inclined to couple the offset and buffer more.

```
int offset;
pbuf* p = socket->get_read_buffer(p, &offset);
```

src/resources/tcp_esp32.cc
```diff
@@ -408,26 +384,44 @@ PRIMITIVE(read)  {
...
+      memcpy(bytes.address() + copied, payload + offset, to_copy);
+      copied += to_copy;
+      offset += to_copy;
+      if (offset == p->len) {
+        pbuf* n = p->next;
+        bytes_to_ack += p->len;
+        // Free the first part of the chain.
+        if (n != null) pbuf_ref(n);
+        pbuf_free(p);
+        p = n;
```
kasperl, line 412: If `n` is null here, we know that we've copied enough because of the way `tot_len` works? Add a comment.

src/resources/tcp_esp32.cc
```diff
@@ -408,26 +384,44 @@ PRIMITIVE(read)  {
...
+        offset = 0;
+      }
     }
 
     socket->set_read_buffer(p);
+    socket->set_read_offset(offset);
```
kasperl, line 418: Would it make sense to join these two calls?

```suggestion
    socket->set_read_buffer(p, offset);
```

kasperl (approved):
Looks good to me.

PR #1305, opened 2022-12-27
Add a note to missing colon and semicolon

erikcorry:


src/compiler/parser.cc
```diff
@@ -1512,7 +1512,10 @@ Expression* Parser::parse_if() {
     condition = parse_expression_or_definition(true);
   }
   if (!optional_delimiter(Token::COLON)) {
-    report_error("Missing ':' for 'if' condition");
+    diagnostics()->start_group();
```
floitsch, line 1515: It feels like second position isn't really helpful anyway. Maybe just
```
report_error(range, "Missing colon for 'if' condition").
```
It might even be worthwhile to shorten the range to 0, so it doesn't underline the following expression.

floitsch (approved):
LGTM.
But I don't think the 'note' is necessary.

floitsch (approved):
LGTM.

PR #1306, opened 2022-12-28
Add flat_map method to strings

erikcorry:


lib/core/collections.toit
```diff
@@ -1078,6 +1078,16 @@ interface ByteArray:
...
 
+  /**
+  Maps every byte with a mapping block.
+  The block is invoked with an integer argument in the range between 0 and
+    0xff, inclusive.
+  If all invocations of the block return integers between 0 and 0xff,
+    inclusive, then the return value is a byte array.
+  If some invocations return other values, the return value is a List.
+  */
+  map [block] -> any
```
floitsch, line 1089: Not sure I like the dynamic changing.
Maybe:
```
map --bytes/bool [block] -> ByteArray:
  if not bytes: throw "INVALID ARGUMENT"

map [block] -> List
```

Edit: I see... later this is used for unicode. Not sure how to handle that.

floitsch (commented):
Still need to think about this.

lib/core/collections.toit
floitsch: Hmm.
what about:

```
/** Maps bytes to bytes. */
map --in_place/bool=false --bytes/bool [block] -> ByteArray

/** Maps UTF8 characters to UTF8 characters */
map --unicode/bool [block] -> ByteArray
// We could even consider:
/**
Maps UTF8 characters to UTF8 characters.
If $in_place is true, attempts to store the result in this byte array.
  If the resulting byte-array is not the same size as this byte array allocates a fresh one and returns it.
  This byte-array should not be used anymore.
*/
map --in_place/bool=false --unicode/bool -> ByteArray

/** Maps each byte to the result of calling the given $block. */
map [block] -> List
```

floitsch: Admittedly, the wording "is not the same size" is not enough (as a single-byte character could explose to multiple bytes, before later characters shrink the size of the byte array again).

erikcorry: The ByteArray code does not handle Unicode in any way.  You are thinking about the string code.

floitsch: I'm suggesting that it should know about it. The main reason that you want to be able to return either an int or a byte is for unicode.

lib/core/string.toit
```diff
@@ -244,6 +244,65 @@ abstract class string implements Comparable:
...
+  To get a list or byte array as the return value instead of a string, use
+    `str.to_byte_array.map` instead.
+  # Examples.
+  heavy_metalize str/string -> string:
+    return str.map: | c |
+      {'o': 'ö', 'a': 'ä', 'u': 'ü', 'ä': "\u{20db}a"}.get c --if_absent=:c
+  lower_case str/string -> string:
+    return str.map: | c | ('A' <= c <= 'Z') ? c - 'A' + 'a' : c
+  */
+  map [block] -> string:
```
erikcorry, line 272: Could be called flat_map https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

floitsch: flat-map would also be an option, but in that case the block would need to return a byte-array. For unicode just returning the rune seems simpler.

floitsch: Sorry. I thought we were still on the ByteArray.

floitsch: Yes. 'flat_map' is definitely better here. Not sure if it should be `map --flatten` or `flat_map`. The latter is more commonly used in literature, but `--flatten` makes the function easier to find.

lib/core/collections.toit
erikcorry: You can always do to_string.map if you want to handle Unicode. 

A byte array can contain byte sequences that are not valid UTF-8.  Feels wrong to mix Unicode into byte array operations. 

The main reason to allow return values that are not ints in the 0-255 range is if you were using the byte array as a more memory compact list of ints, but you don't want to be similarly restricted after map.

eg `(my_byte_array.map: "$(%02x it)").join ""`  // Convert #[1, 2, 3] to "010203"

lib/core/string.toit
erikcorry: I don't think I want to implement the non-flat version for strings, so the --flatten feels a bit strange.

floitsch: Then `flat_map`.

lib/core/string.toit
```diff
@@ -244,6 +244,65 @@ abstract class string implements Comparable:
...
+  The return value is assembled from the return values of the block.
+  If a string is returned from the block it is inserted at that point in the
+    return value.
+  If a byte array is returned from the block it is converted to a string and
+    treated like a string.  The byte array must contain whole, valid UTF-8
+    sequences.
+  If an integer is returned from the block it is treated as a Unicode code
+    point, and the corresponding code point is inserted.
+  If the block returns null, this is treated like the zero length string.
+  If the block returns a collection, then every element in the collection
```
floitsch, line 260: This isn't implemented.

Also, I think we shouldn't use `Collection` in general. (I would even vote in favor of removing it entirely).
Here, the order matters, so `List` is the only thing that would make sense.

lib/core/string.toit
```diff
@@ -244,6 +244,65 @@ abstract class string implements Comparable:
...
+          position = 0
+        position += write_utf_8_to_byte_array byte_array position replacement
+      else if replacement != null:
+        throw "Invalid replacement"
+
+    for i := 0; i < size; i++:
+      rune := this[i]
+      if rune:
+        replacement := block.call rune
+        if replacement is Collection:
```
erikcorry, line 299: Here is where the Collection support is implemented.

tests/string_byte_array_map_test.toit
```diff
@@ -0,0 +1,79 @@
...
+  return str.map: | c | ('A' <= c <= 'Z') ? c - 'A' + 'a' : c
+
+heavy_metalize str/string -> string:
+  return str.map: | c |
+    ROCK_DOTS_MAP_.get c --if_absent=(: c)
+
+natoize str/string -> string:
+  result := str.map: | c |
+    if 'a' <= c <= 'z':
+      [NATO[c - 'a'], " "]
```
erikcorry, line 61: Here is where the Collection support is tested.

lib/core/string.toit
erikcorry: All our collections are ordered.

lib/core/collections.toit
floitsch: I don't think we should optimize our APIs for such a use-case. If you have a list of numbers, then you should use a list; potentially wrapping a byte-array for more efficient storage.  ByteArrays and Lists are obviously related, but there is a reason they don't share an interface or common super class. One is data, the other is sequence of integers. I don't want to have a type (-> any) where we mix them up. That's the reason I suggested to have different functions.

lib/core/string.toit
```diff
@@ -244,6 +244,65 @@ abstract class string implements Comparable:
       rune := this[i]
       if rune: block.call rune
 
+  /**
+  For every unicode character in the string, call the block.
```
floitsch, line 248:
```suggestion
  Calls the given $block for every unicode character in the string.
```

lib/core/string.toit
```diff
@@ -244,6 +244,65 @@ abstract class string implements Comparable:
...
+    is handled like the above actions, but this is only done for one level -
+    lists of lists are not flattened in this way.
+  To get a list or byte array as the return value instead of a string, use
+    `str.to_byte_array.map` instead.
+  # Examples.
+  heavy_metalize str/string -> string:
+    return str.flat_map: | c |
+      {'o': 'ö', 'a': 'ä', 'u': 'ü', 'ä': "\u{20db}a"}.get c --if_absent=:c
+  lower_case str/string -> string:
+    return str.flat_map: | c | ('A' <= c <= 'Z') ? c - 'A' + 'a' : c
```
floitsch, line 270:
~~~suggestion
  ```
  heavy_metalize str/string -> string:
    return str.flat_map: | c |
      {'o': 'ö', 'a': 'ä', 'u': 'ü', 'ä': "\u{20db}a"}.get c --if_absent=: c
  ```
  ```
  lower_case str/string -> string:
    return str.flat_map: | c | ('A' <= c <= 'Z') ? c - 'A' + 'a' : c
  ```
~~~

tests/string_flatmap_test.toit
```diff
@@ -0,0 +1,64 @@
+import expect show *
```
floitsch, line 1:
```suggestion
// Copyright (C) 2023 Toitware ApS.
// Use of this source code is governed by a Zero-Clause BSD license that can
// be found in the tests/LICENSE file.

import expect show *
```

floitsch (approved):
LGTM.

PR #1308, opened 2022-12-30
When rehashing a Set or Map, reduce the peak memory used.

erikcorry:
For large sets or maps we build a new index from scratch when we outgrow the old one.  In this case we reuse the arraylets from the old index, rather than creating new arraylets and keeping the old ones alive during rehashing.

This also changes the rules so that we go to LargeArray_ as soon as we exceed the arraylet size.  Previously we stayed at SmallArray_ until we hit the max size for page allocations.  On embedded this will avoid arrays in the 2-3.9k size range, which could waste memory on fragmentation.

### DIFF
--- a/lib/core/collections.toit
+++ b/lib/core/collections.toit
@@ -845,8 +845,8 @@ abstract class Array_ extends List:
   // Optimized helper method for iterating over the array elements.
   abstract do_ end/int [block] -> none
 
-  /** Create a new array, copying up to $copy_length elements from this array. */
-  abstract resize_for_list_ copy_length/int new_length/int filler -> Array_
+  /// Create a new array, copying up to $copy_size elements from this array.
+  abstract resize_for_list_ copy_size/int new_size/int filler -> Array_
 
   /**
   Returns the given $collection as an $Array_.
@@ -910,15 +910,15 @@ class SmallArray_ extends Array_:
       // argument. We force this to throw by doing the same here.
       block.call this[0]
 
-  /// Creates a new array of size $new_length, copying up to $copy_length elements from this array.
-  resize_for_list_ copy_length/int new_length/int filler -> Array_:
+  /// Creates a new array of size $new_size, copying up to $copy_size elements from this array.
+  resize_for_list_ copy_size/int new_size/int filler -> Array_:
     #primitive.core.array_expand:
       // Fallback if the primitive fails.  For example, the primitive can only
       // create SmallArray_ so we hit this on the border between SmallArray_
       // and LargeArray_.
 
-      result := Array_ new_length filler
-      limit := min copy_length new_length
+      result := Array_ new_size filler
+      limit := min copy_size new_size
       for i := 0; i < limit; i++:
         result[i] = this[i]
       return result
@@ -945,12 +945,12 @@ class LargeArray_ extends Array_:
   constructor .size_/int filler/any:
     if size_ <= ARRAYLET_SIZE: throw "OUT_OF_RANGE"
     full_arraylets := size_ / ARRAYLET_SIZE
-    left := size_ % ARRAYLET_SIZE
-    if left == 0:
+    remaining := size_ % ARRAYLET_SIZE
+    if remaining == 0:
       vector_ = Array_ full_arraylets
     else:
       vector_ = Array_ full_arraylets + 1
-      vector_[full_arraylets] = Array_ left filler
+      vector_[full_arraylets] = Array_ remaining filler
     full_arraylets.repeat:
       vector_[it] = Array_ ARRAYLET_SIZE filler
     // TODO(florian): remove this hack.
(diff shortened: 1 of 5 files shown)


lib/core/collections.toit
```diff
@@ -845,12 +845,8 @@ abstract class Array_ extends List:
   // Optimized helper method for iterating over the array elements.
   abstract do_ end/int [block] -> none
 
-  /** Create a new array, copying up to old_length elements from this array. */
-  resize_for_list_ old_length/int new_length/int -> Array_:
-    result := Array_ new_length
-    for i := 0; i < old_length and i < new_length; i++:
-      result[i] = this[i]
-    return result
+  /** Create a new array, copying up to $copy_length elements from this array. */
```
kasperl, line 848: Use /// for consistency?

Also why do we call this copy_length and new_length here and not copy_size and new_size?

lib/core/collections.toit
```diff
@@ -970,24 +970,31 @@ class LargeArray_ extends Array_:
...
-        if it * ARRAYLET_SIZE < old_size:
+        arraylet = Array_ limit filler
+        if pos < copy_size:
           old_arraylet := vector_[it]
-          arraylet.replace 0 old_arraylet 0 (old_size % ARRAYLET_SIZE)
-        left -= limit
+          arraylet.replace 0 old_arraylet 0 (min limit (copy_size - pos))
+      new_vector[it] = arraylet
+      left -= limit
+      pos += ARRAYLET_SIZE
```
kasperl, line 997: It feels odd to me that you're increasing/decreasing left and pos with what appears to be different amounts. Why not use limit consistently?

lib/core/collections.toit
```diff
@@ -970,24 +970,31 @@ class LargeArray_ extends Array_:
...
+  // An expansion or shrinking that uses the backing arraylets from the old
+  // array.  Used by List, Set, and Map.
+  resize_for_list_ copy_size/int new_size/int filler:
     // Rounding up division.
     number_of_arraylets := ((new_size - 1) / ARRAYLET_SIZE) + 1
-    new_vector := Array_ number_of_arraylets
+    if number_of_arraylets == 1:
+      return vector_[0].resize_for_list_ copy_size new_size filler
+    new_vector := Array_ number_of_arraylets  // new_vector may be a LargeArray_!
     left := new_size
```
kasperl, line 981: remaining?

lib/core/collections.toit
```diff
@@ -970,24 +970,31 @@ class LargeArray_ extends Array_:
...
+        arraylet = vector_[it]
+        if pos + limit > copy_size:
+          arraylet.fill --from=(max 0 (copy_size - pos)) --to=limit filler
       else:
-        limit := min left ARRAYLET_SIZE
-        arraylet := Array_ limit
-        new_vector[it] = arraylet
-        if it * ARRAYLET_SIZE < old_size:
+        arraylet = Array_ limit filler
+        if pos < copy_size:
```
kasperl, line 992: Is this check necessary and worth it?

lib/core/collections.toit
```diff
@@ -970,24 +970,31 @@ class LargeArray_ extends Array_:
...
+    pos := 0
     number_of_arraylets.repeat:
-      if (it + 1) * ARRAYLET_SIZE <= old_size:
-        new_vector[it] = vector_[it]
-        left -= ARRAYLET_SIZE
+      arraylet := ?
+      limit := min left ARRAYLET_SIZE
+      if pos + limit <= size:
+        arraylet = vector_[it]
+        if pos + limit > copy_size:
```
kasperl, line 988: Is this check necessary and worth it?

lib/core/collections.toit
```diff
@@ -2168,10 +2176,19 @@ abstract class HashedInsertionOrderedCollection_:
...
+      // Rebuild the index using the backing array.
+      // By using resize_for_list_ we reuse the arraylets when growing large
+      // arrays.  This reduces GC churn and, more importantly, peak memory
+      // usage.
+      if index_:
+        index_ = index_.resize_for_list_ /*copy_size=*/ 0 new_index_size /*filler=*/0
+        index_.fill 0
+      else:
+        index_ = Array_ new_index_size 0
+      if index_.size != new_index_size: throw "foo"
```
kasperl, line 2191: Should this be an assert?

src/primitive_core.cc
```diff
@@ -1601,17 +1601,16 @@ PRIMITIVE(array_at_put) {
...
 // the new one.
 PRIMITIVE(array_expand) {
-  ARGS(Array, old, word, old_length, word, length);
+  ARGS(Array, old, word, old_length, word, length, Object, filler);
   if (length == 0) return process->program()->empty_array();
   if (length < 0) OUT_OF_BOUNDS;
-  if (length > Array::max_length_in_process()) OUT_OF_RANGE;
-  if (old_length < 0 || old_length > old->length() || old_length > length) OUT_OF_RANGE;
-  Object* result = process->object_heap()->allocate_array(length, Smi::from(0));
+  if (length > Array::ARRAYLET_SIZE) OUT_OF_RANGE;
```
kasperl, line 1607: ARRAYLET_SIZE is 500 right now. Is that a good number? Can we make it much larger on 64-bit platforms where the page size is bigger?

kasperl (approved):
Looks good to me.

lib/core/collections.toit
erikcorry: replace doesn't allow negative values for the `to` argument, so yes.

lib/core/collections.toit
erikcorry: It's hard to measure a difference.  I took it out.

PR #1309, opened 2023-01-02
Fix getting of TCP window size on ESP32

erikcorry:


kasperl (approved):
Looks good to me. Should we test the values from Toit code?

PR #1311, opened 2023-01-02
Keep asserts when running the compiler for analysis.

floitsch:
We would also like to analyze asserts when compiling, but that's harder to do.


kasperl (approved):
Looks good to me.

You should probably try to fix the pkg-modbus issues (frame.id) and then update the gold files before submitting.

floitsch:
> Looks good to me.
> 
> You should probably try to fix the pkg-modbus issues (frame.id) and then update the gold files before submitting.

Yep. There are two packages that need to be fixed. 'modbus' and 'mqtt'. And then updating the packages that have it as dependency. On it.

PR #1312, opened 2023-01-03
Prune dispatch table entries for methods only called with INVOKE_STATIC

kasperl:


src/compiler/backend.cc
```diff
@@ -54,6 +59,28 @@ class BackendCollector : public ir::TraversingVisitor {
...
   }
 
+  void visit_CallStatic(ir::CallStatic* node) {
+    TraversingVisitor::visit_CallStatic(node);
+    // Some static calls target virtual methods. Some of those
+    // virtual methods are never called using a virtual call, so
+    // they only have a single entry in the dispatch table and
+    // no selector offset. For those methods, it is practical to
+    // be able to encode the set of classes that the method can
+    // be called on (the holder and all subclasses), so we extend
```
floitsch, line 69:
```suggestion
    // no selector offset. For those methods, we still would like
    // to know the set of classes that the method can
    // be called on (the holder and all subclasses), so we extend
```

floitsch (approved):
LGTM.

PR #1315, opened 2023-01-03
Avoid blocking interpreter threads when waiting for TLS handshaking

kasperl:


src/event_sources/tls.cc
```diff
@@ -67,10 +67,30 @@ void TlsEventSource::handshake(TlsSocket* socket) {
   OS::signal(sockets_changed_);
 }
 
+void TlsEventSource::close(TlsSocket* socket) {
+  { Locker locker(mutex());
+    for (TlsSocket* it : sockets_) {
```
kasperl, line 72: Is there a nicer way to do this? I don't think I can check if a socket is in the (single) linked list in any other way?

erikcorry: You could change it from LinkedList to LinkedFifo.  LinkedFifo is also singly linked, but keeps track of the last entry in the list.

See https://github.com/toitlang/toit/pull/1316

erikcorry (approved):

src/event_sources/tls.cc
kasperl: It already is a LinkedFifo! :)

PR #1319, opened 2023-01-03
Automatically update 'toit' AUR.

floitsch:


.github/workflows/ci.yml
```diff
@@ -734,3 +734,65 @@ jobs:
...
+        run: |
+          git clone ssh://aur@aur.archlinux.org/toit.git
+
+      - name: Git
+        run: |
+          cd toit
+          git config user.email "leon@toit.io"
+          git config user.name "Leon Gungadin Mogensen"
+
+      - name: Update AUR package
```
kasperl, line 775: This is more involved than for the Jaguar build where we use `updpkgsums: true` and the archlinux-package-action. Add a comment to explain why that doesn't work?

kasperl (approved):
Looks good to me.

.github/workflows/ci.yml
floitsch: done.

PR #1325, opened 2023-01-05
Fix signal-strength in wifi.

floitsch:
Thanks @rikke-rp.


floitsch:
@kasperl is there a way to write a test for this?

kasperl (approved):
Looks good to me.

PR #1326, opened 2023-01-05
Slightly less verbosity on the no-such-method error.

erikcorry:
When the problem is about the wrong number of unnamed args, there's no need to list optional named args to help the user.

Probably want to view this with the ignore-white-space option.


floitsch (approved):
LGTM.

PR #1329, opened 2023-01-08
Deprecate lib/services/wifi_leaker

kasperl:


tests/wifi_leaker_test_slow.toit
```diff
@@ -205,3 +205,434 @@ test_round_trip_x_percent_loss loss insertion:
     ctr++
   expect decoder.got
   expect decoder.msg == MSG
+
+// Each packet of data has a 5 bit index and 5 bits of datum.  There can be
```
kasperl, line 209: I think this is still a good test, so I decided to copy the code here. This can stick around even after getting rid of `lib/services/wifi_leaker.toit`.

floitsch (approved):
LGTM.

PR #1331, opened 2023-01-09
Add '--drop_empty' to 'string.split'

floitsch:


erikcorry (commented):
Maybe it's easier to wrap the block call.

erikcorry (approved):

PR #1339, opened 2023-01-13
Record input types for loads and invokes in type propagator

kasperl:


src/compiler/propagation/type_database.cc
```diff
@@ -209,15 +209,30 @@ std::string TypeDatabase::as_json() const {
...
       first = false;
     } else {
       out << ",\n";
     }
-    std::string type_string = it.second.as_json(program_);
-    out << "  {\"position\": " << it.first;
-    out << ", \"type\": " << type_string << "}";
+
+    int position = it.first;
+    std::string output_string = it.second.as_json(program_);
```
floitsch, line 220: feels like this is still the `usage_string` and not just an `output_string`.

floitsch (approved):
LGTM.

PR #1344, opened 2023-01-16
Fix open-drain.

floitsch:


lib/gpio/pin.toit
```diff
@@ -133,7 +133,12 @@ class Pin:
     state_ = monitor.ResourceState_ resource_group_ resource_
     if input or output:
       try:
-        configure --input=input --output=output --pull_down=pull_down --pull_up=pull_up
+        configure
+            --input=input
+            --output=output
+            --pull_down=pull_down
+            --pull_up=pull_up
+            --open_drain=open_drain
```
floitsch, line 141: We forgot to copy the open-drain parameter to the configure call.

tests/esp32/gpio.toit
```diff
@@ -6,14 +6,14 @@
...
-Connect pin 16 to pin 26, optionally with a 330 Ohm resistor to avoid short circuits.
+Connect pin 18 to pin 19, optionally with a 330 Ohm resistor to avoid short circuits.
 */
 
 import gpio
 import expect show *
 
-PIN1 ::= 16
-PIN2 ::= 26
+PIN1 ::= 18
```
floitsch, line 15: Changing the pins is not necessary.
I can change it back if you prefer.

kasperl (approved):
Looks good to me.

Nice catch.

PR #1346, opened 2023-01-16
Add sdk-version to snapshot bundle.

floitsch:


src/snapshot_bundle.cc
```diff
@@ -63,6 +64,14 @@ SnapshotBundle::SnapshotBundle(List<uint8> snapshot,
     if (status != 0) FATAL("Couldn't create snapshot");
   };
 
+  const char* sdk_version = vm_git_version();
```
kasperl, line 67: There is a `version_string` further down that you can get rid of now. Also a `version_length`.

kasperl (approved):
Looks good to me.

src/snapshot_bundle.cc
floitsch: done.

PR #1347, opened 2023-01-16
Fix 3 BLE bugs and enable compilation without CRYPTO

mikkeldamsgaard:
Fixing 3 errors in BLE. 
 - Init sometimes fails because the sequence of calls were wrong
 - Supporting more devices on peripheral would always fail
 - Connecting as a central would not wait until MTU was negotiation was complete before allowing service discovery, leading to a timeout

Fixing support for compiling without CRYPTO support.

### DIFF
--- a/src/primitive_crypto.cc
+++ b/src/primitive_crypto.cc
@@ -13,6 +13,10 @@
 // The license can be found in the file `LICENSE` in the top level
 // directory of this repository.
 
+#include "top.h"
+
+#if !defined(TOIT_FREERTOS) || CONFIG_TOIT_CRYPTO
+
 #include "mbedtls/gcm.h"
 #include "mbedtls/chachapoly.h"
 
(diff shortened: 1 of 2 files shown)


src/primitive_crypto.cc
```diff
@@ -534,3 +538,5 @@ PRIMITIVE(aes_ecb_close) {
 
 
 }
+
+#endif
```
kasperl, line 542: Add newline at end of the #endif line.

src/primitive_crypto.cc
```diff
@@ -13,6 +13,10 @@
 // The license can be found in the file `LICENSE` in the top level
 // directory of this repository.
 
+#include "top.h"
+
+#if !defined(TOIT_FREERTOS) || CONFIG_TOIT_CRYPTO
```
kasperl, line 18: For consistency?

```suggestion
#if !defined(TOIT_FREERTOS) || defined(CONFIG_TOIT_CRYPTO)
```

kasperl (approved):
Looks good to me.

PR #1349, opened 2023-01-17
Improve check for non-Toitware ESP-IDF fork.

floitsch:
On Windows the simple comparison didn't work.


kasperl (approved):
Looks good to me.

PR #1350, opened 2023-01-17
Minor Windows improvements.

floitsch:


tests/lsp/location_compiler_test_runner_base.toit
```diff
@@ -8,6 +8,11 @@ import reader show BufferedReader
 import .lsp_client show LspClient run_client_test
 import .utils
 
+is_absolute_ path/string -> bool:
+  if platform == PLATFORM_WINDOWS:
+    return path.size > 1 and path[1] == ':'
```
erikcorry, line 13: I wonder if sometimes we see unix-style paths on Windows too?
Perhaps Windows should have both tests.

erikcorry (approved):

tests/lsp/location_compiler_test_runner_base.toit
floitsch: This is only for our test, where we control the inputs (CMake), but done.

PR #1355, opened 2023-01-19
Trigger a GC when UDP allocations fail.

erikcorry:


src/resources/udp_esp32.cc
```diff
@@ -46,8 +46,22 @@ class Packet : public LinkedFifo<Packet>::Element {
     , addr_(addr)
     , port_(port) {}
 
+  Packet()
+    : pbuf_(null) {}
```
kasperl, line 50: 4 space indent here and above?

src/resources/udp_esp32.cc
```diff
@@ -46,8 +46,22 @@ class Packet : public LinkedFifo<Packet>::Element {
...
+    pbuf_ = null;
+  }
+
+  void set(struct pbuf* pbuf, ip_addr_t addr, u16_t port) {
+    pbuf_ = pbuf;
+    addr_ = addr;
+    port_ = port;
   }
 
   struct pbuf* pbuf() { return pbuf_; }
```
kasperl, line 67: Make all these const?

src/resources/udp_esp32.cc
```diff
@@ -66,12 +80,15 @@ class UdpSocket : public Resource {
   UdpSocket(ResourceGroup* group, udp_pcb* upcb)
     : Resource(group)
     , upcb_(upcb)
-    , buffered_bytes_(0) {}
+    , buffered_bytes_(0) {
+    spare_packet_ = _new Packet();
```
kasperl, line 84: Maybe add a comment that it is okay if this fails?

src/resources/udp_esp32.cc
```diff
@@ -111,6 +135,7 @@ class UdpSocket : public Resource {
  private:
   udp_pcb* upcb_;
   LinkedFifo<Packet> packets_;
+  Packet* spare_packet_;
```
kasperl, line 138: Add a comment that explains why we cache this. What do we save by doing that?

kasperl (approved):
Looks good to me.

PR #1357, opened 2023-01-20
Expose names for bundled container images

kasperl:
It is a little bit inconvenient that the names of dynamically installed container images, do not show up. I'll think about a good way of getting them included, but it is hard to modify the system image assets and there are downsides (mostly size) to encoding the names of the images as assets associated with the individual images.

Jaguar already maintains names for the dynamically installed ones. Not sure if we can use that map somehow or if we need to find a way to provide a name when container images are installed.

### DIFF
--- a/lib/system/api/containers.toit
+++ b/lib/system/api/containers.toit
@@ -44,9 +44,9 @@ class ContainerServiceClient extends ServiceClient implements ContainerService:
 
   list_images -> List:
     array := invoke_ ContainerService.LIST_IMAGES_INDEX null
-    return List array.size / 3:
-      cursor := it * 3
-      ContainerImage (uuid.Uuid array[cursor]) array[cursor + 1] array[cursor + 2]
+    return List array.size / 4:
+      cursor := it * 4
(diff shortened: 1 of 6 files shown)


kasperl:
Looks like this needs a bit more work (failing test).

lib/system/api/containers.toit
```diff
@@ -47,9 +47,9 @@ class ContainerServiceClient extends ServiceClient implements ContainerService:
 
   list_images -> List:
     array := invoke_ ContainerService.LIST_IMAGES_INDEX null
-    return List array.size / 3:
-      cursor := it * 3
-      ContainerImage (uuid.Uuid array[cursor]) array[cursor + 1] array[cursor + 2]
+    return List array.size / 4:
+      cursor := it * 4
+      ContainerImage (uuid.Uuid array[cursor]) array[cursor + 1] array[cursor + 2] array[cursor + 3]
```
floitsch, line 52: Would be nice if `ContainerImage` used named parameters.
That would immediately tell us what all these values are.

floitsch (approved):
LGTM.

lib/system/api/containers.toit
kasperl: Done.

PR #1358, opened 2023-01-22
Remove RPC filesystem.

floitsch:
This was a protocol we used with Toit v1. The 'toit' executable would proxy LSP calls to the LSP server and take over reading of the file system. This was necessary, as the Toit LSP server wasn't yet able to read files on Windows.


kasperl (approved):
Looks good to me.

PR #1359, opened 2023-01-22
Fix canonicalize on Windows.

floitsch:
It would trim the trailing '\' of `c:\`.

### DIFF
--- a/src/compiler/filesystem.cc
+++ b/src/compiler/filesystem.cc
@@ -141,13 +141,14 @@ void Filesystem::canonicalize(char* path) {
       path[canonical_pos++] = path[i++];
     }
   }
-  // Drop trailing path seperator.
-  // There can only be one.
-  if (path[canonical_pos - 1] == path_separator()) {
+  // Drop trailing path seperator unless it's the root.
+  path[canonical_pos] = '\0';
+  if (!(is_absolute && is_root(path)) && path[canonical_pos - 1] == path_separator()) {
+    // There can only be one.
     canonical_pos--;
   }
   if (canonical_pos == 0) {
-    path[canonical_pos++] = is_absolute ? path_separator() : '.';
+    path[canonical_pos++] = '.';
   }
   path[canonical_pos] = '\0';
 }
--- a/src/compiler/filesystem_local_win.cc
+++ b/src/compiler/filesystem_local_win.cc
@@ -67,6 +67,16 @@ char* FilesystemLocal::root(const char* path) {
   return strdup("\\\\");
 }
 
+bool FilesystemLocal::is_root(const char* path) {
+  // Something like "c:\".
+  if (path[1] == ':') {
+    return path[0] != '\n' && path[1] == ':' && path[2] == '\\' && path[3] == '\0';
+  }
+  // A network path like '\\Machine1'.
+  return path[0] == '\\' && path[1] == '\\' && path[2] == '\0';
+}
+
+
 char* FilesystemLocal::to_local_path(const char* path) {
   if (path == null) return null;
   char* result = strdup(path);
--- a/src/compiler/filesystem.h
+++ b/src/compiler/filesystem.h
@@ -52,6 +52,9 @@ class Filesystem {
     result[1] = '\0';
     return result;
   }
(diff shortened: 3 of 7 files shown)


src/compiler/filesystem.cc
```diff
@@ -141,13 +141,14 @@ void Filesystem::canonicalize(char* path) {
       path[canonical_pos++] = path[i++];
     }
   }
-  // Drop trailing path seperator.
-  // There can only be one.
-  if (path[canonical_pos - 1] == path_separator()) {
+  // Drop trailing path seperator unless it's the root.
+  path[canonical_pos] = '\0';
+  if (!(is_absolute && is_root(path)) && path[canonical_pos - 1] == path_separator()) {
```
kasperl, line 146: This confuses me. Can `is_absolute` be false and still allow `is_root` to return true?

src/compiler/filesystem.cc
```diff
@@ -141,13 +141,14 @@ void Filesystem::canonicalize(char* path) {
...
+    // There can only be one.
     canonical_pos--;
   }
   if (canonical_pos == 0) {
-    path[canonical_pos++] = is_absolute ? path_separator() : '.';
+    path[canonical_pos++] = '.';
```
kasperl, line 151: So here `is_absolute` cannot be true, right? Assert that?

src/compiler/filesystem.cc
```diff
@@ -141,13 +141,14 @@ void Filesystem::canonicalize(char* path) {
...
-  // Drop trailing path seperator.
-  // There can only be one.
-  if (path[canonical_pos - 1] == path_separator()) {
+  // Drop trailing path seperator unless it's the root.
+  path[canonical_pos] = '\0';
+  if (!(is_absolute && is_root(path)) && path[canonical_pos - 1] == path_separator()) {
+    // There can only be one.
     canonical_pos--;
   }
   if (canonical_pos == 0) {
```
kasperl, line 150: Can this happen after stripping the last separator? That feels weird.

kasperl (approved):
Looks good to me.

src/compiler/filesystem.cc
```diff
@@ -141,13 +141,14 @@ void Filesystem::canonicalize(char* path) {
       path[canonical_pos++] = path[i++];
     }
   }
-  // Drop trailing path seperator.
-  // There can only be one.
-  if (path[canonical_pos - 1] == path_separator()) {
+  // Drop trailing path seperator unless it's the root.
```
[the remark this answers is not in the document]
floitsch: done.

src/compiler/filesystem.cc
floitsch: No. it's just to avoid to do the is_root check when we already know that the path isn't even absolute.
I removed the check.
done.

src/compiler/filesystem.cc
floitsch: Moved the `canonical_pos == 0` check before the 'is_root` check and made the `is_root` check an 'else'.
This also avoids accessing the path at `-1` which could otherwise happen.
done.

src/compiler/filesystem_local_win.cc
```diff
@@ -67,6 +67,16 @@ char* FilesystemLocal::root(const char* path) {
   return strdup("\\\\");
 }
 
+bool FilesystemLocal::is_root(const char* path) {
+  // Something like "c:\".
+  if (path[1] == ':') {
```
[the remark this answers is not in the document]
floitsch: In theory we were only calling this with absolute paths, in which case it was safe, but I agree.
Way too brittle.
Changed it to test the length first.

src/compiler/filesystem_local_win.cc
```diff
@@ -67,6 +67,16 @@ char* FilesystemLocal::root(const char* path) {
...
+    return path[0] != '\n' && path[1] == ':' && path[2] == '\\' && path[3] == '\0';
+  }
+  // A network path like '\\Machine1'.
+  return path[0] == '\\' && path[1] == '\\' && path[2] == '\0';
```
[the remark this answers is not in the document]
floitsch: done.

PR #1362, opened 2023-01-22
Handle compiler paths better.

floitsch:
When running the compiler in LSP mode, it uses '/' as signal that a path is absolute. Enforce that.


kasperl (approved):
Looks good to me.

PR #1371, opened 2023-01-23
Introduce relative_anchor function.

floitsch:
On Windows a relative path can be either relative to the current working directory (or another file), or relative to the drive.

### DIFF
--- a/src/compiler/filesystem.h
+++ b/src/compiler/filesystem.h
@@ -48,27 +48,23 @@ class Filesystem {
   // The path the non-absolute path is relative to.
   // On Posix systems this is equal to `cwd`.
   // On Windows, it can be `cwd`, or a drive (like "c:"), if the path starts with '\' or '/'.
-  virtual const char* relative_anchor(const char* path) {
-    return cwd();
-  }
-  virtual char path_separator() { return '/'; }
-  // On Windows both '/' and '\\' are path separators. It's thus not
+  virtual const char* relative_anchor(const char* path) = 0;
+  // The default path separator.
+  // On Posix systems equal to '/'.
+  // On Windows set to '\\' (although '/' also works).
+  virtual char path_separator() = 0;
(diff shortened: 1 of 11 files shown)


src/compiler/filesystem.h
```diff
@@ -45,11 +45,25 @@ class Filesystem {
   virtual List<const char*> package_cache_paths() = 0;
 
   virtual bool is_absolute(const char* path) = 0;
+  // The path the non-absolute path is relative to.
+  // On Posix systems this is equal to `cwd`.
+  // On Windows, it can be `cwd`, or a drive (like "c:"), if the path starts with '\' or '/'.
+  virtual const char* relative_anchor(const char* path) {
+    return cwd();
```
kasperl, line 52: As discussed, having these defined in the super class is a source of confusion. Maybe just drop them here?

src/compiler/filesystem_local_win.cc
```diff
@@ -62,19 +62,33 @@ static int root_prefix_length(const char* path) {
...
     return 3;  // Drive root.
   }
   return 0;
 }
 
+const char* FilesystemLocal::relative_anchor(const char* path) {
+  ASSERT(!is_absolute(path));
+  if (path[0] == '/') {  // Safe to access even for empty strings.
+    // TODO(florian): error checking.
+    DWORD result_length = GetFullPathName("/", 0, NULL, NULL);
```
kasperl, line 82: Can this be cached in any meaningful way? Looks like a ton of allocations.

kasperl (approved):
Looks good to me.

src/compiler/filesystem_local_win.cc
floitsch: I would like to revisit this area again.
The current setup (mostly?) fixes the immediate issues, but I hope to find a better API.

PR #1372, opened 2023-01-23
Fix absolute path in primitives.

floitsch:


src/primitive_file_win.cc
```diff
@@ -108,27 +108,42 @@ const char* current_dir(Process* process) {
...
-      (pathname_length > 2 && pathname[1] == ':' && (pathname[2] == '\\' || pathname[2] == '/'))) {
+  if (!PathIsRelative(pathname)) {
     if (GetFullPathName(pathname, MAX_PATH, output, NULL) == 0) WINDOWS_ERROR;
+    return null;
+  }
+
+  const char* current_directory = current_dir(process);
+  if (!current_directory) MALLOC_FAILED;
+
+  // Check if the path is rooted.
```
kasperl, line 123: Explain what rooted is?

kasperl (approved):
Looks good to me.

src/primitive_file_win.cc
floitsch: done.

PR #1373, opened 2023-01-24
Break down the components of Toit memory use in OOM message.

erikcorry:
The same message is produced by the
serial_print_heap_report function.

### DIFF
--- a/src/heap.h
+++ b/src/heap.h
@@ -34,10 +34,12 @@ class InitialMemoryManager {
  public:
   Chunk* initial_chunk = null;
   Object** global_variables = null;
+  Mutex* heap_mutex = null;
 
   void dont_auto_free() {
     initial_chunk = null;
     global_variables = null;
+    heap_mutex = null;
   }
 
   // Allocates initial pages for heap.  Returns success.
@@ -49,7 +51,7 @@ class InitialMemoryManager {
 
 class ObjectHeap {
  public:
-  ObjectHeap(Program* program, Process* owner, Chunk* initial_chunk, Object** global_variables);
+  ObjectHeap(Program* program, Process* owner, Chunk* initial_chunk, Object** global_variables, Mutex* mutex);
   ~ObjectHeap();
 
   // TODO: In the new heap there need not be a max allocation size.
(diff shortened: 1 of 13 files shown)


src/os_esp32.cc
```diff
@@ -601,6 +604,9 @@ class HeapSummaryPage {
     return "unknown";
   }
 
+  Process* get_toit_process() const { return owning_process_; }
```
kasperl, line 607: What's up with the `get_` prefix?

src/os_esp32.cc
```diff
@@ -550,6 +552,7 @@ class HeapSummaryPage {
     memset(void_cast(counts_), 0, sizeof counts_);
     users_ = 0;
```
kasperl, line 553: Why not initialize these at the declaration sites?

src/os_esp32.cc
```diff
@@ -659,37 +668,80 @@ class HeapSummaryCollector {
     counts_[type]++;
   }
 
+  void identify_toit_processes() {
```
kasperl, line 671: I'd try to be more consistent with the naming. Not sure the `toit_` prefix helps much here.

src/os_esp32.cc
kasperl, line 607: I'd try to be more consistent with the naming. Not sure the toit_ prefix helps much here. Maybe just call this `owning_process`?

src/heap.h
```diff
@@ -213,6 +212,7 @@ class ObjectHeap {
   Object** global_variables_ = null;
 
   HeapRootList external_roots_;
+  Mutex* mutex_;
```
kasperl, line 215: Maybe add a comment that explains what this is used to guard against?

src/third_party/dartino/object_memory.cc
```diff
@@ -295,7 +302,7 @@ void ObjectMemory::set_up() {
   spare_chunk_ = allocate_chunk(null, TOIT_PAGE_SIZE);
   if (!spare_chunk_) FATAL("Can't allocate initial spare chunk");
   if (spare_chunk_mutex_) FATAL("Can't call ObjectMemory::set_up twice");
```
kasperl, line 304: We should probably check that we can allocate the spare chunk mutex. It feels a bit weird that it is on the mutex we check that we cannot call set_up twice.

kasperl (approved):
Looks good to me.

src/os_esp32.cc
erikcorry: Page objects get reused when we increment dropped_pages_ so this code is not only run in the constructor.

PR #1376, opened 2023-01-25
Fork environment

erikcorry:


src/utils.cc
```diff
@@ -409,18 +409,47 @@ word Utils::utf_16_to_8(const uint16* input, word length, uint8* output, word ou
...
+
+  // Quick out for different first ASCII letter.
+  if ((input1[0] <= MAX_ASCII || input2[0] <= MAX_ASCII) && input1[0] != input2[0]) return false;
+
+  // Start with length comparison of the UTF-16 version.
+  if (length2 != utf_8_to_16(input1, length1, null, 0)) return false;
+
+  // Now we know the UTF-16 versions are the same length, generate the UTF-16
+  // version of the UTF-8 input, and compare them.
+  uint16* wide_input1 = unvoid_cast<uint16*>(malloc(sizeof(uint16) * length2));
```
floitsch, line 432: Maybe add a TODO that this could be done without allocation?

src/resources/pipe_posix.cc
```diff
@@ -451,6 +475,21 @@ PRIMITIVE(fork) {
   do {
     if (command == null) break;
 
+    for (int i = 0; environment && i < environment->length(); i += 2) {
+      Blob key;
+      environment->at(i)->byte_content(process->program(), &key, STRINGS_ONLY);
+      auto key_cstr = strndup(char_cast(key.address()), key.length());
```
floitsch, line 481: Do we not need to free the object?

src/resources/pipe_posix.cc
```diff
@@ -451,6 +475,21 @@ PRIMITIVE(fork) {
...
+      Object* value = environment->at(i + 1);
+      if (value == process->program()->null_object()) {
+        unsetenv(key_cstr);
+      } else {
+        Blob value_blob;
+        value->byte_content(process->program(), &value_blob, STRINGS_ONLY);
+        auto value_cstr = strndup(char_cast(value_blob.address()), value_blob.length());
```
floitsch, line 488: Same: Do we not need to free the value?

src/utils.cc
```diff
@@ -312,6 +314,207 @@ bool Utils::is_valid_utf_8(const uint8* buffer, int length) {
 #endif
 }
 
+// Assumes the input is valid UTF-8, eg from a Toit string.
```
floitsch, line 317:
```suggestion
// Assumes the input is valid UTF-8, for example from a Toit string.
```

floitsch (approved):
LGTM.

src/resources/pipe_posix.cc
erikcorry: No, this is between fork and exec, so no leaks are possible.

floitsch: Please add a comment.

src/primitive_file_win.cc
```diff
@@ -405,111 +403,104 @@ PRIMITIVE(stat) {
...
 }
 
 PRIMITIVE(rename) {
-  ARGS(cstring, old_name, cstring, new_name);
-  int result = rename(old_name, new_name);
+  ARGS(StringOrSlice, old_name_blob, StringOrSlice, new_name_blob);
+  WideCharAllocationManager allocation1(process), allocation2(process);
+  wchar_t* old_name = allocation1.to_wcs(&old_name_blob);
+  wchar_t* new_name = allocation2.to_wcs(&new_name_blob);
+  // TODO(erik): Should this use absolute path?
```
floitsch, line 427: Yes.
Since we change current working directory without the system knowing about it, we can't give the system relative paths.
Add a test, where we change directory and then move a file.

src/primitive_file_win.cc
```diff
@@ -405,111 +403,104 @@ PRIMITIVE(stat) {
...
 
-  char path[MAX_PATH];
-  auto error = get_absolute_path(process, pathname, path);
-  if (error) return error;
-
-  char* copy = strdup(path);
-  if (!copy) MALLOC_FAILED;
+  wchar_t* copy = wcsdup(path);
 
   process->set_current_directory(copy);
```
floitsch, line 440: Should we check that the path exists? Feels weird to change to a directory that can't even work.

src/primitive_file_win.cc
```diff
@@ -405,111 +403,104 @@ PRIMITIVE(stat) {
...
   return process->program()->null_object();
 }
 
 PRIMITIVE(mkdtemp) {
-  ARGS(cstring, prefix);
+  ARGS(StringOrSlice, prefix_blob);
   DWORD ret;
 
+  WideCharAllocationManager allocation(process);
+  wchar_t* prefix = allocation.to_wcs(&prefix_blob);
```
floitsch, line 459: We can't give the system relative paths.
Once we know that the path isn't `/tmp/` we need to make it absolute.

src/primitive_file_win.cc
```diff
@@ -92,65 +92,68 @@ static const int FILE_ST_ATIME = 8;
...
-  char temp[MAX_PATH];
-  if (snprintf(temp, MAX_PATH, "%s\\%s", relative_to, pathname) >= MAX_PATH) INVALID_ARGUMENT;
-  if (GetFullPathName(temp, MAX_PATH, output, NULL) == 0) WINDOWS_ERROR;
+  wchar_t temp[MAX_PATH];
+  if (snwprintf(temp, MAX_PATH, L"%ls\\%ls", relative_to, pathname) >= MAX_PATH) INVALID_ARGUMENT;
+  if (GetFullPathNameW(temp, MAX_PATH, output, NULL) == 0) WINDOWS_ERROR;
   return null;
 }
 
+#define BLOB_TO_ABSOLUTE_PATH(result, blob)                        \
```
floitsch, line 147: Add a comment, that almost every function should use this macro, as we use a per-process working directory. As such we can never give a relative path to the system function.

On Posix, we change the paths already in Toit land (I guess to avoid allocations). Maybe/probably we should add checks in all primitives that the incoming paths are indeed absolute.

floitsch (approved):
Still LGTM.

PR #1377, opened 2023-01-25
Fix load-global-var bytecode.

floitsch:
The wide variant wasn't immediately following the non-wide one.


kasperl (approved):
Looks good to me.

kasperl (approved):
Looks good to me.

But you will need to fix the bytecode table here: https://github.com/toitlang/toit/blob/master/tools/snapshot.toit#L785.

floitsch:
Done, and added a test.

kasperl (commented):
Still looks good!

PR #1383, opened 2023-01-27
Use fewer linux builders

floitsch:
Some of the builders couldn't start immediately, because we sharded out the normal tests too much. By reducing it a bit, the 'cross' and 'esp32' workflows can now start at the same time as all the other builders.


kasperl (approved):
Looks good to me.

PR #1385, opened 2023-01-28
Fix `null` vs null_object in i2c read.

floitsch:


tests/esp32/bad_i2c.toit
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2023 Toitware ApS.
+// Use of this source code is governed by a Zero-Clause BSD license that can
+// be found in the tests/LICENSE file.
+
+/**
+Tests the errors of i2c.
+
+Setup:
+pin 32 and pin 33 should not be connected.
```
kasperl, line 9:
```suggestion
Pin 32 and pin 33 should not be connected.
```

kasperl (approved):
Looks good to me.

PR #1386, opened 2023-01-29
Add Task.divide

kasperl:


lib/core/task.toit
```diff
@@ -75,6 +76,90 @@ interface Task:
...
     return Task_.current
 
+  /**
+  Divides the work given by $lambdas up into a number of
+    separate tasks that run concurrently.
+
+  Returns the results of running the $lambdas as a $Map
+    keyed by the lambda index. If the lambda at a given
+    index did not return, the results map will not contain
+    and entry for it. The results map is insertion ordered,
```
floitsch, line 86: I generally prefer not to use future tense, but it would work here.
Up to you.

```suggestion
    index doesn't return, the results map doesn't contain
    and entry for it. The results map is insertion ordered,
```

lib/core/task.toit
```diff
@@ -75,6 +76,90 @@ interface Task:
...
+    so it is possible to tell which lambda returned first.
+
+  If any of $lambdas throw an exception, the exception is
```
floitsch, line 89:
```suggestion
  If any of the $lambdas throws an exception, the exception is
```

lib/core/task.toit
```diff
@@ -75,6 +76,90 @@ interface Task:
...
+    propagated to the caller of $Task.divide which in
+    returns also throws.
+
+  If $required is less than the number of $lambdas, the
+    method returns when $required tasks have completed.
+  */
+  static divide lambdas/List -> Map
```
floitsch, line 96: ?
```suggestion
  static run_in_parallel lambdas/List -> Map
```

lib/core/task.toit
```diff
@@ -75,6 +76,90 @@ interface Task:
...
+      --required/int=lambdas.size:
+    count ::= lambdas.size
+    tasks ::= Array_ count
+    results ::= {:}
+    if count < 2 or not (1 <= required <= count):
+      throw "Bad Argument"
```
floitsch, line 102: We usually use "INVALID_ARGUMENT", but we need to go through all 'throw' places anyway, so I don't mind too much.
```suggestion
      throw "INVALID_ARGUMENT"
```

lib/core/task.toit
floitsch, line 96: The main issue I have with this API is that it is hard to give additional properties to the individual lambdas.

I could imagine that the following related scenarios would be useful:
- run in parallel, and restart tasks that fail
- ignore failed tasks completely
- trace failed tasks, but don't stop the others

If we didn't provide a lambda, but an instance, then we could add these flags to them:
```
Task.run_in_parallel [
  XX --ignore_failure:: do_something,
  XX --keep_running --trace_errors:: ...,
]
```

Maybe, if we decide to adopt `|>`:
```
results := Task.in_parallel
|> %.run --keep_running:: do_something
|> %.run --ignore_error:: do_something_else
|> %.run ...
|> %.wait_for_results
```

The latter could already be implemented now:
```
scheduler := Task.in_parallel
scheduler.run --ignore_failure:: ...
scheduler.run --keep_running:: ...
results := scheduler.wait_for_results
```

Looking at it, we could probably always add the more advanced version later, as it doesn't have the same function shape as this implementation.

floitsch (approved):
LGTM.

lib/core/task.toit
kasperl: The three suggestions are all achievable (quite easily) from within the individual lambdas. It is essentially just variants of `catch` used in them.

lib/core/task.toit
kasperl: Maybe I picked the wrong place for inspiration :)

lib/core/task.toit
kasperl: I thought about such names, but it does much more than just run *concurrently*. It is the basis for a lot of divide and conquer operations, so I thought about things like split, join, or merge, but I couldn't think of something much better than raw `divide` to indicate that you're dividing one task up into many dependent ones.

PR #1394, opened 2023-01-31
Use resolv.conf to find name servers on Posix

erikcorry:
On Posix we keep an eye on the /etc/resolv.conf file and
reparse it when it changes to get the currrent name servers.
This is also the way golang does it, and glibc does something
similar.

The reason I made a new primitive is that otherwise we
would end up copying chunks of the file.toit into the core
(currently in pkg-host), which would bloat the VM more on
ESP32 because we can't tree shake based on platform.  Since
the file reading code is in C++ it doesn't affect the size on ESP32
as much.

If we get smarter at tree shaking in a platform-dependent
way we should probably make this more a Toit than C++
implementation and clean it up.

Still to do: Something for ESP32 that sets the default DNS
resolver from the DHCP data.

### DIFF
--- a/lib/net/modules/dns.toit
+++ b/lib/net/modules/dns.toit
@@ -57,7 +57,55 @@ DEFAULT_CLIENT ::= DnsClient [
 // are separated by slash (/) in the key.
 ALL_CLIENTS_ ::= {DEFAULT_CLIENT.servers_.join "/": DEFAULT_CLIENT}
 
-default_client := DEFAULT_CLIENT
+user_set_client_/DnsClient? := null
+dhcp_client_/DnsClient? := null
+
+/**
+On Unix systems the default client is one that keeps an eye on changes in
+  /etc/resolv.conf.
+On FreeRTOS systems the default client is set by DHCP.
+On Windows we currently default to using Google and Cloudflare DNS servers.
+On all platforms you can set a custom default client with the
+  $(default_client=) setter.
+*/
+default_client -> DnsClient:
+  if user_set_client_: return user_set_client_
+  if platform == "FreeRTOS" and dhcp_client_: return dhcp_client_
+  if platform == "Linux" or platform == "macOS": return etc_resolv_client_
+  return DEFAULT_CLIENT
+
+default_client= client/DnsClient -> none:
+  user_set_client_ = client
+
+current_etc_resolv_client_/DnsClient? := null
+etc_resolv_update_time_/Time? := null
+
+etc_resolv_client_ -> DnsClient:
+  error := catch --trace:
+    resolv_conf_stat := stat_ "/etc/resolv.conf" true
+    etc_stat := stat_ "/etc" true
+    resolv_conf_time := Time.epoch --ns=resolv_conf_stat[ST_MTIME_]
+    etc_time := Time.epoch --ns=etc_stat[ST_MTIME_]
+    modification_time := resolv_conf_time > etc_time ? resolv_conf_time : etc_time
+    if etc_resolv_update_time_ == null or etc_resolv_update_time_ < modification_time:
+      etc_resolv_update_time_ = modification_time
+      // Create a new client from resolv.conf.
+      resolv_conf := (read_file_content_posix_ "/etc/resolv.conf" resolv_conf_stat[ST_SIZE_]).to_string
+      nameservers := []
+      resolv_conf.split "\n": | line |
+        if line.starts_with "nameserver ":
+          nameservers.add line[11..].copy
+      current_etc_resolv_client_ = DnsClient nameservers
+  return current_etc_resolv_client_ or DEFAULT_CLIENT
+
+ST_SIZE_ ::= 7
+ST_MTIME_ ::= 9
+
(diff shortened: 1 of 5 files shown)


src/compiler/propagation/type_primitive_file.cc
```diff
@@ -38,6 +38,7 @@ TYPE_PRIMITIVE_ANY(mkdtemp)
 TYPE_PRIMITIVE_ANY(is_open_file)
 TYPE_PRIMITIVE_ANY(realpath)
 TYPE_PRIMITIVE_ANY(cwd)
+TYPE_PRIMITIVE_ANY(read_file_content_posix)
```
kasperl, line 41: As far as I can tell, this always returns a byte array?

lib/net/modules/dns.toit
```diff
@@ -57,7 +57,60 @@ DEFAULT_CLIENT ::= DnsClient [
...
+On Unix systems the default client is one that keeps an eye on changes in
+  /etc/resolv.conf.
+On FreeRTOS systems the default client is set by DHCP.
+On Windows we currently default to using Google and Cloudflare DNS servers.
+On all platforms you can set a custom default client with the
+  $(default_client= client) setter.
+*/
+default_client -> DnsClient:
+  if user_set_client_: return user_set_client_
+  if platform == "FreeRTOS" and dhcp_client_: return dhcp_client_
```
kasperl, line 73: Use the PLATFORM_ constants for this matching.

lib/net/modules/dns.toit
```diff
@@ -57,7 +57,60 @@ DEFAULT_CLIENT ::= DnsClient [
...
+  if platform == "Linux" or platform == "macOS": return etc_resolv_client_
+  return DEFAULT_CLIENT
+
+default_client= client/DnsClient -> none:
```
kasperl, line 77: Should you be able to clear this again?

```suggestion
default_client= client/DnsClient? -> none:
```

lib/net/modules/dns.toit
```diff
@@ -57,7 +57,60 @@ DEFAULT_CLIENT ::= DnsClient [
...
+      resolv_conf := (read_file_content_posix_ "/etc/resolv.conf" resolv_conf_stat[ST_SIZE_]).to_string
+      nameservers := []
+      resolv_conf.split "\n": | line |
+        hash := line.index_of "#"
+        if hash >= 0: line = line[..hash]
+        line = line.trim
+        if line.starts_with "nameserver ":
+          server := line[11..].trim.copy
+          if net.IpAddress.is_valid server --accept_ipv4=true --accept_ipv6=false:
+            nameservers.add server
```
kasperl, line 102: Maybe delay copying to here? It feels like you're only doing it to not store slices in the store.

lib/net/modules/dns.toit
```diff
@@ -57,7 +57,60 @@ DEFAULT_CLIENT ::= DnsClient [
...
+
+default_client= client/DnsClient -> none:
+  user_set_client_ = client
+
+current_etc_resolv_client_/DnsClient? := null
+etc_resolv_update_time_/Time? := null
+
+etc_resolv_client_ -> DnsClient:
+  error := catch --trace:
+    resolv_conf_stat := stat_ "/etc/resolv.conf" true
```
kasperl, line 85: Put the path in a constant?

src/primitive_file_win.cc
```diff
@@ -563,6 +563,11 @@ PRIMITIVE(cwd) {
   return result;
 }
 
+PRIMITIVE(read_file_content_posix) {
+  // This is currenly only used for resolv.conf.
```
kasperl, line 567:
```suggestion
  // This is currenly only used for /etc/resolv.conf.
```

lib/net/modules/dns.toit
```diff
@@ -57,7 +57,60 @@ DEFAULT_CLIENT ::= DnsClient [
...
+  return DEFAULT_CLIENT
+
+default_client= client/DnsClient -> none:
+  user_set_client_ = client
+
+current_etc_resolv_client_/DnsClient? := null
+etc_resolv_update_time_/Time? := null
+
+etc_resolv_client_ -> DnsClient:
+  error := catch --trace:
```
kasperl, line 84: What do you use the error for? I guess you want to make sure you always get a DnsClient back.

lib/net/modules/dns.toit
```diff
@@ -57,7 +57,60 @@ DEFAULT_CLIENT ::= DnsClient [
 // are separated by slash (/) in the key.
 ALL_CLIENTS_ ::= {DEFAULT_CLIENT.servers_.join "/": DEFAULT_CLIENT}
```
kasperl, line 58: I'm not sure I follow how this ALL_CLIENTS_ stuff works. You're adding new clients to this on demand, but you seed it with a pretty specific single entry which is:

8.8.8.8/1.1.1.1/8.8.4.4/1.0.0.1 -> DEFAULT_CLIENT

Does anyone ever ask for that?

kasperl (approved):

PR #1395, opened 2023-02-01
Get DNS server from DHCP on ESP32

erikcorry:


erikcorry:
This doesn't work because it only sets the DNS server in the process that brought up the Wifi, not in the other processes.

lib/net/modules/dns.toit
```diff
@@ -155,8 +155,14 @@ class DnsClient:
...
   constructor servers/List:
-    if servers.size == 0 or (servers.any: it is not string): throw "INVALID_ARGUMENT"
-    servers_ = servers
+    if servers.size == 0 or (servers.any: it is not string and it is not net.IpAddress): throw "INVALID_ARGUMENT"
+    servers_ = List servers.size:
+      server := servers[it]
+      if server is string:
+        net.IpAddress.parse server
+      else:
+        server
```
kasperl, line 164:
```suggestion
    servers_ = servers.map: (it is string) ? net.IpAddress.parse it : it
```

src/resources/wifi_esp32.cc
```diff
@@ -57,14 +57,25 @@ class WifiResourceGroup : public ResourceGroup {
...
+  static const int NUMBER_OF_ADDRESSES = 3;
 
-  void set_ip_address(uint32 address) { ip_address_ = address; }
-  void clear_ip_address() { ip_address_ = 0; }
+  uint32 ip_address(int index) const { return ip_address_[index]; }
+  bool has_ip_address() const { return ip_address_[OWN_ADDRESS] != 0; }
+
+  void set_ip_address(int index, uint32 address) { ip_address_[index] = address; }
+  void clear_ip_addresses() {
+    for (int i = 0; i < NUMBER_OF_ADDRESSES; i++) {
```
kasperl, line 74: memset?

system/extensions/esp32/wifi.toit
```diff
@@ -248,15 +254,34 @@ class WifiModule implements NetworkModule:
...
 
   wait_for_static_ip_address_ -> none:
-    ip ::= wifi_get_ip_ resource_group_
+    ip ::= wifi_get_ip_ resource_group_ OWN_ADDRESS_INDEX_
     address_ = net.IpAddress ip
     logger_.info "network address statically assigned" --tags={"ip": address_}
+    configure_dns_ --from_dhcp=false
+
+  configure_dns_ --from_dhcp/bool -> none:
+    main_dns := wifi_get_ip_ resource_group_ MAIN_DNS_ADDRESS_INDEX_
```
kasperl, line 270: Maybe this should return null instead of NULL_IP? It would be simpler to deal with here.

system/extensions/esp32/wifi.toit
```diff
@@ -248,15 +254,34 @@ class WifiModule implements NetworkModule:
...
+    if backup_dns != NULL_IP: dns_ips.add (net.IpAddress backup_dns)
+    if dns_ips.size != 0:
+      dns.dhcp_client_ = dns.DnsClient dns_ips
+      if from_dhcp:
+        logger_.info "DNS server address dynamically assigned through dhcp" --tags={"dns": dns_ips}
+      else:
+        logger_.info "DNS server statically configured" --tags={"dns": dns_ips}
+    else:
+      dns.dhcp_client_ = null
+      logger_.info "No DNS server supplied by network.  Using fallback DNS servers."
```
kasperl, line 284:
```suggestion
      logger_.info "DNS server not supplied by network; using fallback DNS servers"
```

kasperl (approved):
Looks good to me modulo the missing service extension.

PR #1396, opened 2023-02-02
Deprecate at and xmodem_1k libraries

kasperl:


lib/at/locker.toit
```diff
@@ -7,6 +7,8 @@ import .session
 /**
 The locker can wrap a $Session, and ensure both exclusive access to the
   Session and a queue in case of concurrent requests.
+
+Deprecated.
```
floitsch, line 11: Provide information on what should/can be used instead. Same for the others.

floitsch (approved):
LGTM.
Maybe wait with committing, until the alternatives are available (unless they already are).

lib/at/locker.toit
kasperl: There are no obvious alternatives available.

PR #1399, opened 2023-02-03
Windows: Handle relative mkdtemp correctly

erikcorry:


src/primitive_file_win.cc
```diff
@@ -107,7 +107,7 @@ const wchar_t* current_dir(Process* process) {
   return current_directory;
 }
 
-HeapObject* get_absolute_path(Process* process, const wchar_t* pathname, wchar_t* output) {
+HeapObject* get_absolute_path(Process* process, const wchar_t* pathname, wchar_t* output, const wchar_t* used_for_relative = null) {
```
floitsch, line 110:
```suggestion
HeapObject* get_absolute_path(Process* process, const wchar_t* pathname, wchar_t* output, const wchar_t* relative_to = null) {
```

floitsch (approved):
LGTM.

PR #1400, opened 2023-02-03
Use a single macro in the ARGS line of primitives to indicate a Windows path

erikcorry:
Instead of using argv[0] to get the executable.


floitsch (approved):
LGTM.

PR #1405, opened 2023-02-04
Fix uninitialized access.

floitsch:
Also add a valgrind test to the buildbot.


tests/valgrind/run.cmake
```diff
@@ -0,0 +1,50 @@
...
+  ERROR_VARIABLE STDERR
+  RESULT_VARIABLE EXIT_CODE
+)
+
+file(READ ${VALGRIND_XML} VALGRIND_OUTPUT)
+
+# Extract all lines of the form '<kind>...</kind>'.
+string(REGEX MATCHALL "<kind>[^<]*</kind>" VALGRIND_ERRORS "${VALGRIND_OUTPUT}")
+
+# Filter out lines that contain "Leak_".
```
kasperl, line 44: Why do we have to do this? If we want to filter out leaks, wouldn't it be better to valgrind to not look for leaks? (show-leak-kinds=none or something like that).

It would probably be ideal to have tests that run from snapshots where we expect no leaks, and tests that run from source (and include the compiler) where we expect no weird uninitialized access.

kasperl (approved):
Looks good to me.

How did you find this?

tests/valgrind/run.cmake
floitsch: done.

PR #1411, opened 2023-02-06
Prune device library

kasperl:


lib/device.toit
```diff
@@ -2,172 +2,87 @@
...
-  return impl.Device_.instance.estimate_time_accuracy
+hardware_id/uuid.Uuid ::= uuid.uuid5 "hw_id" get_mac_address_
 
 /** Simple key-value store. */
 interface Store:
-  get name/string
-  delete name/string
-  set name/string value/any
+  get name/string -> any
+  delete name/string -> none
```
floitsch, line 22: We need to decide whether we want to call these "delete", or "remove".
I think we use the term "remove" in the core Map.

floitsch (approved):
LGTM.

PR #1413, opened 2023-02-07
Clarify imports in net/impl.toit

erikcorry:


lib/net/impl.toit
```diff
@@ -8,8 +8,8 @@ import .udp as udp
 import .socket_address
 
 import .modules.dns as dns
```
kasperl, line 10:
```suggestion
import .modules.dns as dns_module
```

kasperl (approved):
Looks good to me.

PR #1415, opened 2023-02-07
Clarify more imports of network modules.

erikcorry:
Also, don't use the low level network modules in the LSP server.
Instead use regular net.open, which picks the default network
  interface.


tests/ethernet_compiles_test.toit
```diff
@@ -0,0 +1,7 @@
+// Copyright (C) 2023 Toitware ApS.
+// Use of this source code is governed by a Zero-Clause BSD license that can
+// be found in the tests/LICENSE file.
+
+import net.ethernet
+
+main:
```
kasperl, line 7: Maybe test something really basic here with the expect library?

kasperl (approved):
Looks good to me.

PR #1416, opened 2023-02-07
Uart esp32 rewrite

mikkeldamsgaard:
@erikcorry Go nuts.


src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+      }
+      memcpy(reinterpret_cast<uint8_t*>(&transfer_header_) + read, header_data, header_received);
+      // No blocking calls ever on TxBuffer, so ignore last parameter
+      vRingbufferReturnItemFromISR(ring_buffer(), header_data, null);
+      read += header_received;
+    } while (read < sizeof(TxTransferHeader));
+  }
+
+  size_t maximum_to_receive = max_length;
+  if (transfer_header_.remaining_data_length_ < maximum_to_receive) maximum_to_receive = transfer_header_.remaining_data_length_;
```
erikcorry, line 322:
```suggestion
  size_t maximum_to_receive = Utils::min(max_length, transfer_header_.remaining_data_length_);
```

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+}
+
+size_t TxBuffer::free_size() {
+  size_t ringbuffer_free = xRingbufferGetCurFreeSize(ring_buffer());
+  if (ringbuffer_free < sizeof(TxTransferHeader)) return 0;
+  return ringbuffer_free - sizeof(TxTransferHeader);
+}
+
+void UART_ISR_INLINE RxBuffer::send(uint8_t* buffer, uint32_t length) {
+  // We are never blocking on the ring buffer
```
erikcorry, line 340:
```suggestion
  // We are never blocking on the ring buffer.
```

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+}
+
+UART_ISR_INLINE uint8_t* TxBuffer::read(size_t *received, size_t max_length) {
+  IsrSpinLocker locker(&spinlock_);
+
+  if (transfer_header_.remaining_data_length_ == 0 && transfer_header_.break_length_ == 0) {
+    // No more data in the latest package. Need to read header
+    size_t header_received;
+    size_t read = 0;
+    do { // As the header is multiple bytes, it could be split over the edge of the ring buffer
```
erikcorry, line 308:
```suggestion
    do {  // As the header is multiple bytes, it could be split over the edge of the ring buffer.
```

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+      void* header_data = xRingbufferReceiveUpToFromISR(ring_buffer(), &header_received, sizeof(TxTransferHeader) - read);
+      if (!header_data) {
+        // TxBuffer is empty
```
erikcorry, line 311:
```suggestion
        // TxBuffer is empty.
```

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+        return null;
+      }
+      memcpy(reinterpret_cast<uint8_t*>(&transfer_header_) + read, header_data, header_received);
+      // No blocking calls ever on TxBuffer, so ignore last parameter
```
erikcorry, line 315:
```suggestion
      // No blocking calls ever on TxBuffer, so ignore last parameter.
```

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+  uint8_t break_length = transfer_header_.break_length_;
+  transfer_header_.break_length_ = 0; // To avoid it being read again and the break continuing...
+  return break_length;
+}
+
+UART_ISR_INLINE uint8_t* TxBuffer::read(size_t *received, size_t max_length) {
+  IsrSpinLocker locker(&spinlock_);
+
+  if (transfer_header_.remaining_data_length_ == 0 && transfer_header_.break_length_ == 0) {
+    // No more data in the latest package. Need to read header
```
erikcorry, line 305:
```suggestion
    // No more data in the latest package. Need to read header.
```

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
 
-MODULE_IMPLEMENTATION(uart, MODULE_UART);
+  uart()->enable_interrupt_index(UART_TOIT_INTR_TXFIFO_EMPTY);
+}
+
+UART_ISR_INLINE uint8_t TxBuffer::read_break() {
+  IsrSpinLocker locker(&spinlock_);
+
+  uint8_t break_length = transfer_header_.break_length_;
+  transfer_header_.break_length_ = 0; // To avoid it being read again and the break continuing...
```
erikcorry, line 297:
```suggestion
  transfer_header_.break_length_ = 0;  // To avoid it being read again and the break continuing...
```

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+  }
+}
+
+UART_ISR_INLINE uart_event_types_t UartResource::interrupt_handler_write() {
+  uint32_t tx_fifo_free = get_tx_fifo_available_count();
+  size_t received_count;
+  uint8_t* buffer = tx_buffer().read(&received_count, tx_fifo_free);
+  if (buffer) {
+    uint32_t written;
+    uart_toit_hal_write_txfifo(hal_, buffer, received_count, &written); // TODO: Is the RS485 workaround necessary here?
```
erikcorry, line 492:
```suggestion
    uart_toit_hal_write_txfifo(hal_, buffer, received_count, &written);  // TODO: Is the RS485 workaround necessary here?
```

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+  if (event != UART_EVENT_MAX) {
+    portBASE_TYPE hp_task_awoken = 0;
+    uart->send_event_to_queue(event, &hp_task_awoken);
+    if (hp_task_awoken) portYIELD_FROM_ISR();
+  }
+}
+
+UART_ISR_INLINE void UartResource::send_event_to_queue(uart_event_types_t event, int* hp_task_awoken) {
+  IsrSpinLocker locker(&spinlock_);
+  // Data and Tx Event receive special care, as to not overflow the queue
```
erikcorry, line 545:
```suggestion
  // Data and Tx Event receive special care, so as to not overflow the queue.
```

src/resources/uart_esp32.cc
```diff
@@ -119,135 +608,230 @@ PRIMITIVE(init) {
...
+  uint16_t rx_buffer_size, tx_buffer_size;
+  int interrupt_flags = ESP_INTR_FLAG_IRAM | ESP_INTR_FLAG_SHARED;
+  if ((options & 8) != 0) {
+    // High speed setting.
+    interrupt_flags |= ESP_INTR_FLAG_LEVEL3;
+    full_interrupt_threshold = 35;
+    tx_buffer_size = 1024;
+    rx_buffer_size = 2048;
+  } else if ((options & 4) != 0) {
+    // Medium speed
```
erikcorry, line 676:
```suggestion
    // Medium speed setting.
```

src/resources/uart_esp32.cc
```diff
@@ -119,135 +608,230 @@ PRIMITIVE(init) {
...
+                                         init.tx_buffer, tx_buffer_size);
+  if (!init.uart_resource) {
+    uart_init_cleanup(init);
+    MALLOC_FAILED;
   }
 
+  periph_module_enable(uart_periph_signal[port].module);
+    // Workaround for ESP32C3: enable core reset
+    // before enabling uart module clock
+    // to prevent uart output garbage value.
```
erikcorry, line 737:
```suggestion
    // to prevent uart from outputting garbage value.
```

src/resources/uart_esp32.cc
```diff
@@ -119,135 +608,230 @@ PRIMITIVE(init) {
...
+      .irq = uart_periph_signal[port].irq,
+      .interrupt_flags = interrupt_flags,
+      .uart_resource = init.uart_resource,
+      .err = ESP_OK
+  };
 
-  if (err == ESP_OK) {
-    err = uart_set_mode(port, static_cast<uart_mode_t>(mode));
-  }
+  // Install the ISR on the SystemEventSource's main thread that runs on core 0, to allocate the interrupts on core 0
```
erikcorry, line 813:
```suggestion
  // Install the ISR on the SystemEventSource's main thread that runs on core 0, to allocate the interrupts on core 0.
```

src/resources/uart_esp32.cc
```diff
@@ -119,135 +608,230 @@ PRIMITIVE(init) {
...
+    uart_init_cleanup(init);
+    return Primitive::os_error(args.err, process);
   }
-  group->register_resource(res);
 
-  proxy->set_external_address(res);
+  init.uart_resource->set_interrupt_handle(args.intr_handle);
 
-  uart_flush_input(port);
+  // Enable read interrupts
```
erikcorry, line 827:
```suggestion
  // Enable read interrupts.
```

src/resources/uart_esp32.cc
```diff
@@ -257,117 +841,69 @@ PRIMITIVE(create_path) {
...
-  esp_err_t err = uart_wait_tx_done(uart->port(), 0);
-  if (err == ESP_ERR_TIMEOUT) {
-    return BOOL(false);
-  }
-  if (err != ESP_OK) {
-    return Primitive::os_error(err, process);
-  }
+  if (!uart->tx_buffer().is_empty()) return BOOL(false);
 
+  // Busy wait for the fifo to become empty
```
erikcorry, line 890:
```suggestion
  // Busy wait for the fifo to become empty.
```

src/resources/uart_esp32_hal.c
```diff
@@ -0,0 +1,155 @@
...
+
+void uart_toit_hal_set_rx_timeout(uart_hal_handle_t hal, uint8_t timeout) {
+  uart_hal_set_rx_timeout(HAL, timeout);
+}
+
+#if SOC_UART_REQUIRE_CORE_RESET
+void uart_toit_hal_set_reset_core(uart_hal_handle_t hal, bool reset) {
+  uart_hal_set_reset_core(HAL, reset);
+}
+#endif // SOC_UART_REQUIRE_CORE_RESET
```
erikcorry, line 97:
```suggestion
#endif  // SOC_UART_REQUIRE_CORE_RESET
```

src/resources/uart_esp32_hal.h
```diff
@@ -0,0 +1,84 @@
...
+#endif
+  
+typedef enum {
+  UART_TOIT_INTR_RXFIFO_FULL  = 0, // The RxFifo is over its threshold
+  UART_TOIT_INTR_TXFIFO_EMPTY,     // The TxFifo is under its threshold
+  UART_TOIT_INTR_PARITY_ERR,       // Parity error
+  UART_TOIT_INTR_RXFIFO_OVF,       // RxFifo overflow, not emptied fast enough
+  UART_TOIT_INTR_TX_BRK_DONE,      // Transfer break completed
+  UART_TOIT_INTR_RX_TIMEOUT,       // The RxFifo has data, not over its threshold, but some time since last byte
+  UART_TOIT_ALL_INTR_MASK,         // All interrupt bits
```
erikcorry, line 36:
```suggestion
  UART_TOIT_INTR_RXFIFO_FULL  = 0,  // The RxFifo is over its threshold.
  UART_TOIT_INTR_TXFIFO_EMPTY,      // The TxFifo is under its threshold.
  UART_TOIT_INTR_PARITY_ERR,        // Parity error.
  UART_TOIT_INTR_RXFIFO_OVF,        // RxFifo overflow, not emptied fast enough.
  UART_TOIT_INTR_TX_BRK_DONE,       // Transfer break completed.
  UART_TOIT_INTR_RX_TIMEOUT,        // The RxFifo has data, not over its threshold, but some time since last byte.
  UART_TOIT_ALL_INTR_MASK,          // All interrupt bits.
```

lib/uart.toit
```diff
@@ -240,6 +231,18 @@ class Port implements reader.Reader:
       if flushed: return
       sleep --ms=1
 
+  write_no_wait_ data from=0 to=data.size --break_length=0:
+    while true:
+      s := state_.wait_for_state WRITE_STATE_ | ERROR_STATE_
+      if not uart_: throw "CLOSED"
+      written := uart_write_ uart_ data from to break_length
+      if written < to - from:
+        // Not every thing was written, clear write flag and try again
```
erikcorry, line 240:
```suggestion
        // Not everything was written, clear write flag and try again.
```

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
-  uint32_t on_event(Resource* r, word data, uint32_t state) {
-    switch (data) {
-      case UART_DATA:
-        state |= kReadState;
-        break;
+  uint32_t on_event(Resource* r, word data, uint32_t state) override;
+};
+
+UART_ISR_INLINE void RxTxBuffer::return_buffer(uint8_t* buffer) {
+  // No blocking calls ever on TxBuffer, so ignore last parameter
```
erikcorry, line 268:
```suggestion
  // No blocking calls ever on TxBuffer, so ignore last parameter.
```

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+  if (ringbuffer_free < sizeof(TxTransferHeader)) return 0;
+  return ringbuffer_free - sizeof(TxTransferHeader);
+}
+
+void UART_ISR_INLINE RxBuffer::send(uint8_t* buffer, uint32_t length) {
+  // We are never blocking on the ring buffer
+  xRingbufferSendFromISR(ring_buffer(), buffer, length, null);
+}
+
+void RxBuffer::read(uint8_t* buffer, uint32_t length) {
```
erikcorry, line 344: Do I understand it right that we only call this if there is enough data in the ring buffer?  So the reason for the loop is that there might be a wraparound, so that two reads are needed.  Please add a comment that says something like this (if it's correct).

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+  size_t received;
+  size_t read = 0;
+  do {
+    auto from_ring_buffer = static_cast<uint8_t*>(xRingbufferReceiveUpTo(ring_buffer(), &received, 0, length - read));
```
erikcorry, line 348: Is there a missing call to vRingbufferReturnItem here?

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+
+UartResource::~UartResource() {
+  disable_interrupt_index(UART_TOIT_ALL_INTR_MASK);
+  if (interrupt_handle_) esp_intr_free(interrupt_handle_);
+
+  vQueueDelete(queue());
+
+  // Empty tx buffer.
+  uint32_t len;
+  do {
```
erikcorry, line 368: Not sure this is good.  It will hang for an indefinite time, no?

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+
+void UartResource::clear_interrupt_index(uart_toit_interrupt_index_t index) {
+  uart_toit_hal_clr_intsts_mask(hal_, interrupt_mask(index));
+}
+
+void UartResource::clear_interrupt_mask(uint32_t mask) {
+  uart_toit_hal_clr_intsts_mask(hal_, mask);
+}
+
+bool UartResource::try_set_iomux_pin(gpio_num_t pin, uint32_t iomux_index) const {
```
erikcorry, line 457: What is this for, and what does the return value mean?

erikcorry (commented):
This looks great, but it's a bit big, so I'm not done reviewing yet.

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+  size_t read_length = uart_toit_hal_get_rxfifo_len(hal_);
+  if (read_length == 0) return UART_EVENT_MAX;
+
+  size_t rx_buffer_free = rx_buffer_.free_count();
+  if (rx_buffer_free == 0) {
+    disable_read_interrupts();
+    return UART_BUFFER_FULL;
+  } else {
+    if (rx_buffer_free < read_length) read_length = rx_buffer_free;
+    uint8_t buffer[read_length];
```
erikcorry, line 479: This is an alloca in an interrupt handler, which makes me very nervous.  What are the bounds on the size of this, and how big is an interrupt handler stack on ESP-IDF?  There is a new call in the pipeline that can help us avoid this issue: https://github.com/espressif/esp-idf/issues/10527

In the mean time perhaps a temp buffer in the resource is best?  Or perhaps a fixed size on-stack buffer with a loop that does uart_toit_hal_read_rxfifo and rx_buffer().send (aka xRingbufferSendFromISR) until it is finished?

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+  }
+}
+
+IRAM_ATTR void UartResource::interrupt_handler(void* arg) {
+  auto uart = unvoid_cast<UartResource*>(arg);
+  uint32_t uart_interrupt_status = uart_toit_hal_get_intsts_mask(uart->hal_);
+  if (uart_interrupt_status == 0) return;
+  uart_event_types_t event = UART_EVENT_MAX;
+
+  if (uart_interrupt_status & uart->interrupt_mask(UART_TOIT_INTR_TXFIFO_EMPTY)) {
```
erikcorry, line 514: I think the calls to interrupt_mask will return the same every time, and perhaps at some point (not now) we should cache them to improve the performance of this dispatcher.

Are we sure that only one can be true?  Should we loop until all bits have been handled?

src/resources/uart_esp32.cc
```diff
@@ -119,135 +608,230 @@ PRIMITIVE(init) {
...
 
+typedef struct {
+  QueueHandle_t queue;
+  uart_hal_handle_t hal;
+  bool hardware_initialized;
+  UartResource* uart_resource;
+  uart_port_t port;
+  uint8_t* rx_buffer;
+  uint8_t* tx_buffer;
+} uart_initialization_record_t;
```
erikcorry, line 619: I think it's better to make this a class (with public fields) and convert uart_init_cleanup to a destructor that does not have to be called explicitly.

src/resources/uart_esp32.cc
```diff
@@ -119,135 +608,230 @@ PRIMITIVE(init) {
...
+    uart_init_cleanup(init);
+    MALLOC_FAILED;
+  }
+
+  init.uart_resource = _new UartResource(group, port, init.queue, init.hal,
+                                         init.rx_buffer, rx_buffer_size,
+                                         init.tx_buffer, tx_buffer_size);
+  if (!init.uart_resource) {
+    uart_init_cleanup(init);
+    MALLOC_FAILED;
```
erikcorry, line 731: Here you would add a call to the init object that flips a flag so that the destructor does nothing if called after this point.

erikcorry (commented):
Looking very good.

src/resources/uart_esp32.cc
```diff
@@ -257,117 +841,69 @@ PRIMITIVE(create_path) {
...
-  }
-  if (wrote == -1) {
-    OUT_OF_RANGE;
+  size_t available = uart->tx_buffer().free_size();
+  if (available < size) {
+    size = static_cast<int>(available);
+    break_length = 0;
   }
 
+  if (size == 0) return Smi::from(0);
```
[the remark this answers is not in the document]
mikkeldamsgaard: That is actually a very good question. This implies that the user is trying to send an empty byte array with a break length. That would also need some changes to the toit code. I think for now we can assume that breaks will be following data?

src/resources/uart_esp32.cc
```diff
@@ -17,95 +17,584 @@
...
+  uint32_t baud_rate;
+  uart_toit_hal_get_baudrate(hal_, &baud_rate);
+  return baud_rate;
+}
+
+void UartResource::set_tx_pin(gpio_num_t tx_pin) {
+  if (!try_set_iomux_pin(tx_pin, SOC_UART_TX_PIN_IDX)) {
+    gpio_hal_iomux_func_sel(GPIO_PIN_MUX_REG[tx_pin], PIN_FUNC_GPIO);
+    gpio_set_level(tx_pin, 1);
+    esp_rom_gpio_connect_out_signal(tx_pin, UART_PERIPH_SIGNAL(port_, SOC_UART_TX_PIN_IDX), false, false);
```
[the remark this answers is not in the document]
mikkeldamsgaard: No, see set_line_inverse

src/resources/uart_esp32.cc
mikkeldamsgaard: I "fixed" it, however Utils::min requires the arguments to be identical types, so I had to do: 
```
uword maximum_to_receive = Utils::min(max_length, static_cast<uword>(transfer_header_.remaining_data_length_));
```

src/resources/uart_esp32.cc
mikkeldamsgaard: I tried to add an appropriate comment. But it basicaly tries to connect the pin via the direct IO mux (for higher speed). If that does not work, the individual pin configurations needs to go via the GPIO matrix (slower speeds). For UART it probably does not matter at all. But the driver in the esp-idf does it.

src/resources/uart_esp32.cc
mikkeldamsgaard: As this is interacting directly with the uart controller it will actually complete quickly. It depends on the baud rate. The max size is 128 bytes, so for 9600 baud it empties in a little over a second, worst case.
The reason that this is here, is that if you for example use jaguar and deploys a new version while a transmit is in progress, it will not work at all. I could not find a better way to do it, due to hardware limitations.

src/resources/uart_esp32.cc
mikkeldamsgaard: The max size is 128 bytes. It has been tested at very high speeds (3 Mbit), so I am not worried at all.

src/resources/uart_esp32.cc
mikkeldamsgaard: Multiple interrupts can be set at once. However, since we do not clear the interrupt flag, the handle will just be called multiple times. I found it more readable this way. It might be interesting to explore what is faster.

src/resources/uart_esp32.cc
erikcorry: Is it not limited by the max ring buffer size, 2048?

erikcorry: Ah, no I see it's limited by the HAL. Perhaps insert an assert here that the size is no more than 256?

src/resources/uart_esp32.cc
```diff
@@ -119,135 +610,230 @@ PRIMITIVE(init) {
...
+  ~UartInitialization() {
+    if (keep) return;
+
+    if (!uart_resource) {
+      if (queue) vQueueDelete(queue);
+      if (hal) uart_toit_hal_deinit(hal);
+      if (rx_buffer) free(rx_buffer);
+      if (tx_buffer) free(tx_buffer);
+    } else {
+      delete uart_resource;
```
erikcorry, line 624: I think this causes a use-after-free because the resource is stil linked in the linked list of the ResourceGroup. The correct way to delete is to call unregister_resource on the group.

mikkeldamsgaard: If keep is false, then the resource is not yet registered with the group.

erikcorry: Ah yes, I was thinking of SimpleResource and its subclasses, which all register with the group in the constructor.

erikcorry:
Needs a merge up to master.

mikkeldamsgaard:
> 

Done

erikcorry (approved):

PR #1419, opened 2023-02-09
esp32s3 improved support

mikkeldamsgaard:
The esp32s3 comes in three versions. Without spi-ram, with quad spi-ram and with octo-spiram. The first two can use the same config file, but the last octo one needs its own. I added a esp32s3octo toolchain to differentiate the two and changed the esp32s3 one to be the quad (most common and cheapest).


floitsch (approved):
LGTM.

floitsch:
Thanks again.

PR #1424, opened 2023-02-14
Fixes for esp32c3

mikkeldamsgaard:
On the ESP32-C3 (risc v5 architecture) the uint32_t and uint32 were not equivalent types in the eyes of the C++ compiler and therefore some virtual functions where not dispatched correctly.

The main change is to shift all the typedef's for ints and uints in top.h to use the stdint versions int8_t, etc.

A few other places had warnings/errors and they have been addressed.


floitsch (approved):
LGTM!

PR #1425, opened 2023-02-15
Better handling of errors when there is no source map.

floitsch:


tools/snapshot.toit
```diff
@@ -191,15 +193,19 @@ class SnapshotBundle:
     if magic_content.to_string != MAGIC_CONTENT: return false
     return true
 
+  has_source_map -> bool: return source_map != null
```
kasperl, line 196: For consistency.

```suggestion
  has_source_map -> bool: 
    return source_map != null
```

kasperl (approved):
Looks good to me.

PR #1428, opened 2023-02-16
Fix for gitversion when there isn't any release branch.

floitsch:


floitsch:
Diff gets tiny if whitespace changes are disabled.

kasperl (approved):
Looks good to me.

PR #1430, opened 2023-02-21
Disabling interrupts for pins that do not need interrupts.

mikkeldamsgaard:


src/resources/gpio_esp32.cc
```diff
@@ -155,6 +155,10 @@ void GpioResourceGroup::on_register_resource(Resource* r) {
   gpio_num_t pin = static_cast<gpio_num_t>(static_cast<GpioResource*>(r)->pin());
   SystemEventSource::instance()->run([&]() -> void {
     FATAL_IF_NOT_ESP_OK(gpio_isr_handler_add(pin, isr_handler, reinterpret_cast<void*>(pin)));
+    // The gpio_isr_handler_add also enables interrupts on the pin. This is undesirable as all changes to the
+    // pin will then call the ISR. In toit we pass the Pin object around to other peripherals, and therefore we
```
kasperl, line 159:
```suggestion
    // pin will then call the ISR. In Toit, we pass the Pin object around to other peripherals, and therefore we
```

kasperl (approved):
Looks good to me.

PR #1431, opened 2023-02-21
Make sure the queue set is emptied at the same rate as the queues it contains

mikkeldamsgaard:


src/event_sources/ev_queue_esp32.cc
```diff
@@ -69,33 +73,42 @@ void EventQueueEventSource::entry() {
...
       // Wait for any queue/semaphore to wake up.
-      xQueueSelectFromSet(queue_set_, portMAX_DELAY);
+      handle = xQueueSelectFromSet(queue_set_, portMAX_DELAY);
     }
 
-    // First test if we should shut down.
-    if (xSemaphoreTake(stop_, 0)) {
-      return;
-    }
+    // The handle is now the queue/semaphore that has woken up. Remove at most one event from the underlying queues,
```
erikcorry, line 82: Please limit block comments to 80 chars in width, so they are more readable.

erikcorry (approved):
Thankyou!  Great find!

PR #1432, opened 2023-02-21
Make wait_for_state_ always either yield or await

mikkeldamsgaard:


lib/core/events_.toit
```diff
@@ -49,9 +49,14 @@ monitor ResourceState_:
     // into a situation, where timeouts might be ignored.
     super
 
+  // This method will always either yeild or wait.
   wait_for_state_ bits:
-    result := null
+    result := state_ & bits
     if not resource_: return 0
+    if result != 0:
+      // For cooporation, any resource that already has its state satisfied yields.
```
kasperl, line 57:
```suggestion
    if not resource_: return 0
    result := state_ & bits
    if result != 0:
      // For co-operation, waiting for a resource that already has its state 
      // satisfied yields in the calling task.
```

kasperl (commented):
I'm not sure I fully understand this change. The `wait` and `wait_for_state` methods are monitor methods which means that we will yield when we leave them (unless they are called from a critical region).

Are we calling `wait_for_state_` directly in a place where the suggested new yield makes a difference?

lib/core/events_.toit
mikkeldamsgaard:
```monitor M:
 run:
  await: true

main:
  task::
    i := 0
    while true:
      print i++
      yield

  m := M
  while true:
    m.run
```
I do not see the run method of the monitor yielding when leaving it?

kasperl: Thanks for sharing the snippet! The problem is a lack of 'message processing', so we never realize that resolving the `PrintService` is done. The monitor code sees that no other task is ready to run, so it assumes that we should just continue in main. 

Let me think about the best way of fixing this. It is sort of an optimization gone wrong and just yielding and getting messages processed would solve it.

kasperl: https://github.com/toitlang/toit/blob/ea412081e5dbf5af23e443878e1ee08e85d9fb24/lib/core/monitor_impl_.toit#L1

```
if is_non_critical: yield
```

mikkeldamsgaard:
The issue that was being solved was not an issue after all

PR #1433, opened 2023-02-21
Default major to 1 and minor to 0 for all new services

mikkeldamsgaard:


kasperl:
Hi @mikkeldamsgaard. I believe what you're trying to achieve with this is good (when there is relatively close coupling between client and service it sucks to have to mess with version numbers), but maybe we can do better. I'll keep you posted if my idea works out.

PR #1435, opened 2023-02-22
Bigger metadata mappings

erikcorry:


src/os.cc
```diff
@@ -151,13 +151,26 @@ static void* try_grab_aligned(void* suggestion, uword size) {
 
 OS::HeapMemoryRange OS::single_range_ = { 0 };
 
+// Protected by the resource mutex.
+static const int RECENTLY_FREED_SIZE = 10000;
```
kasperl, line 155: Add a comment about this choice? How many do we typically see used when running benchmarks or other things?

kasperl (approved):
Looks good to me.

PR #1438, opened 2023-02-24
On windows, delete read-only files as well.

floitsch:
The `_wunlink` function fails on read-only files. One has to remove the read-only property first.


kasperl (approved):
Looks good to me.

PR #1442, opened 2023-02-26
Don't assume that separators are '\' on Windows.

floitsch:
It is legal to use '/' as well.

### DIFF
--- a/tools/firmware.toit
+++ b/tools/firmware.toit
@@ -506,18 +506,17 @@ flash parsed/cli.Parsed -> none:
   firmware_bin := extract_binary envelope --config_encoded=config_encoded
   binary := Esp32Binary firmware_bin
 
-  separator := ?
   bin_extension := ?
+  bin_name := program_name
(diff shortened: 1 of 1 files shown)


tools/firmware.toit
```diff
@@ -506,18 +506,17 @@ flash parsed/cli.Parsed -> none:
   firmware_bin := extract_binary envelope --config_encoded=config_encoded
   binary := Esp32Binary firmware_bin
 
-  separator := ?
   bin_extension := ?
+  bin_name := program_name
   if platform == PLATFORM_WINDOWS:
-    separator = "\\"
+    bin_name.replace --all "\\" "/"
```
kasperl, line 512: I'm surprised by this. Is replace in-place?

floitsch: It's not...
I moved the code around to make it nicer, and the forgot to assign it...
done.

kasperl (approved):
Looks good to me.

PR #1444, opened 2023-02-27
Fix string interpolation type inference.

floitsch:
When a string interpolation started with the interpolated expression and didn't have any other interpolation we were not checking that the result of calling `stringify` was a string.

We also didn't correctly infer that the type of the expression is a string.


kasperl (approved):
Looks good to me.

PR #1445, opened 2023-02-27
Don't use 'FATAL_IF_NOT_ESP_OK' in espnow code

floitsch:
It's almost never the correct approach to reboot if something doesn't go right. Many of the cases in the espnow library can be influenced by the user and should thus give an error message.


kasperl (approved):
Looks good to me.

PR #1450, opened 2023-03-01
Unify ram buckets

kasperl:


kasperl:
This looks like a big change, but it really is just a refactoring that splits the storage implementation into a couple of files and unifies the functionality, so we use the same code for device and host targets. 

On host targets, we now simulate the RTC memory using a static array. This means that we can test the (slightly broken) implementation of RAM buckets more easily.

system/storage/region.toit
```diff
@@ -0,0 +1,120 @@
...
+    revoke
+
+  static open provider/StorageServiceProvider client/int -> List
+      --path/string
+      --capacity/int?:
+    registry := provider.registry
+    id := uuid.uuid5 NAMESPACE path
+    allocation := find_allocation_ registry --id=id --if_absent=:
+      if not capacity: throw "FILE_NOT_FOUND"
+      new_allocation_ registry --id=id --path=path --size=capacity + FLASH_REGISTRY_PAGE_SIZE
```
floitsch, line 57: Add a comment explaining why we need the extra FLASH_REGISTRY_PAGE_SIZE

system/storage/region.toit
```diff
@@ -0,0 +1,120 @@
...
+        Region.MODE_WRITE_CAN_CLEAR_BITS_
+    ]
+
+  static delete registry/FlashRegistry -> none
+      --path/string:
+    id := uuid.uuid5 NAMESPACE path
+    allocation := find_allocation_ registry --id=id --if_absent=: return
+    offset := allocation.offset + FLASH_REGISTRY_PAGE_SIZE
+    size := allocation.size - FLASH_REGISTRY_PAGE_SIZE
+    if flash_is_accessed_ offset size: throw "ALREADY_IN_USE"
```
floitsch, line 76: `STILL_IN_USE`, or something similar sounds more correct?

floitsch (approved):
LGTM.

system/storage/region.toit
kasperl: ALREADY_IN_USE is a "standard" exception we generate from primitives too. I think I prefer sticking to that.

system/storage/region.toit
kasperl: Done.

PR #1452, opened 2023-03-01
Fix RAM storage buckets and improve testing

kasperl:


system/storage/bucket.toit
```diff
@@ -15,92 +15,93 @@
...
   set key/string value/ByteArray -> none:
-    flash_kv_write_bytes_ group (compute_path_ key) value
+    flash_kv_write_bytes_ group (compute_id_ key) value
 
   remove key/string -> none:
-    flash_kv_delete_ group (compute_path_ key)
-
-  compute_path_ key/string -> string:
-    return paths.get key --init=: (uuid.uuid5 root key).stringify[..13]
+    flash_kv_delete_ group (compute_id_ key)
```
floitsch, line 64: In theory it could make sense to keep a delete counter and clear the `ids_` when we have done a lot of removals and have a big set.

floitsch (approved):
LGTM.

system/storage/bucket.toit
kasperl: Yeah, or just cap the number of cached ids -- or the number of unused cached ids. For another PR :)

PR #1455, opened 2023-03-02
Fix BLE can't send large packet which is larger than 255 bytes.

donghengqaz:


donghengqaz:
@mikkeldamsgaard @floitsch PTAL.

src/resources/ble_esp32.cc
```diff
@@ -233,6 +233,21 @@ class BleReadWriteElement : public BleErrorCapableResource {
     return unvoid_cast<BleReadWriteElement*>(arg)->_on_access(ctxt);
   }
 
+  static uint16_t mbuf_total_len(os_mbuf* om) {
+    uint16_t total_len = om->om_len;
```
kasperl, line 237: I'd prefer using int for these even though I understand that it isn't going to be bigger than 64k.

```suggestion
  static int mbuf_total_len(os_mbuf* om) {
    int total_len = om->om_len;
```

src/resources/ble_esp32.cc
```diff
@@ -233,6 +233,21 @@ class BleReadWriteElement : public BleErrorCapableResource {
...
+    om = SLIST_NEXT(om, om_next);
+    while (om) {
+      total_len += om->om_len;
+      om = SLIST_NEXT(om, om_next);
+    }
+
+    return total_len;
+  }
+
+  uint16_t send_mbuf_len(void) {
```
kasperl, line 247: Maybe even consider making this `const`?

```suggestion
  int send_mbuf_len() {
```

src/resources/ble_esp32.cc
```diff
@@ -233,6 +233,21 @@ class BleReadWriteElement : public BleErrorCapableResource {
     return unvoid_cast<BleReadWriteElement*>(arg)->_on_access(ctxt);
   }
 
+  static uint16_t mbuf_total_len(os_mbuf* om) {
```
kasperl, line 236: You could make this work for the `om == null` case quite easily and the code would be shorter. You would get one extra null check per call though.

src/resources/ble_esp32.cc
mikkeldamsgaard: Changing it to an int, would require static casting back to uint16_t when calling os_mbuf_appendfrom.

src/resources/ble_esp32.cc
mikkeldamsgaard: Maybe just remove this function and inline the call later? At least rename it to mbuf_to_send_len

src/resources/ble_esp32.cc
donghengqaz: Yes, the result is only used for function:

```c
int os_mbuf_appendfrom(struct os_mbuf *dst, const struct os_mbuf *src, uint16_t src_off, uint16_t len)
```
`mbuf_total_len` for `len`, so using `uint16_t is better`.

src/resources/ble_esp32.cc
```diff
@@ -233,6 +233,25 @@ class BleReadWriteElement : public BleErrorCapableResource {
...
+    }
+
+    uint16_t total_len = om->om_len;
+    om = SLIST_NEXT(om, om_next);
+    while (om) {
+      total_len += om->om_len;
+      om = SLIST_NEXT(om, om_next);
+    }
+
+    return total_len;
```
kasperl, line 248:
```suggestion
    uint16_t total_len = 0;
    while (om) {
      total_len += om->om_len;
      om = SLIST_NEXT(om, om_next);
    }
    return total_len;
```

kasperl (approved):
Looks good to me.

PR #1465, opened 2023-03-04
Fix storage.Bucket.[]= return value

kasperl:


lib/system/storage.toit
```diff
@@ -106,8 +106,9 @@ class Bucket extends ServiceResourceProxy:
   operator [] key/string -> any:
     return get key --if_present=(: it) --if_absent=(: throw "key not found")
 
-  operator []= key/string value/any -> none:
+  operator []= key/string value/any -> any:
```
floitsch, line 109: I really should fix this in the compiler...

floitsch (approved):
LGTM.

PR #1466, opened 2023-03-04
Add support for Container.on_stopped callbacks

kasperl:


lib/system/containers.toit
```diff
@@ -64,12 +68,18 @@ class Container extends ServiceResourceProxy:
...
     return code
 
+  on_stopped lambda/Lambda? -> none:
+    on_stopped_ = lambda
+    if not lambda or not result_.has_value: return
+    code := result_.get
+    if not code: throw "CLOSED"
+    lambda.call code
+
   on_notified_ code/int -> none:
```
floitsch, line 78: Does "on_notified_" mean that the container is stopped?

tests/containers_test.toit
```diff
@@ -17,3 +28,35 @@ main:
...
+  sub4 := containers.start containers.current {:}
+  sub4.on_stopped::
+    expect_not lambda4_called
+    lambda4_value = it
+    lambda4_called = true
+  // Make sure we get the lambda called before we call
+  // wait on the container. Shouldn't take too long.
+  with_timeout --ms=5_000: while not lambda4_called: sleep --ms=50
+  expect_equals 0 lambda4_value
+  expect_equals 0 sub4.wait
```
floitsch, line 59: Add a test for when the container throws an exception?

floitsch (approved):
LGTM.

lib/system/containers.toit
kasperl: Not necessarily, but here it does. In general `on_notified_` is called on the resource proxy when the `notify` method has been called on the resource.

tests/containers_test.toit
kasperl: Will do in another PR. It is a little bit tricky, because we seem to sometimes shut down everything when things fail.

kasperl: It's because the primary application container image is automatically marked RUN_BOOT | RUN_CRITICAL. The child container inherits that.

PR #1470, opened 2023-03-07
Various missing BLE features.

mikkeldamsgaard:
Added support for
 -- Bonding
 -- Secure connection to remote peers
 -- Callback for read requests


lib/ble.toit
```diff
@@ -527,6 +532,17 @@ class LocalCharacteristic extends LocalReadWriteElement_ implements Attribute:
     return read_
 
   /**
+  Manual handling of read requests. If the characteristics supports reads and the value in the constructor is null
+    then a read request will call the $block. The $block should return a ByteArray
+  */
+  read_request [block]:
```
floitsch, line 538: ?
```suggestion
  listen_for_read_request [block]:
```

```suggestion
  handle_read_request [block]:
```

```suggestion
  accept_read_requests [block]:
```

lib/ble.toit
```diff
@@ -527,6 +532,17 @@ class LocalCharacteristic extends LocalReadWriteElement_ implements Attribute:
...
+    while true:
+      resource_state_.wait_for_state DATA_READ_REQUEST_EVENT_
+      if not resource_: return
+      value := block.call
+      resource_state_.clear_state DATA_READ_REQUEST_EVENT_
+      ble_read_request_reply_ resource_ value
+  /**
```
floitsch, line 545:
```suggestion

  /**
```

floitsch (approved):
LGTM, although I'm confused how this ever worked, since the characteristic value was always `#[]`.

lib/ble.toit
```diff
@@ -427,21 +426,27 @@ class LocalService extends Resource_ implements Attribute:
     $CHARACTERISTIC_PROPERTY_BROADCAST and similar).
   $permissions is one of the CHARACTERISTIC_PERMISSIONS_* values (see
     $CHARACTERISTIC_PERMISSION_READ and similar).
-  if $value is specified, it is used as the initial value for the characteristic.
+  If $value is specified, it is used as the initial value for the characteristic. If $value is null,
+    then the characteristic supports callback reads and the client needs
+    to call $read_request to provide the value upon request. NOTE: Read callbacks are not supported
+    in MacOS.
```
floitsch, line 432:
```suggestion
    to call $LocalCharacteristic.read_request to provide the value upon request.
  NOTE: Read callbacks are not supported on MacOS.
```

lib/ble.toit
```diff
@@ -427,21 +426,27 @@ class LocalService extends Resource_ implements Attribute:
...
+    to use.
   Throws if the service is already deployed.
   */
-  add_characteristic
+  add_characteristic -> LocalCharacteristic
       uuid/BleUuid
       --properties/int
       --permissions/int
-      --value/ByteArray=#[] -> LocalCharacteristic:
+      --value/ByteArray=#[]
```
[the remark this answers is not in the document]
floitsch: Many options:

1. If the value is not allowed to be null, then using `#[]` as indicator works fine.
2. create different functions, where one has a `--value` and the other one doesn't. This would also make sense, since the `--read_timeout_ms` only applies to one of the two. However, it would require to always provide a value.
3. Use a different flag to indicate that the value can change. For example, you could require `read_timeout_ms` to be specified if the value is dynamic. You could even have a '--dynamic' flag (or so), that must be provided when the value can change.
4. Use a token: `--value/ByteArray=MY_TOKEN_BYTE_ARRAY_`. This way you would know whether nothing was provided (since the value would be equal to the token). You would need to use `identical`.

floitsch:
Thanks!

PR #1480, opened 2023-03-09
The pin restrictions for MCLK is not relevant for ESP32S3.

mikkeldamsgaard:
Removing the check for the restriction on MCLK pin from the toit code. The C++ code does still check it.


kasperl (approved):
Looks good to me.

PR #1481, opened 2023-03-09
I2S: Removing alignment requirement

mikkeldamsgaard:
Simplifying the I2S driver. 
It no longer requires aligned writes (which probably didnt work any way).


kasperl (approved):
Looks good to me!

PR #1483, opened 2023-03-10
Add setenv primitive.

erikcorry:
Also makes Unicode work for setenv and getenv on Windows.


src/os_posix.cc
```diff
@@ -255,8 +255,20 @@ void OS::out_of_memory(const char* reason) {
   abort();
 }
 
-const char* OS::getenv(const char* variable) {
-  return ::getenv(variable);
+char* OS::getenv(const char* variable) {
+  // TODO(florian): getenv is not reentrant.
```
kasperl, line 259: That is really easy to do. We have lots of global mutex dying to be used from a Locker.

src/primitive_core.cc
```diff
@@ -2295,11 +2295,28 @@ PRIMITIVE(get_env) {
...
   if (result == null) return process->program()->null_object();
-  return process->allocate_string_or_error(result, strlen(result));
+  Object* string_or_error = process->allocate_string_or_error(result, strlen(result));
+  free(result);
+  return string_or_error;
+#endif
+}
+
+PRIMITIVE(set_env) {
+#if defined (TOIT_FREERTOS)
```
kasperl, line 2307: I think we typically use #ifdef here.

src/primitive_core.cc
```diff
@@ -2295,11 +2295,28 @@ PRIMITIVE(get_env) {
...
+  // FreeRTOS supports environment variables, but we prefer not to expose them.
+  UNIMPLEMENTED_PRIMITIVE;
+#else
+  ARGS(cstring, key, cstring, value);
+  // TODO(florian): setenv may not be reentrant.
```
kasperl, line 2312: Maybe just fix this while you're here? It is really easy.

PR #1488, opened 2023-03-14
Provide a way to run only fast tests.

erikcorry:
Skips LSP tests, type propagation tests, the _slow tests and the valgrind tests.  Run make test in about 21 seconds.

Switch to fast mode with:

NO_SLOW_TESTS=1 make rebuild-cmake

switch back with:

make rebuild-cmake


floitsch:
I think a better approach:
https://github.com/toitlang/toit/pull/1489

PR #1490, opened 2023-03-14
Fixing missing IRAM markers

mikkeldamsgaard:
Some of the methods called from the ISR was not marked as IRAM. This causes flashing to fail while UART is being used.


kasperl (approved):
Looks good to me.

PR #1496, opened 2023-03-17
Warn if a toitdoc interpolation ends with '-X'.

floitsch:
These would be interpreted differently with kebap case.

### DIFF
--- a/src/compiler/parser.cc
+++ b/src/compiler/parser.cc
@@ -2094,6 +2094,14 @@ ToitdocReference* Parser::parse_toitdoc_identifier_reference(int* end_offset) {
     *end_offset = current_state().scanner_state.to;
     consume();
     is_setter = true;
+  } else if (!is_operator
+        && is_current_token_attached()
+        && current_token() == Token::SUB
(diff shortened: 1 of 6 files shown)


src/compiler/parser.cc
```diff
@@ -2094,6 +2094,14 @@ ToitdocReference* Parser::parse_toitdoc_identifier_reference(int* end_offset) {
     *end_offset = current_state().scanner_state.to;
     consume();
     is_setter = true;
+  } else if (!is_operator
+        && is_current_token_attached()
+        && current_token() == Token::SUB
+        && is_next_token_attached()
+        && (peek_token() == Token::IDENTIFIER || peek_token() == Token::INTEGER || peek_token() == Token::DOUBLE)) {
+    // This would become a valid identifier with kebap case.
```
kasperl, line 2102: This is Turkish, I believe. English is kebab :)

kasperl (approved):
Looks good to me.

src/compiler/parser.cc
floitsch: done.

PR #1499, opened 2023-03-20
Add net.Interface.on_closed callback support

kasperl:


lib/system/base/network.toit
```diff
@@ -104,6 +122,38 @@ monitor NetworkState:
...
 
+/**
+The $CloseableNetwork is a convenience base class for implementing
+  closeable networks that keep track of a single listener for the
+  on-closed events.
+
+Subclasses must take care to provide an implementation of the $close_
+  method instead of overriding the $close method.
+*/
+abstract class CloseableNetwork:
```
floitsch, line 133: Add a comment that this should be a mixin.

floitsch (approved):
LGTM.

lib/system/base/network.toit
kasperl: Done.

PR #1506, opened 2023-03-21
Don't pre-allocate memory for TLS handshake

erikcorry:
Now that tagging_mbedtls_calloc tries to GC-then-retry failed allocations, perhaps we don't need to free up memory before doing a handshake.  Lets try this and see how it works.  It's also possible that a less
agressive version of this code would be right.


kasperl (approved):
Looks good to me!

PR #1510, opened 2023-03-22
Make profile-flags directories

floitsch:
When opening multiple editors we start different LSP servers. Having a unique filename in the option would make them fight for that file.


kasperl (approved):
Looks good to me.

PR #1516, opened 2023-03-24
Avoid creating new tasks for service calls within a process

kasperl:


lib/system/services.toit
```diff
@@ -272,24 +273,34 @@ class ServiceClient:
...
 
   stringify -> string:
     return "service:$_name_@$(_major_).$(_minor_).$(_patch_)"
 
   invoke_ index/int arguments/any -> any:
     id := _id_
     if not id: throw "Client closed"
-    return rpc.invoke _pid_ RPC_SERVICES_INVOKE_ [id, index, arguments]
+    pid := _pid_
+    if not ServiceManager_.uninitialized and pid == Process.current.id:
```
erikcorry, line 287:
```suggestion
    // If we are communicating with our own process, use a short cut.
    // The check on the uninitialized flag causes this code to be
    // optimized out by the type inferencer in simple programs.
    if not ServiceManager_.uninitialized and pid == Process.current.id:
```

erikcorry (approved):

PR #1522, opened 2023-03-27
Add crypto.md5 support

kasperl:


lib/crypto/md5.toit
```diff
@@ -0,0 +1,149 @@
...
+
+  a_/int := 0x67452301
+  b_/int := 0xefcdab89
+  c_/int := 0x98badcfe
+  d_/int := 0x10325476
+
+  add data from/int to/int -> none:
+    if not buffer_: throw "ALREADY_CLOSED"
+    slice := data[from..to]
+    add_bytes_ (slice is ByteArray ? slice : slice.to_byte_array)
```
erikcorry, line 42: You can't take arbitrary slices of strings because of malformed UTF-8.

```suggestion
    if data is string:
      add_bytes_ (data.to_byte_array from to)
    else:
      add_bytes_ data[from..to]
```

kasperl: As discussed, we will leave this out.

PR #1524, opened 2023-03-27
Reland stack shrinking during GC (#1518).

erikcorry:
This change was innocent of the crashes we had attributed to it.


kasperl (approved):
Looks good to me!

PR #1533, opened 2023-03-30
Tiny adjustment to metadata calculation.

erikcorry:
Only makes a difference for theoretical devices with very small memories.


kasperl (approved):
Looks good to me.

PR #1535, opened 2023-03-30
Add support for opencore ethernet MAC.

floitsch:
This MAC is supported in Qemu (but not in actual hardware).


src/resources/ethernet_esp32.cc
```diff
@@ -200,7 +203,16 @@ PRIMITIVE(init_esp32) {
   mac_config.smi_mdio_gpio_num = mdio_num;
 
   // TODO(anders): If phy initialization fails, we're leaking this.
```
kasperl, line 205: Maybe fix this while you're in here? Might be harder than you would think.

lib/net/modules/ethernet.toit
```diff
@@ -39,11 +41,20 @@ class Ethernet:
 
     if mac_chip == MAC_CHIP_ESP32:
       resource_group_ = ethernet_init_esp32_
+        mac_chip
         phy_chip
         phy_addr
         (phy_reset ? phy_reset.num : -1)
         mac_mdc.num
         mac_mdio.num
+    else if mac_chip == MAC_CHIP_OPENETH:
```
kasperl, line 50: Maybe share more of this code with the case above?

kasperl (approved):
Looks good to me.

How much extra space does this take up?

lib/net/modules/ethernet.toit
floitsch: done.

src/resources/ethernet_esp32.cc
floitsch: Will do it in a separate PR.

PR #1539, opened 2023-03-31
Add Hmac support to core crypto lib

erikcorry:


lib/core/collections.toit
```diff
@@ -1462,6 +1462,7 @@ abstract class ByteArrayBase_ implements ByteArray:
 
   /**
   Concatenates this instance with $other.
+  Always creates a fresh ByteArray even if the receiver or the argument is empty.
```
kasperl, line 1465: For consistency?

```suggestion
  Always creates a fresh byte array even if the receiver or the argument is empty.
```

kasperl (approved):
Looks good to me.

PR #1540, opened 2023-04-02
Parse ISO 8601 with +/- time zones

erikcorry:


lib/core/time.toit
```diff
@@ -714,10 +714,29 @@ class Time implements Comparable:
   The $str must be in ISO 8601 format. For example "2019-12-18T06:22:48Z".
   */
   constructor.from_string str/string:
-    is_utc := str.ends_with "Z"
+    zone_is_adjusted := str.ends_with "Z"
     str = str.trim --right "Z"
+    str_to_int ::= : | s/string | int.parse s --on_error=: throw "Cannot parse $s as integer"
```
floitsch, line 719: I think a better error message is to say that this is an invalid Time string.

floitsch (approved):
LGTM.

PR #1542, opened 2023-04-04
Add function to get cryptographically random bytes from MbedTLS

erikcorry:


floitsch (approved):
LGTM.

We should consider exposing the function in the `crypto` library.

PR #1543, opened 2023-04-04
Implement pseudo-random-function for TLS.

erikcorry:
This algorithm, from RFC 5246, is used to expand the mutually agreed secret into enough key data (keys, IVs) to run a connection with symmetric ciphers.


floitsch (approved):
LGTM.

PR #1544, opened 2023-04-05
Allow fractional seconds in time parsing.

erikcorry:
It turns out that ISO 8601 is a very lax standard allowing all sorts of strange date/time formats, inclusing missing parts, eg. 2023-01 for January 2023.  It also allows fractional months, like 2023-01.53743.

Instead of trying to support the whole ISO standard, which would cause huge bloat in the standard library, we instead support the sane subset, which is documented in RFC 3339.


lib/core/time.toit
```diff
@@ -726,32 +739,35 @@ class Time implements Comparable:
...
     parts ::= str.split "T"
     if parts.size != 2: throw "INVALID_ARGUMENT"
     date_parts ::= (parts[0].split "-").map str_to_int
     if date_parts.size != 3: throw "INVALID_ARGUMENT"
-    time_parts ::= (parts[1].split ":").map str_to_int
+    time_parts ::= parts[1].split ":"
     if time_parts.size != 3: throw "INVALID_ARGUMENT"
+    h := str_to_int.call time_parts[0]
+    m := str_to_int.call time_parts[1]
+    s_float := float.parse time_parts[2]  // Float.parse has no --on_error parameter.
```
floitsch, line 758: Wouldn't it be better to split the seconds and then int-parse the nano-seconds?

This is how I did it in my version:
```
    if time_string_parts[2].contains ".":
      split := time_string_parts[2].split "."
      time_string_parts[2] = split[0]
      time_string_parts.add "$split[1]000000000"[..9]
    else:
      time_string_parts.add "0"
    time_parts := time_string_parts.map: int.parse it
    return Time.local_or_utc_
      date_parts[0]
      date_parts[1]
      date_parts[2]
      time_parts[0]
      time_parts[1]
      time_parts[2]
      --ms=0
      --us=0
      --ns=time_parts[3]
      --is_utc=is_utc
```

floitsch (approved):
LGTM, although I would prefer not to use float.parse.

floitsch:
LGTM. thanks.

PR #1549, opened 2023-04-12
Try to make UART code ISR safe

kasperl:


src/resources/uart_esp32.cc
```diff
@@ -888,24 +913,28 @@ PRIMITIVE(wait_tx) {
...
+  RxBuffer* buffer = uart->rx_buffer();
+  uword available = buffer->available_size();
+  if (available == 0) return process->program()->null_object();
 
-  ByteArray* data = process->allocate_byte_array(static_cast<int>(available), /*force_external*/ available != 0);
+  // TODO(kasper): It isn't obviously a good idea to just return
+  // all the data in a potentially rather large external byte array.
+  // For reads from TCP sockets, we chunk it up instead and prefer
+  // to return multiple smaller byte arrays.
+  ByteArray* data = process->allocate_byte_array(available);
```
kasperl, line 933: We probably need to think about sticking to external byte arrays for larger reads, if that is necessary for performance. @mikkeldamsgaard might know.

erikcorry (approved):

src/resources/uart_esp32.cc
mikkeldamsgaard: Maximum buffer size is 2k, so there should be no problem in having this not external.

src/resources/uart_esp32.cc
```diff
@@ -262,7 +269,35 @@ class UartResourceGroup : public ResourceGroup {
   uint32 on_event(Resource* r, word data, uint32 state) override;
 };
 
-UART_ISR_INLINE void RxTxBuffer::return_buffer(uint8* buffer) {
+UART_ISR_INLINE uword RxTxBuffer::free_size_isr() {
```
mikkeldamsgaard, line 272: I am not sure this makes a difference as `portENTER_CRITICAL` and `portENTER_CRITICAL_ISR` on esp-idf are identical.

src/resources/uart_esp32.cc
```diff
@@ -262,51 +275,85 @@ class UartResourceGroup : public ResourceGroup {
...
+    abort();
   }
 
   uart()->enable_interrupt_index(UART_TOIT_INTR_TXFIFO_EMPTY);
 }
 
-UART_ISR_INLINE uint8 TxBuffer::read_break() {
-  IsrSpinLocker locker(&spinlock_);
-
+UART_ISR_INLINE void TxBuffer::handle_empty_isr(IsrSpinLocker& locker) {
```
mikkeldamsgaard, line 330: I miss a `IsrSpinLocker locker(&spinlock_);` here.

src/resources/uart_esp32.cc
```diff
@@ -319,6 +366,9 @@ UART_ISR_INLINE uint8* TxBuffer::read(uword *received, uword max_length) {
   uword maximum_to_receive = Utils::min(max_length, static_cast<uword>(transfer_header_.remaining_data_length_));
 
   if (maximum_to_receive == 0) {
+    // TODO(kasper): What does this mean? The payload is empty? Do
+    // we need to handle this as a case where the TxBuffer is empty?
```
mikkeldamsgaard, line 370: This will be the case when there is no data but there is a break length.

kasperl: So maybe we don't need the call to is_empty_isr? If I can add a parameter to handle_empty_isr that would be pretty nice.

PR #1552, opened 2023-04-13
Limit concurrent TLS handshakes

kasperl:


src/resources/tls.cc
```diff
@@ -667,5 +667,98 @@ PRIMITIVE(get_random) {
   return process->program()->null_object();
 }
 
+#ifdef TOIT_FREERTOS
+// On small platforms we disallow concurrent handshakes
+// to avoid running into memory issues.
+static const int HANDSHAKE_CONCURRENCY = 1;
+#else
+static const int HANDSHAKE_CONCURRENCY = 1;
```
kasperl, line 675: This should probably be 16/32/64 or something. I'm using 1 for testing.

erikcorry (approved):

PR #1559, opened 2023-04-18
Add RS485 half-duplex RTS handling to UART driver

kasperl:


src/resources/uart_esp32_hal.c
```diff
@@ -112,6 +112,14 @@ void uart_toit_hal_set_mode(uart_hal_handle_t hal, uart_mode_t mode) {
   uart_hal_set_mode(HAL, mode);
 }
 
+bool IRAM_ATTR uart_toit_hal_is_tx_idle(uart_hal_handle_t hal) {
+  return uart_hal_is_tx_idle(HAL);
+}
+
+void IRAM_ATTR uart_toit_hal_set_rts(uart_hal_handle_t hal, bool active) {
+  uart_hal_set_rts(HAL, active ? 0 : 1);
```
kasperl, line 120: This surprised me a bit, but I deduced the `active` is zero from another UART driver.

src/resources/uart_esp32.cc
```diff
@@ -428,12 +435,32 @@ UART_ISR_INLINE void UartResource::disable_rx_interrupts() {
   disable_interrupt_index(UART_TOIT_INTR_RX_TIMEOUT);
 }
 
-UART_ISR_INLINE void UartResource::enable_tx_empty_interrupts() {
+UART_ISR_INLINE void UartResource::enable_tx_interrupts(bool begin) {
```
floitsch, line 438: I think the name is now a bit confusing: it's true that the TXFIFO interrupt is enabled, but the TX_DONE is disabled.
Maybe: `set_tx_start_interrupts`
and then `set_tx_stop_interrupts`

src/resources/uart_esp32_hal.c
floitsch: Interesting. Until now I didn't even realize that we would "bind" the RTS to the UART. I thought we would just set the pin directly.
No need for comments, but do you know if there is an advantage in using the `uart_hal` for the RTS, compared to just bit-banding the RTS by hand?

floitsch (approved):
LGTM.

src/resources/uart_esp32.cc
```diff
@@ -488,7 +515,20 @@ UART_ISR_INLINE void UartResource::handle_isr() {
...
   } else if (status & interrupt_mask(UART_TOIT_INTR_TX_BRK_DONE)) {
     disable_interrupt_index(UART_TOIT_INTR_TX_BRK_DONE);
-    enable_tx_empty_interrupts();
+    clear_interrupt_index(UART_TOIT_INTR_TX_BRK_DONE);
+    enable_tx_interrupts(false);
+  } else if (status & interrupt_mask(UART_TOIT_INTR_TX_DONE)) {
+    if (rts_active_) {
+      // If the transmit is still active, we can't de-activate
+      // the RTS just yet. Just postpone interrupt processing
+      // for next interrupt.
```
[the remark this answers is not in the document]
kasperl: TX_DONE. Added comment.

src/resources/uart_esp32.cc
kasperl: Reworked.

PR #1560, opened 2023-04-18
Add clone method to all Checksum objects.

erikcorry:
This lets you save work by summing two data sets that start with the same data.  This is useful for HMAC and for the handshake checksum of TLS.


lib/crypto/checksum.toit
```diff
@@ -40,6 +40,9 @@ abstract class Checksum:
   /** Computes the checksum from the added data. */
   abstract get -> ByteArray
 
+  /** Clones the internal state so we can compute checksums of multiple data with the same prefix. */
```
kasperl, line 43: Maybe break the line?

PR #1565, opened 2023-04-20
Pure-Toit TLS session resume.

erikcorry:


lib/tls/session.toit
```diff
@@ -180,6 +213,18 @@ class Session:
...
+
+    if session_state:
+      toit_handshake_ = ToitHandshake_ this
+      try:
+        toit_handshake_.handshake
+        // Connected.
+        return
+      finally: | is_exception exception |
+        value := is_exception ? exception.value : null
+        handshake_in_progress_.set value
```
kasperl, line 225: The latch needs to know if you set a value or an exception, so the call to get can throw if necessary.

```suggestion
        value := is_exception ? exception.value : null
        handshake_in_progress_.set value --exception=is_exception
```

lib/tls/session.toit
```diff
@@ -180,6 +213,18 @@ class Session:
       throw "TLS_ALREADY_HANDSHAKEN"
 
     handshake_in_progress_ = monitor.Latch
+
+    if session_state:
+      toit_handshake_ = ToitHandshake_ this
+      try:
+        toit_handshake_.handshake
```
kasperl, line 220: Do you need to keep the `toit_handshake_` around after this? Seems a bit off. Maybe it shouldn't be stored in a field at all.

lib/tls/session.toit
```diff
@@ -534,6 +573,317 @@ class Session:
...
+        --secret=master_secret_
+        --label="client finished"
+        --seed=client_handshake_hash
+        cipher_suite_.hmac_hasher
+    client_finished = #[FINISHED_, 0x00, 0x00, client_finished.size] + client_finished
+    set_up_session_ write_key read_key client_finished server_finished_expected
+
+  set_up_session_ write_key/KeyData_ read_key/KeyData_ client_finished/ByteArray server_finished_expected/ByteArray -> none:
+    // Add message header - 1 byte for type, 3 bytes for length.
+    session_.reads_encrypted_ = true
```
kasperl, line 726: Do we need to change stuff in the session before we know we'll complete this without errors? It also feels a bit weird that the `ToitHandshake_` updates state in the session.

Maybe the handshake could just build the symmetric session and the session state and return them as a handshake result of sorts and then the session could apply it by updating its state?

kasperl (approved):
Looks good to me.

Maybe move the handshake code to a handshake.toit file in a separate PR?

PR #1566, opened 2023-04-20
Buffer firmware writes

kasperl:


lib/system/firmware.toit
```diff
@@ -105,22 +105,71 @@ It is common that newly installed firmware boots with
...
   commit --checksum/ByteArray?=null -> none:
+    flush_
     _client_.firmware_writer_commit handle_ checksum
 
+  flush_ buffer/ByteArray?=buffer_ buffered/int=buffered_ -> none:
+    if not buffer: return
+    // When the buffer is full, we avoid sending a slice because
+    // that disables the neutering transfer. This is the common
+    // case, so we prefer a little less copying.
+    if buffered < BUFFER_SIZE_: buffer = buffer[..buffered]
```
erikcorry, line 168: ByteArray just returns `this` if you ask for a slice that is the whole ByteArray, so I don't think the `if` is necessary.

lib/system/firmware.toit
```diff
@@ -105,22 +105,71 @@ It is common that newly installed firmware boots with
...
 class FirmwareWriter extends ServiceResourceProxy:
+  static BUFFER_SIZE_ ::= 4096
+  buffer_/ByteArray? := null
+  buffered_/int := 0
+
   constructor from/int to/int:
     if not _client_: throw "UNSUPPORTED"
     super _client_ (_client_.firmware_writer_open from to)
 
   write bytes/ByteArray -> none:
```
erikcorry, line 116: Does this function deserve a comment saying that it may neuter the byte arrays that are passed to it?

lib/system/firmware.toit
```diff
@@ -105,22 +105,71 @@ It is common that newly installed firmware boots with
...
+
+    if size >= BUFFER_SIZE_:
+      _client_.firmware_writer_write handle_ bytes
+    else:
+      buffer := ByteArray BUFFER_SIZE_
+      buffer.replace 0 bytes 0 size
+      buffer_ = buffer
+      buffered_ = size
+
+  copy mapping/FirmwareMapping -> none:
```
erikcorry, line 140: Is this only called with large areas?

erikcorry (approved):

lib/system/firmware.toit
kasperl: Good point!

lib/system/firmware.toit
kasperl: Yes. Whole sections.

lib/system/firmware.toit
kasperl: Done.

PR #1567, opened 2023-04-21
Primitive to count the bytes matching in two blobs

erikcorry:


erikcorry:
Probably not worth a new primitive.

PR #1571, opened 2023-04-24
fix `make clean` when there are no sdkconfig files

erikcorry:
find toolchains -name sdkconfig -exec rm '{}' ';' when there are no sdkconfig files


kasperl (approved):
Looks good to me.

PR #1572, opened 2023-04-24
Make ChaCha20/Poly1305 support optional.

erikcorry:
With this change you can disable these TLS encryption protocols in sdkconfig, and the VM will still compile.

Almost all web-accessible servers will work just fine with AES based encryption, so this can save some space.

For desktop/server, the correct place to make this change would be in
third_party/esp_idf/components/mbedtls/mbedtls/include/mbedtls/config.h However, space is not an issue on these platforms.

### DIFF
--- a/src/primitive_crypto.cc
+++ b/src/primitive_crypto.cc
@@ -32,6 +32,12 @@
 #include "siphash.h"
 #include "tags.h"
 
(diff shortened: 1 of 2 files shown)


kasperl (approved):
Looks good to me.

Maybe consider using a switch to avoid messing with nesting, if-else, and ifdefs?

PR #1574, opened 2023-04-25
Recognize OOM errors from TLS more reliably.

erikcorry:
Low-level error codes are globally unique, but are combined bitwise with
high level errors. This sorts them out when reporting exceptions up to the
Toit code.


kasperl (approved):
Looks good to me.

PR #1580, opened 2023-04-27
Better SPI RAM support. 

mikkeldamsgaard:
On some configurations the malloc will allocate in SPI RAM, so changed the malloc's to use heap_cap_malloc to make sure the ram needed by the interrupt handler stays in internal memory.


kasperl (approved):
Looks good to me.

PR #1582, opened 2023-04-27
Prepare for esp-idf 5.0

erikcorry:
These are changes to make newer compiler happy that we can already do now.


kasperl (approved):
Looks good to me.

PR #1585, opened 2023-04-28
Shrink the file read buffers (1MB -> 64KB)

kasperl:


kasperl:
Better performance / memory tradeoff.

TBR'ed.

PR #1591, opened 2023-04-30
Support esptool in path.

floitsch:


floitsch:
When packaging Toit for Linux, there is sometimes (for example Arch Linux) already an esptool package which Toit can depend on.

tools/firmware.toit
```diff
@@ -577,14 +577,26 @@ find_esptool_ -> List:
...
+      return esptool
+  // Try to find esptool in PATH.
+  esptool := ["esptool$bin_extension"]
+  catch:
+    pipe.backticks esptool "version"
+    // Succeeded, so just return it.
     return esptool
+  // An exception was thrown.
+  // Try to find esptool.py in PATH.
+  if platform != PLATFORM_WINDOWS:
```
kasperl, line 594: Don't we need something similar on Windows?

kasperl (approved):
Looks good to me.

tools/firmware.toit
floitsch: I don't thing shebang scripts work on Windows.
Python adds a file association of Python to the launcher, but I don't think that triggers when directly called (even through a shell).
I think on Windows we need a real executable.

PR #1592, opened 2023-04-30
Add support for cross-compilation.

floitsch:


src/compiler/toitc.cc
```diff
@@ -39,7 +39,7 @@ static void print_usage(int exit_code) {
   printf("  [--force]                                // Finish compilation even with errors (if possible).\n");
   printf("  [-Werror]                                // Treat warnings like errors.\n");
   printf("  [--show-package-warnings]                // Show warnings from packages.\n");
-  printf("  [--vessels-root <dir>]                   // Path to vessels when compiling executables.\n");
+  printf("  [--arch <architecture>]                  // Cross-compilation target.\n");
   printf("  [--strip]                                // Strip the output of debug information.\n");
   printf("  { -o <executable> <toitfile|snapshot> |  // Write executable.\n");
   printf("    -w <snapshot> <toitfile|snapshot> |    // Write snapshot file.\n");
```
kasperl, line 45: Does this really take a snapshot as input?

src/compiler/executable.cc
```diff
@@ -48,10 +48,14 @@ static const uint8 VESSEL_TOKEN[] = { VESSEL_TOKEN_VALUES };
 // just much more complicated for something that doesn't change that frequently.
 static int VESSEL_SIZES[] = { 128, 256, 512, 1024, 8192, };
 
-static int sign_if_necessary(const char* out_path) {
+static int sign_if_necessary(const char* out_path, const char* arch) {
 #ifndef TOIT_DARWIN
+  // TODO(florian): sign if arch equals "macos".
```
kasperl, line 53: macos isn't an architecture?

Typically, architecture will be something like arm64 and operating system will be macos, linux, win.

src/compiler/toitc.cc
```diff
@@ -39,7 +39,7 @@ static void print_usage(int exit_code) {
   printf("  [--force]                                // Finish compilation even with errors (if possible).\n");
   printf("  [-Werror]                                // Treat warnings like errors.\n");
   printf("  [--show-package-warnings]                // Show warnings from packages.\n");
-  printf("  [--vessels-root <dir>]                   // Path to vessels when compiling executables.\n");
+  printf("  [--arch <architecture>]                  // Cross-compilation target.\n");
```
kasperl, line 42: It feels like this isn't an architecture.

kasperl (approved):
Looks good to me.

PR #1594, opened 2023-04-30
Add different vessels to release builds.

floitsch:


.github/workflows/ci.yml
```diff
@@ -712,6 +712,52 @@ jobs:
         with:
           name: macOS-build
 
+      - name: Unpack SDKs
+        run: |
+          for os in windows linux macos rpi rpi64; do
```
kasperl, line 717: Can the list of operating systems be shared? `windows linux macos rpi rpi64`

It is a bit of a weird mix with linux, rpi, and rpi64 all being linux (os), but probably for different architectures.

kasperl (approved):

PR #1599, opened 2023-05-04
Reduce IRAM use to get under limit on old ESP32.

erikcorry:
We did this by disabling support for external SPIRAM on ESP32. This enabled Wifi optimization, for some reason. We disabled that because it cost a lot of IRAM space, negating the benefit.

Also includes some compile fixes for other ESP32 variants.


kasperl (approved):
Looks good to me.

PR #1601, opened 2023-05-07
Fix compilation warnings.

floitsch:


src/scheduler.cc
```diff
@@ -479,7 +479,7 @@ void Scheduler::gc(Process* process, bool malloc_failed, bool try_hard) {
     }
   }
 
-  int gcs = 0;
+  int gcs [[maybe_unused]] = 0;
```
kasperl, line 482: We typically use `USE(gcs)` instead.

kasperl (approved):
Looks good to me.

src/scheduler.cc
floitsch: done.

PR #1602, opened 2023-05-07
First steps to support wasm.

floitsch:


floitsch:
Currently stuck on mbedtls compilation issues.

Also not sure how fatal the missing epoll is.

PR #1604, opened 2023-05-07
Simplify sysroot handling.

floitsch:


Makefile
```diff
@@ -240,6 +248,9 @@ arm-linux-gnueabi: arm-linux-gnueabi-sysroot
 TOITLANG_SYSROOTS := armv7 aarch64 riscv64
 ifneq (,$(filter $(CROSS_ARCH),$(TOITLANG_SYSROOTS)))
 SYSROOT_URL=https://github.com/toitlang/sysroots/releases/download/v1.2.0/sysroot-$(CROSS_ARCH).tar.gz
+
+rebuild-cross-cmake: sysroot
```
kasperl, line 252: Not sure I understand this. Here the `rebuild-cross-cmake` dependency is added unconditionally. Why do we need the two conditional dependencies?

kasperl (approved):
Looks good to me.

Makefile
floitsch: As discussed in person: the rule is conditional on the filter in line 249 matching.

PR #1605, opened 2023-05-07
Unify host and cross-compilation

floitsch:


Makefile
```diff
@@ -51,6 +51,12 @@ endif
 
 CROSS_ARCH=
 
+ifneq ($(CROSS_ARCH),)
```
kasperl, line 54: Shouldn't this just be passed as `TARGET` instead?

I imagine you'll get a few HOST == TARGET checks instead of CROSS_ARCH == "", but that seems okay.

kasperl (approved):
Looks good to me.

Makefile
floitsch: done.

PR #1620, opened 2023-05-13
Don't require 'all' for 'install'.

floitsch:
When compiling for Archlinux we don't need the esptool. Don't create it during installation.


kasperl (approved):
Looks good to me.

PR #1637, opened 2023-06-02
Fix: Avoid extra old-space GCs

erikcorry:


src/heap.h
```diff
@@ -159,7 +159,10 @@ class ObjectHeap {
   bool has_max_heap_size() const { return max_heap_size_ != 0; }
   bool has_pending_limit() const { return limit_ != pending_limit_; }
 
+  bool retrying_primitive() const { return retrying_primitive_; }
+
   void check_install_heap_limit() {
```
kasperl, line 164: Maybe this should be called something like 'leave_primitive' now?

kasperl (approved):
Looks good to me.

PR #1638, opened 2023-06-02
Put sdkconfig files under git control

erikcorry:


erikcorry:
The utility that generates the sdkconfig.defaults file doesn't exist in this version of esp-idf

PR #1641, opened 2023-06-06
Changes to ci.yml to compile with esp-idf 5.0

erikcorry:


.github/workflows/ci.yml
```diff
@@ -411,6 +407,8 @@ jobs:
       - name: Install ESP32 tools
         run: |
           third_party/esp-idf/install.sh
+          $HOME/.espressif/python_env/idf5.0_py3.10_env/bin/python3.10 -m pip install jsonschema
```
floitsch, line 410: Why is this 3.10, if we install 3.11 in line 162.

.github/workflows/ci.yml
```diff
@@ -156,20 +156,16 @@ jobs:
           sudo tar -czf apt-cacher-ng/cache.tar.gz -C /var/cache/apt-cacher-ng .
           sudo chown $(whoami) apt-cacher-ng/cache.tar.gz
 
-      - name: Set up Python - non-Windows
+      - name: Set up Python
```
floitsch, line 159: I'm guessing we need to do the same thing on the esp32 workflow.

floitsch (approved):
LGTM.

Does this mean that we now need Python to compile Toit?
The current AUR build doesn't need it, and it would be nice if we could keep it this way.

erikcorry:
It's because of this: https://github.com/Mbed-TLS/mbedtls/blob/development/docs/proposed/psa-driver-wrappers-codegen-migration-guide.md

PR #1651, opened 2023-06-08
Update esp idf to 4.4.4

floitsch:


toolchains/esp32/sdkconfig.defaults
```diff
@@ -32,6 +32,7 @@ CONFIG_SPIRAM_IGNORE_NOTFOUND=y
 CONFIG_SPIRAM_USE_CAPS_ALLOC=y
 # CONFIG_SPIRAM_CACHE_LIBNUMPARSER_IN_IRAM is not set
 # CONFIG_SPIRAM_CACHE_LIBCHAR_IN_IRAM is not set
+# CONFIG_SPIRAM_CACHE_LIBFILE_IN_IRAM is not set
```
kasperl, line 35: Do we need something similar for the other variants?

kasperl (approved):
Looks good to me.

toolchains/esp32/sdkconfig.defaults
floitsch: I compiled all the other variants and they don't seem to run into iram issues.

PR #1652, opened 2023-06-08
Reduce generated code size in primitives.

erikcorry:
3.7% decreas in the size of the disassembly, but only about 1% decrease in the size of the bin.


src/process.h
```diff
@@ -290,6 +313,29 @@ class Process : public ProcessListFromProcessGroup::Element,
 
   Profiler* profiler_ = null;
 
+  Object* marked_allocation_failed_;
```
kasperl, line 316: It would be nice if the interpreter could create these from a table in the program -- and then we can let the primitives themselves return the (marked) indexes into the table marked as the failure.

kasperl (approved):
Looks good to me.

PR #1658, opened 2023-06-15
Remove autogenerated unchecked call to new

erikcorry:


kasperl (approved):
Looks good to me.

Maybe we should just always do this and never use capture.process in any other position?

PR #1659, opened 2023-06-16
Add ethernet service

kasperl:
Users of the ethernet can use:

```
import net.ethernet
main:
  network := ethernet.open
  ...
  network.close
```

but someone has to provide the ethernet driver as a service by importing `esp32.net.ethernet` and installing a configured `EthernetServiceProvider` instance.

### DIFF
--- a/lib/esp32/net/ethernet.toit
+++ b/lib/esp32/net/ethernet.toit
@@ -3,7 +3,7 @@
 // found in the lib/LICENSE file.
 
 /**
-Network driver for wired Ethernet.
+Network driver as a service for wired Ethernet.
 */
 
 import esp32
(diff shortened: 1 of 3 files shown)


lib/esp32/net/ethernet.toit
```diff
@@ -69,190 +55,141 @@ index df798c8..fef8c8a 100644
...
+  handle index/int arguments/any --gid/int --client/int -> any:
+    if index == EthernetService.CONNECT_INDEX:
+      return connect client
+    if index == NetworkService.ADDRESS_INDEX:
+      network := (resource client arguments) as NetworkResource
+      return address network
+    unreachable
+
+  connect client/int -> List:
+    module ::= (state_.up: EthernetModule_ this create_resource_group_) as EthernetModule_
```
floitsch, line 116: Why do we store the result of the state_.up call in a local?

floitsch (approved):
LGTM.

lib/esp32/net/ethernet.toit
kasperl: Very good question. I think we need this in the wifi provider, so I just blindly copied it :) 

Dropped.

PR #1666, opened 2023-06-21
Better support custom esptool paths.

floitsch:


tools/firmware.toit
```diff
@@ -123,6 +124,7 @@ main arguments/List:
       ]
   root_cmd.add create_cmd
   root_cmd.add extract_cmd
+  root_cmd.add esptool_cmd
```
kasperl, line 127: Should we put this under a subcommand so it is easier to extend for non-ESP targets in the future?

`firmware tool esptool`

kasperl (approved):
Looks good to me.

tools/firmware.toit
floitsch: done.

PR #1667, opened 2023-06-21
Add 'describe' to firmware tool.

floitsch:


tools/firmware.toit
```diff
@@ -906,6 +907,48 @@ extract_binary_content -> ByteArray
   binary.patch_extend_drom system_uuid table_address extension
   return binary.bits
 
+describe_cmd -> cli.Command:
```
kasperl, line 910: Maybe this should just be 'show' like 'device show'?

tools/firmware.toit
```diff
@@ -906,6 +907,48 @@ extract_binary_content -> ByteArray
...
+  envelope := Envelope.load input_path
+
+  if output_format == "text":
+    print "Envelope format version: $envelope.version_"
+    print "SDK version: $envelope.sdk_version"
+    print
+    print "Containers:"
+    envelope.entries.do: | name entry |
+      if show_all or not name.starts_with "\$":
+        print "  $name: $(entry.size) bytes"
```
kasperl, line 936: Don't we have more to show if the entry is a snapshot? Can you share this with code with 'container list'?

kasperl (approved):
Looks good to me.

kasperl (approved):

PR #1669, opened 2023-06-23
Fix 'wait_for'.

floitsch:
We didn't clear the edge-trigger after receiving a spurious (from a previous interrupt-configuration) event.

We were crashing when the pin was closed in a separate task.


floitsch:
Alternative is to not return from the `wait_for` or to throw (a "CLOSED") exception.

kasperl (approved):
Looks good to me.

PR #1674, opened 2023-06-26
Add more Ethernet documentation and examples.

floitsch:


examples/ethernet/containers/eth.toit
```diff
@@ -0,0 +1,56 @@
...
+    super
+        --phy_chip=esp32.PHY_CHIP_LAN8720
+        --phy_address=0
+        --mac_chip=esp32.MAC_CHIP_ESP32
+        --mac_mdc=gpio.Pin 23
+        --mac_mdio=gpio.Pin 18
+        --mac_spi_device=null
+        --mac_interrupt=null
+
+  connect client/int -> List:
```
kasperl, line 33: I don't think this is really right. The `connect` can be called from multiple clients -- so also when ethernet is already connected. The `on_module_closed` is only called when last client is closed, so this isn't symmetrical.

examples/ethernet/single_container.toit
```diff
@@ -0,0 +1,51 @@
...
+      --phy_chip=esp32.PHY_CHIP_LAN8720
+      --phy_address=0
+      --mac_chip=esp32.MAC_CHIP_ESP32
+      --mac_mdc=gpio.Pin 23
+      --mac_mdio=gpio.Pin 18
+      --mac_spi_device=null
+      --mac_interrupt=null
+  provider.install
+  network := ethernet.open
+  // Use the network.
```
kasperl, line 48: Again this look weird, because the comment is associated with the `close` call.

lib/esp32/net/ethernet.toit
```diff
@@ -31,18 +33,55 @@ PHY_CHIP_DP83848  ::= 3
...
+``` toit
+import gpio
+import net.ethernet
+import esp32.net.ethernet as esp32
+
+main:
+  power := gpio.Pin --output 12
+  power.set 1
+  // sleep --ms=1000
+  provider := esp32.EthernetServiceProvider
```
kasperl, line 54: Wouldn't it be more logical to just use the specialized OlimexPoeProvider?

lib/net/ethernet.toit
```diff
@@ -4,6 +4,30 @@
...
+It requires a $ServiceProvider to be installed. On the ESP32, this
+  could be the $EthernetServiceProvider.
+
+# Example
+
+```
+import net.ethernet
+
+main:
+  client := ethernet.open
```
kasperl, line 25: We typically call this 'network'.

examples/ethernet/full/client.toit
```diff
@@ -8,9 +8,13 @@ A simple client for an Ethernet network.
 Requires the Ethernet provider to be installed.
 */
 
+import net
 import net.ethernet
 
+use network/net.Client:
```
kasperl, line 14: I would probably put this after main everywhere.

examples/ethernet/full/olimex_poe.toit
```diff
@@ -0,0 +1,59 @@
...
+    finally: | is_exception _ |
+      if is_exception and power_:
+        // If we don't succeed here, turn the power off.
+        critical_do:
+          power_.close
+          power_ = null
+
+  on_opened client:
+    super client
+    connected_clients_++
```
kasperl, line 48: The `NetworkState` also does this counting, but there it is based on connect/disconnect which is arguably the correct way.

examples/ethernet/full/olimex_poe.toit
```diff
@@ -0,0 +1,59 @@
...
+      power_.set 1
+      return super client
+    finally: | is_exception _ |
+      if is_exception and power_:
+        // If we don't succeed here, turn the power off.
+        critical_do:
+          power_.close
+          power_ = null
+
+  on_opened client:
```
kasperl, line 46: I feel this is still wrong. It should really be turned on when `EthernetModule_.connect` is called -- and turned off again when `EthernetModule_.disconnect` has called `OlimexPoeProvider.on_module_closed`.

It requires a bit of refactoring to make it easier to act on the `EthernetModule_.connect` call.

examples/ethernet/full/client.toit
floitsch: done.

examples/ethernet/full/olimex_poe.toit
```diff
@@ -0,0 +1,81 @@
...
+        --phy_address=0
+        --mac_chip=esp32.MAC_CHIP_ESP32
+        --mac_mdc=gpio.Pin 23
+        --mac_mdio=gpio.Pin 18
+        --mac_spi_device=null
+        --mac_interrupt=null
+
+  connect client/int -> List:
+    try:
+      power_ = gpio.Pin --output 12
```
kasperl, line 61: I would think you shouldn't need to do anything special in this overridden `connect`? As long as `on_module_closed` is always called when `on_module_opened` has been called, but fails.

floitsch: done.

examples/ethernet/full/olimex_poe.toit
```diff
@@ -0,0 +1,69 @@
...
+ CONFIG_ETH_DMA_TX_BUFFER_NUM=10
+```
+*/
+
+import gpio
+import esp32.net.ethernet as esp32
+
+class OlimexPoeProvider extends esp32.EthernetServiceProvider:
+  power_/gpio.Pin? := null
+  connected_clients_/int := 0
```
kasperl, line 47: Drop this again.

floitsch: done.

examples/ethernet/full/olimex_poe.toit
```diff
@@ -0,0 +1,68 @@
...
+        --mac_spi_device=null
+        --mac_interrupt=null
+
+  on_module_opened module:
+    super module
+    power_ = gpio.Pin --output 12
+    power_.set 1
+
+  on_module_closed module:
+    super module
```
kasperl, line 64: Maybe the `if power_` thing should be in a finally block?

kasperl (approved):
Looks good to me.

examples/ethernet/full/olimex_poe.toit
floitsch: moved the `super` call to last.

PR #1680, opened 2023-06-28
Add more service examples

kasperl:


examples/service/random.toit
```diff
@@ -0,0 +1,112 @@
...
+  die.close
+
+class RandomServiceClient extends ServiceClient implements RandomService:
+  static SELECTOR ::= RandomService.SELECTOR
+  constructor selector/ServiceSelector=SELECTOR:
+    assert: selector.matches SELECTOR
+    super selector
+
+  next limit/int -> int:
+    return invoke_ RandomService.NEXT_INDEX limit
```
erikcorry, line 52: Does the example show us using a private method that might not be available if we tighten up the rules on privacy?

erikcorry (approved):

examples/service/random.toit
kasperl: Yes. There are 1^20 occurrences of this.

PR #1684, opened 2023-07-03
Fix dangling 'svcs' pointer in BLE implementation

kasperl:


src/resources/ble_esp32.cc
```diff
@@ -2133,25 +2140,30 @@ PRIMITIVE(deploy_service) {
...
 
-  struct ble_gatt_svc_def gatt_svcs[2] = {
-    {
-      .type = BLE_GATT_SVC_TYPE_PRIMARY,
-      .uuid = service_resource->ptr_uuid(),
-      .includes = 0,
-      .characteristics = gatt_svr_chars
-    },
+
+  auto gatt_svcs = static_cast<ble_gatt_svc_def*>(calloc(2, sizeof(ble_gatt_svc_def)));
```
floitsch, line 2144: We are using the fact that `BLE_GATT_SVC_TYPE_END` equals 0. It's probably never going to change, but I think it helps readability to assert it.
```suggestion
  auto gatt_svcs = static_cast<ble_gatt_svc_def*>(calloc(2, sizeof(ble_gatt_svc_def)));
  static_assert(BLE_GATT_SVC_TYPE_END == 0, "Unexpected TYPE_END value");
```

Alternatively, or in addition, we could set the last entry:
```
gatt_svcs[1].type = BLE_GATT_SVC_TYPE_END;
```

src/resources/ble_esp32.cc
```diff
@@ -499,9 +496,20 @@ class BleServiceResource:
     return BLE_ERR_SUCCESS;
   }
 
-  static void dispose_gatt_svr_chars(ble_gatt_chr_def* gatt_svr_chars, int count) {
-    for (int i = 0; i < count; i++) {
-      free(gatt_svr_chars[i].descriptors);
+  static void dispose_gatt_svcs(ble_gatt_svc_def* gatt_svcs) {
+    ble_gatt_svc_def* cursor = gatt_svcs;
+    while (cursor->type != 0) {
```
floitsch, line 501:
```suggestion
    while (cursor->type != BLE_GATT_SVC_TYPE_END) {
```

floitsch (approved):
LGTM.

src/resources/ble_esp32.cc
kasperl: Done.

PR #1690, opened 2023-07-05
Improve tree shaking and fix optimization bug

floitsch:
The tree-grower would grow static calls even if they were to instance methods of classes that were not instantiated. These static calls were introduced by an earlier optimization. For example, if there is a static type check before that guarantees (assuming the check works) only one dynamic method can be reached.

Now the grower treats these calls as "filtered" virtual calls. That is, they are selector calls, but if they trigger they can only grow to the target method. As such, they also have to wait for classes to be instantiated before they are added to the grown method set.

Care must be taken with super calls: these are also static calls to instance methods, but we have to handle them as soon as we seen them.

This patch also fixes a bug where a call to `operator ==` was changed into a static call, even when the RHS was not guaranteed to be non-null.

### DIFF
--- a/src/compiler/optimizations/virtual_call.cc
+++ b/src/compiler/optimizations/virtual_call.cc
@@ -76,28 +76,28 @@ Expression* optimize_virtual_call(CallVirtual* node,
     if (direct_method != null && direct_method->is_abstract()) {
       direct_method = null;
     }
+  }
 
-    // We need to be careful not to directly call an operator ==.
-    // If the RHS is nullable, the interpreter shortcuts the call.
-    if (direct_method != null && opcode == INVOKE_EQ) {
-      auto param = node->arguments().first();
-      Type param_type = compute_guaranteed_type(param, holder, method);
-      if (param_type.is_valid()) {
-        // Unless the param-type is nullable, we can change to a static call.
-        if (param_type.is_nullable()) direct_method = null;
-      } else if (param->is_Literal()) {
-        // Unless the literal is `null`, we can change to a static call.
-        if (param->is_LiteralNull()) direct_method = null;
-      } else {
-        // Not enough information, so abandon and assume we can't change to static call.
-        direct_method = null;
-      }
-    }
-    // Similarly, we don't want to change any of the really efficient INVOKE_X opcodes.
-    if (direct_method != null && opcode != INVOKE_VIRTUAL) {
-      // TODO(florian): we may switch to a static call if the receiver isn't an int/Array.
+  // We need to be careful not to directly call an operator ==.
+  // If the RHS is nullable, the interpreter shortcuts the call.
+  if (direct_method != null && opcode == INVOKE_EQ) {
+    auto param = node->arguments().first();
+    Type param_type = compute_guaranteed_type(param, holder, method);
+    if (param_type.is_valid()) {
+      // Unless the param-type is nullable, we can change to a static call.
+      if (param_type.is_nullable()) direct_method = null;
+    } else if (param->is_Literal()) {
+      // Unless the literal is `null`, we can change to a static call.
+      if (param->is_LiteralNull()) direct_method = null;
+    } else {
+      // Not enough information, so abandon and assume we can't change to static call.
       direct_method = null;
     }
+  }
+  // Similarly, we don't want to change any of the really efficient INVOKE_X opcodes.
+  if (direct_method != null && opcode != INVOKE_VIRTUAL) {
+    // TODO(florian): we may switch to a static call if the receiver isn't an int/Array.
+    direct_method = null;
   }
 
   if (direct_method == null) {
--- a/src/compiler/optimizations/virtual_call.cc
+++ b/src/compiler/optimizations/virtual_call.cc
@@ -78,25 +78,16 @@ Expression* optimize_virtual_call(CallVirtual* node,
     }
   }
 
-  // We need to be careful not to directly call an operator ==.
-  // If the RHS is nullable, the interpreter shortcuts the call.
-  if (direct_method != null && opcode == INVOKE_EQ) {
-    auto param = node->arguments().first();
-    Type param_type = compute_guaranteed_type(param, holder, method);
-    if (param_type.is_valid()) {
-      // Unless the param-type is nullable, we can change to a static call.
-      if (param_type.is_nullable()) direct_method = null;
-    } else if (param->is_Literal()) {
-      // Unless the literal is `null`, we can change to a static call.
-      if (param->is_LiteralNull()) direct_method = null;
-    } else {
-      // Not enough information, so abandon and assume we can't change to static call.
-      direct_method = null;
-    }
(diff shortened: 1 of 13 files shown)


src/compiler/optimizations/virtual_call.cc
```diff
@@ -76,29 +76,29 @@ Expression* optimize_virtual_call(CallVirtual* node,
...
+    } else {
+      // Not enough information, so abandon and assume we can't change to static call.
       direct_method = null;
     }
   }
+  // Similarly, we don't want to change any of the really efficient INVOKE_X opcodes.
+  if (direct_method != null && opcode != INVOKE_VIRTUAL) {
+    // TODO(florian): we may switch to a static call if the receiver isn't an int/Array.
+    direct_method = null;
+  }
```
kasperl, line 101: I don't understand why we go through all the trouble for `INVOKE_EQ` when we just drop the `direct_method` later?

```suggestion
  } else if (direct_method != null && opcode != INVOKE_VIRTUAL) {
    // We don't want to change any of the other really efficient INVOKE_X opcodes.
    // TODO(florian): we may switch to a static call if the receiver isn't an int/Array.
    direct_method = null;
  }
```

src/compiler/set.h
```diff
@@ -101,13 +101,30 @@ template<typename T> class UnorderedSet {
   }
   bool erase(T x) { return set_.erase(x) > 0; }
 
+  template<typename F>
+  void erase_if(const F& callback) {
```
kasperl, line 105: I think `erase_if` sounds a bit too much like erase-everything-if-the-predicate-returns-true-when-called-once. Maybe `erase_where` or `erase_all_where`?

```suggestion
  void erase_all_where(const F& callback) {
```

src/compiler/tree.cc
```diff
@@ -30,23 +31,62 @@ using namespace ir;
...
+    return false;
+  }
+
   void visit_CallStatic(CallStatic* node) {
-    found_methods_.insert(node->target()->target());
+    auto target = node->target()->target();
+    if (target->is_instance()) {
+      if (is_super_call(node)) {
+        // For super calls we don't need to ensure that the
+        // holder class is actually instantiated and its method not shadowed..
```
kasperl, line 81:
```suggestion
        // holder class is actually instantiated and its method isn't shadowed.
```

src/compiler/tree.cc
```diff
@@ -332,8 +423,23 @@ void TreeGrower::grow(Program* program) {
             method_queue.push_back(probe);
           }
         }
+        found_filtered_selectors.for_each([&](CallSelector selector, UnorderedSet<Method*>& methods) {
+          if (methods.empty()) return;
```
kasperl, line 427: This code is repeated. Maybe factor it out by adding an abstraction around the filtered selectors?

src/compiler/tree.cc
```diff
@@ -274,6 +316,10 @@ void TreeGrower::grow(Program* program) {
     logger = &null_logger;
   }
 
+  // Selectors for which the dynamic call was changed to a static call to
+  // one of the methods in the set.
+  Map<CallSelector, UnorderedSet<Method*>> filtered_selectors;
```
kasperl, line 321: I don't like the name of this so much. I think this is easier to understand as a set of methods where we changed a dynamic call to a static call. I think adding an abstraction around this (`StaticallyCalledInstanceMethodSet`) could be nice.

src/compiler/tree.cc
```diff
@@ -318,11 +380,40 @@ void TreeGrower::grow(Program* program) {
...
     }
 
     found_selectors.erase_all(handled_selectors);
     handled_selectors.insert(found_selectors.begin(), found_selectors.end());
-    if (!found_selectors.empty()) {
+    bool has_filtered_selectors = false;
+    found_filtered_selectors.for_each([&](CallSelector selector, UnorderedSet<Method*>& methods) {
+      if (handled_selectors.contains(selector)) {
+        // All methods that fit this selector are handled anyway.
+        // No need to filter them.
```
kasperl, line 402: Should the `methods` set be erased -- or perhaps better yet should the entry for the selector in the `found_filtered_selectors` be cleared? It seems annoying that you have to do this here when you already did the thinning out earlier, but I guess the `handled_selectors` set may have grown.

src/compiler/tree.cc
```diff
@@ -30,23 +31,62 @@ using namespace ir;
...
     found_methods_.insert(node->target()->target());
     TraversingVisitor::visit_CallConstructor(node);
   }
 
+  bool is_super_call(CallStatic* node) const {
+    auto current_holder = current_method_->holder();
+    auto target = node->target()->target();
+    if (!is_This(node->arguments()[0], current_holder, target)) return false;
+    // Make sure this is actually a super call (and not a sub call).
+    auto current_class = current_holder;
```
kasperl, line 67: This would probably feel natural to have on ir::Class.

kasperl (approved):
Looks good to me.

src/compiler/tree.cc
floitsch: done.

src/compiler/tree.cc
floitsch: done.

src/compiler/tree.cc
floitsch: changed.

src/compiler/tree.cc
floitsch: changed.

src/compiler/tree.cc
floitsch: changed.

src/compiler/optimizations/virtual_call.cc
```diff
@@ -76,28 +76,19 @@ Expression* optimize_virtual_call(CallVirtual* node,
...
-    // Similarly, we don't want to change any of the really efficient INVOKE_X opcodes.
-    if (direct_method != null && opcode != INVOKE_VIRTUAL) {
-      // TODO(florian): we may switch to a static call if the receiver isn't an int/Array.
-      direct_method = null;
-    }
+  if (direct_method != null && opcode != INVOKE_VIRTUAL) {
+    // We don't want to change any of the really efficient INVOKE_X opcodes.
+    // These bytecodes are optimized for numbers/arrays and shortcut
+    // lots of bytecodes.
+    // TODO(florian): change to a static call when the receiver isn't one of
```
kasperl, line 85: Nice.

src/compiler/tree.cc
```diff
@@ -30,23 +31,107 @@ using namespace ir;
...
+      }
+    });
+  }
+
+  bool empty() const {
+    // Only looks at whether there are entries in the map. Does not
+    // run through them to see if all of the sets are empty.
+    return selectors_.empty();
+  }
+ private:
```
kasperl, line 86:
```suggestion

 private:
```

kasperl (approved):
Still looks good to me.

PR #1695, opened 2023-07-11
Fix pulse counter.

floitsch:
We forgot to register the unit in the resource group.

### DIFF
--- a/src/resources/pulse_counter_esp32.cc
+++ b/src/resources/pulse_counter_esp32.cc
@@ -165,6 +165,8 @@ class PcntUnitResource : public Resource {
     }
     cleared_ = false;
 
+    pcnt_unit_ids.put(unit_id());
+
     // In v4.4.1 there is no way to shut down the counter.
     // In later versions we have to call `pcnt_del_unit`.
     // https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/pcnt.html#install-pcnt-unit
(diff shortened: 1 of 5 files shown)


src/resources/pulse_counter_esp32.cc
```diff
@@ -268,8 +271,7 @@ PRIMITIVE(new_unit) {
 
 PRIMITIVE(close_unit) {
   ARGS(PcntUnitResource, unit)
-  unit->tear_down();
-  pcnt_unit_ids.put(unit->unit_id());
+  unit->resource_group()->unregister_resource(unit);
```
kasperl, line 274: It feels odd that we're overwriting on_unregister_resource in the group instead of just letting the tear_down functionality be the destructor.

floitsch: done.

src/resources/pulse_counter_esp32.cc
```diff
@@ -55,6 +55,29 @@ class PcntUnitResource : public Resource {
     , high_limit_(high_limit)
     , glitch_filter_ns_(glitch_filter_ns) {}
 
+  ~PcntUnitResource() override {
+    for (int i = 0; i < PCNT_CHANNEL_MAX; i++) {
+      if (!used_channels_[i]) continue;
+      auto channel = static_cast<pcnt_channel_t>(i);
+      if (channel != kInvalidChannel) {
+        // In the teardown we don't handle errors for cloning the channel.
```
kasperl, line 63: wat

kasperl (approved):
Looks good to me.

src/resources/pulse_counter_esp32.cc
floitsch: Removed.

PR #1696, opened 2023-07-12
Improve DNS fallback.

floitsch:
Fall back to the next DNS server if we can connect but get an error after that.

We had a user encountering the following stacktrace:
```
EXCEPTION error.
Connection refused
  0: DnsClient.get.<block>.<block> <sdk>/net/modules/dns.toit:218:62
  1: DnsClient.get.<block>     <sdk>/net/modules/dns.toit:182:25
  2: Task_.with_deadline_.<block> <sdk>/core/task.toit:203:16
  3: Task_.with_deadline_      <sdk>/core/task.toit:197:3
  4: with_timeout              <sdk>/core/utils.toit:181:24
  5: with_timeout              <sdk>/core/utils.toit:164:12
  6: DnsClient.get             <sdk>/net/modules/dns.toit:182:5
  7: dns_lookup                <sdk>/net/modules/dns.toit:46:17
```
Where the call to `socket.receive` throw the "Connection refused": https://github.com/toitlang/toit/blob/3f93e33624b4fe89433dbabbb56a6cb24ef75d5e/lib/net/modules/dns.toit#L218

Because we had already connected we didn't even try another DNS server.

### DIFF
--- a/lib/net/modules/dns.toit
+++ b/lib/net/modules/dns.toit
@@ -8,6 +8,7 @@ import net
 
 DNS_DEFAULT_TIMEOUT ::= Duration --s=20
 DNS_RETRY_TIMEOUT ::= Duration --s=1
+MAX_QUERY_ATTEMPTS_ ::= 2
 HOSTS_ ::= {"localhost": "127.0.0.1"}
 MAX_CACHE_SIZE_ ::= platform == "FreeRTOS" ? 30 : 1000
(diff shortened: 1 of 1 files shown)


floitsch:
Unfortunately I couldn't figure out how to (easily) add a test.

The DNS port is hard-coded, so I can't just create a local server that simulates the "Connection refused" case. It also doesn't seem to take any `net.Client` where I could have "cheated", by intercepting the socket creation and changing the port there.

erikcorry (approved):

PR #1697, opened 2023-07-13
Fix esptool detection in firmware tool

floitsch:


tools/firmware.toit
```diff
@@ -632,16 +633,16 @@ find_esptool_ -> List:
     esptool_py := "$dir/../third_party/esp-idf/components/esptool_py/esptool/esptool.py"
     if file.is_file esptool_py:
       return ["python$bin_extension", esptool_py]
-  else:
+  else if dir != "":
     esptool := ["$dir/esptool$bin_extension"]
```
kasperl, line 637: I'd let `esptool` be the string and return `[esptool]` for consistency.

kasperl (approved):
Looks good to me.

tools/firmware.toit
floitsch: Agreed. That's nicer.
done.

PR #1700, opened 2023-07-17
Improve espnow.

floitsch:
ESP Now still needs to run without WiFi interfering, but it's now possible to run WiFi programs after the container with ESP Now has finished.


kasperl (approved):
Looks good to me.

PR #1701, opened 2023-07-17
Refactor espnow.

floitsch:
Use a spinlock instead of a semaphore.

Make sure that the initialization and deinitialization is done in sync.


src/resources/espnow.cc
```diff
@@ -32,19 +32,88 @@
 #include "../event_sources/system_esp32.h"
 #include "../event_sources/ev_queue_esp32.h"
 
-#define ESPNOW_RX_DATAGRAM_NUM      8
 #define ESPNOW_RX_DATAGRAM_LEN_MAX  250
-#define ESPNOW_TX_WAIT_US           1000
+#define ESPNOW_RX_DATAGRAM_NUM      8
```
kasperl, line 36: I'd probably drop the alignment of the values now that they are so far apart due to the comment.

src/resources/espnow.cc
```diff
@@ -32,19 +32,88 @@
...
+  spinlock_t* spinlock() const { return spinlock_; }
+ private:
+  spinlock_t* spinlock_;
+};
+
+struct Datagram {
   int len;
-  uint8_t mac[6];
-  uint8_t buffer[ESPNOW_RX_DATAGRAM_LEN_MAX];
+  uint8* mac[6];
```
kasperl, line 57:
```suggestion
  uint8 mac[6];
```

src/resources/espnow.cc
```diff
@@ -32,19 +32,88 @@
...
+  uint8* buffer[ESPNOW_RX_DATAGRAM_LEN_MAX];
```
kasperl, line 58:
```suggestion
  uint8 buffer[ESPNOW_RX_DATAGRAM_LEN_MAX];
```

src/resources/espnow.cc
```diff
@@ -32,19 +32,88 @@
...
+        FAIL(MALLOC_FAILED);
+      }
+      datagrams_[i] = datagram;
+    }
+    return null;
+  }
+
+  Datagram* take() {
+    SpinLocker locker(&spinlock_);
+    int mask = 1;
```
kasperl, line 90: It feels like this could be written with clz/ctz, but maybe this is fine.

src/resources/espnow.cc
```diff
@@ -32,19 +32,88 @@
...
+        used_ &= ~mask;
+        return;
+      }
+      mask <<= 1;
+    }
+  }
+
+ private:
+  spinlock_t spinlock_{};
+  struct Datagram* datagrams_[ESPNOW_RX_DATAGRAM_NUM] = {};
```
kasperl, line 115: Is this guaranteed to be filled with nulls? Maybe assert it in the constructor.

src/resources/espnow.cc
```diff
@@ -304,54 +418,26 @@ PRIMITIVE(create) {
...
-  err = esp_now_init();
-  if (err != ESP_OK) return Primitive::os_error(err, process);
-  err = esp_now_register_send_cb(espnow_send_cb);
-  if (err != ESP_OK) return Primitive::os_error(err, process);
-  err = esp_now_register_recv_cb(espnow_recv_cb);
-  if (err != ESP_OK) return Primitive::os_error(err, process);
-  if (pmk.length() > 0) {
-    err = esp_now_set_pmk(pmk.address());
-    if (err != ESP_OK) return Primitive::os_error(err, process);
+  Object* init_result = resource->init(process, mode, pmk, phy_rate);
```
kasperl, line 437: It would be easier to track the lifetime of the pool and the event queue if they were also allocated from within `init`.

src/resources/espnow.cc
```diff
@@ -105,126 +213,124 @@ class EspNowResource : public EventQueueResource {
 
   bool receive_event(word* data) override;
 
+  // Returns null if the initializations succeeded.
+  Object* init(Process* process, int mode, Blob pmk, wifi_phy_rate_t phy_rate);
+
  private:
+  enum class State {
+    constructed,
```
kasperl, line 221: I think I'd prefer `SCREAMING_CASE`.

kasperl (approved):
Looks good to me.

src/resources/espnow.cc
floitsch: The event queue is given to the Resource constructor (and its super), but I could allocate the espnow_pool in the init function.

PR #1703, opened 2023-07-17
Switch to an event-based model for ESP Now.

floitsch:


src/resources/espnow.cc
```diff
@@ -46,43 +49,67 @@ struct DataGram {
 
 const int kInvalidEspNow = -1;
 
+// These constants must be synchronized with the Toit code.
+const int kDataAvailableState = 1 << 0;
+const int kSendDoneState = 1 << 1;
+
+enum class EspNowEvent {
+  new_data_available = 1 << 0,
```
kasperl, line 57: `SCREAMING_CASE` and why bother with the bit assignment? That just seems confusing since you can only have on `EspNowEvent` value.

src/resources/espnow.cc
```diff
@@ -157,10 +191,10 @@ static void free_datagram(struct DataGram* datagram) {
 
 static void espnow_send_cb(const uint8_t *mac_addr, esp_now_send_status_t status) {
```
kasperl, line 192:
```suggestion
static void espnow_send_cb(const uint8* mac_addr, esp_now_send_status_t status) {
```

src/resources/espnow.cc
```diff
@@ -46,43 +49,67 @@ struct DataGram {
...
+  send_done = 1 << 1,
+};
+
 // Only allow one instance to use espnow.
 static  ResourcePool<int, kInvalidEspNow> espnow_pool(
   0
 );
 
-static SemaphoreHandle_t tx_sem;
 static esp_now_send_status_t tx_status;
```
kasperl, line 68: It feels like this could be in the resource, but maybe that is a bit hard to do.

src/resources/espnow.cc
```diff
@@ -157,10 +191,10 @@ static void free_datagram(struct DataGram* datagram) {
 
 static void espnow_send_cb(const uint8_t *mac_addr, esp_now_send_status_t status) {
   tx_status = status;
```
kasperl, line 193: Which thread calls this? The wifi thread?

kasperl (approved):
Looks good to me.

src/resources/espnow.cc
floitsch: Wanted to make it similar to the `kDataAvailableState` and `KSendDoneState` but really not worth it.
done.

src/resources/espnow.cc
floitsch: This variable is set from a callback (`espnow_send_cb`) and there is no way to get the resource in that function.

src/resources/espnow.cc
floitsch: Leaving as is, at this code will be changed in a depending PR that fixes this.

src/resources/espnow.cc
floitsch: Yes. On a "high-priority Wi-Fi task".
Will add a comment in a follow-up PR.

PR #1704, opened 2023-07-17
Allocate receive array in primitive.

floitsch:
Also only allocate the memory we actually need.


floitsch:
currently fails because we are running out of iram...

src/resources/espnow.cc
```diff
@@ -468,34 +468,47 @@ PRIMITIVE(send_succeeded) {
...
   if (ret != pdTRUE) {
+    // Should not happen: there is only one process owning this resource,
+    // and there can't be two tasks executing a primitive call at the same
+    // time.
+    ESP_LOGE("ESPNow", "Didn't get peeked queue entry");
     return process->null_object();
   }
 
-  data->resize_external(process, datagram->len);
+  datagram_pool->release(actual);
```
kasperl, line 504: Not entirely sure what this is yet, but maybe I'll get there when I review more of the PRs :)

kasperl (approved):
Looks good to me.

PR #1711, opened 2023-07-24
Make RootCertificate a real object.

erikcorry:
This is a response to the comment on https://github.com/toitware/toit-cert-roots/pull/18

It makes an incompatible change that makes a public method private, but nobody has started using the global trusted roots yet afaik.

An upcoming change to toit-cert-roots will make use of this new type.

### DIFF
--- a/lib/tls/certificate.toit
+++ b/lib/tls/certificate.toit
@@ -41,7 +41,7 @@ class RootCertificate:
   Constructs a RootCertificate from a binary DER-endoded certificate
     or an ASCII PEM-encoded certificate.
 
-  The $raw certificate, is in unparsed form, ie either in PEM (ASCII) format or
+  The $raw certificate is in unparsed form, either in PEM (ASCII) format or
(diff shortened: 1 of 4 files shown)


lib/tls/certificate.toit
```diff
@@ -22,6 +23,59 @@ class Certificate:
...
+  raw ::= ?
+
+  stringify -> string:
+    return name or "Root certificate w/ fingerprint $fingerprint"
+
+  /**
+  Constructs a RootCertificate from a binary DER-endoded certificate
+    or an ASCII PEM-encoded certificate.
+
+  The $raw certificate, is in unparsed form, ie either in PEM (ASCII) format or
```
floitsch, line 44:
```suggestion
  The $raw certificate is in an unparsed form; that is either in PEM (ASCII) format or
```

floitsch (approved):
LGTM.
once the accidental submodule change has been reverted.

PR #1716, opened 2023-07-26
Deprecate the certificate roots in the SDK

erikcorry:


floitsch (approved):
LGTM once the third-party change has been reverted.

PR #1722, opened 2023-08-02
Zlib primitives and library.

erikcorry:
This is libs and primitives for zlib, which means programs can use the deflate algorithm.

It is not currently made available on embedded platforms because it uses too much
memory and bloats the VM. This may become a config option at some time.

It is not currently available on Windows or 32 bit Linux since the OSs tend to ship without
those dynamic libraries, so it would prevent the Toit VM from booting.

It works on MacOS and 64 bit Linux. Those platforms ship with the dynamic library.

The Toit VM already has primitives to support run length encoding in a format that
is compatible with zlib. Those are available on all platforms. The classes for that support
have the read and write methods on the same object. In this case I chose not to do that
because it can cause confusion as to the meaning of the close method. Therefore the
zlib.Encoder and zlib.Decoder have a separate reader.


src/CMakeLists.txt
```diff
@@ -186,6 +186,10 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   set(TOIT_BLUETOOTH_LIBS "-framework Foundation" "-framework CoreBluetooth" ObjC)
 endif()
 
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin" OR (${CMAKE_SYSTEM_NAME} MATCHES "Linux" AND NOT CMAKE_SIZEOF_VOID_P EQUAL 4))
```
floitsch, line 189: This should probably use `FindZLIB` with `ZLIB_USE_STATIC_LIBS` to `ON`.
https://cmake.org/cmake/help/latest/module/FindZLIB.html

The way it's written now, it will use the zlib library if it finds it, and otherwise ignore it.
You can force the inclusion with `find_package(ZLIB REQUIRED)`.
On Linux64 the `ZLIB_US_STATIC_LIBS` is probably not needed either.

On the buildbot we should now ensure that the Windows bot has a static version of the zlib.

``` diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index c8195a3a..d8f3a6cd 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -13,6 +13,8 @@
 # The license can be found in the file `LICENSE` in the top level
 # directory of this repository.
 
+cmake_minimum_required(VERSION 3.23.0)
+
 if (NOT "${TOIT_SYSTEM_NAME}" MATCHES "esp32")
   add_custom_target(build_tools)
   add_subdirectory(compiler)
@@ -186,6 +188,13 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   set(TOIT_BLUETOOTH_LIBS "-framework Foundation" "-framework CoreBluetooth" ObjC)
 endif()
 
+set(ZLIB_USE_STATIC_LIBS "ON")
+find_package(ZLIB)
+if (ZLIB_FOUND)
+  include_directories(${ZLIB_INCLUDE_DIRS})
+  set(TOIT_LINK_LIBS ${TOIT_LINK_LIBS} ${ZLIB_LIBRARIES})
+endif(ZLIB_FOUND)
+
 # Because of the `CACHE INTERNAL ""` at the end of the `set` we can
 #   use this variable outside of the directory.
 set(TOIT_LINK_LIBS
```

src/top.h
```diff
@@ -104,6 +104,9 @@
 #define CONFIG_TOIT_BYTE_DISPLAY 1
 #define CONFIG_TOIT_BIT_DISPLAY 1
 #define CONFIG_TOIT_FONT 1
+#if !defined(TOIT_WINDOWS) and !defined(BUILD_32)
```
floitsch, line 107: This should not be determined here.
The CMakeLists.txt should set the constant depending on whether it found the zlib.
```cmake
if (ZLIB_FOUND)
  add_definitions(-DCONFIG_TOIT_FULL_ZLIB)
endif()
```

src/primitive_zlib.cc
```diff
@@ -107,4 +113,173 @@ PRIMITIVE(rle_finish) {
...
+PRIMITIVE(zlib_init_deflate) {
+#ifndef CONFIG_TOIT_FULL_ZLIB
+  FAIL(UNIMPLEMENTED);
+#else
+  ARGS(SimpleResourceGroup, group, int, compression_level)
+  ByteArray* proxy = process->object_heap()->allocate_proxy();
+  Zlib* zlib = _new Zlib(group);
+  if (!zlib) FAIL(MALLOC_FAILED);
+  int result = zlib->init_deflate(compression_level);
+  if (result < 0) return zlib_error(process, result);
```
floitsch, line 216: Don't we need to delete the `zlib` class here?

src/primitive_zlib.cc
```diff
@@ -107,4 +113,173 @@ PRIMITIVE(rle_finish) {
...
+PRIMITIVE(zlib_init_inflate) {
+#ifndef CONFIG_TOIT_FULL_ZLIB
+  FAIL(UNIMPLEMENTED);
+#else
+  ARGS(SimpleResourceGroup, group);
+  ByteArray* proxy = process->object_heap()->allocate_proxy();
+  Zlib* zlib = _new Zlib(group);
+  if (!zlib) FAIL(MALLOC_FAILED);
+  int result = zlib->init_inflate();
+  if (result < 0) return zlib_error(process, result);
```
floitsch, line 231: Don't we need to delete `zlib` here?

floitsch (approved):
LGTM.

floitsch (approved):
yield->wait change LGTM.

lib/zlib.toit
```diff
@@ -199,6 +195,11 @@ abstract class Coder_:
   zlib_ ::= ?
   closed_write_ := false
   closed_read_ := false
+  signal_ /monitor.Signal := monitor.Signal
+  state_/int := STATE_READY_TO_READ_ | STATE_READY_TO_WRITE_
+
+  static STATE_READY_TO_READ_ ::= 1
+  static STATE_READY_TO_WRITE_ ::= 2
```
floitsch, line 202:
```
static STATE_READY_TO_READ_ ::= 1 << 0
static STATE_READY_TO_WRITE_ ::= 1 << 1
```

Makes it easier to see that it's used as bitfield.

lib/zlib.toit
```diff
@@ -207,21 +208,53 @@ abstract class Coder_:
...
+    while result and wait and result.size == 0:
+      state_ &= ~STATE_READY_TO_READ_
+      signal_.wait: state_ & STATE_READY_TO_READ_ != 0
+      result = zlib_read_ zlib_
+    state_ |= STATE_READY_TO_WRITE_
+    signal_.raise
+    return result
 
   close_read_ -> none:
+    state_ |= STATE_READY_TO_WRITE_
```
floitsch, line 228: Probably want to make this a critical_do section.
If a `task` is canceled, you still want to uninit_, and without a critical section that might not happen. iirc.

lib/zlib.toit
```diff
@@ -207,21 +208,53 @@ abstract class Coder_:
...
+          signal_.wait: state_ & STATE_READY_TO_WRITE_ != 0
+      else:
+        state_ |= STATE_READY_TO_READ_
+        signal_.raise
+      if not wait: return bytes_written
+      pos += bytes_written
+    return pos
+
+  close -> none:
+    if not closed_write_:
```
floitsch, line 252: Might need a critical do.

PR #1726, opened 2023-08-03
Make RLE encoder more robust.

erikcorry:
Allow double close and move the read method to a new reader class.


lib/zlib.toit
```diff
@@ -8,23 +8,45 @@ import crypto
...
   static MINIMAL_GZIP_HEADER_ ::= [0x1f, 0x8b, 8, 0, 0, 0, 0, 0, 0, 0xff]
 
   constructor --gzip_header/bool:
+    reader.owner_ = this
     header := gzip_header ? MINIMAL_GZIP_HEADER_ : SMALL_BUFFER_DEFLATE_HEADER_
     channel_.send (ByteArray header.size: header[it])
 
+  /**
+  Deprecated.
+  Use the $reader to get an object with a read method.
```
floitsch, line 36: Add comment that this should also remove the `implements reader.Reader` when the method is removed.

floitsch (approved):
LGTM.

PR #1727, opened 2023-08-04
Improvements to DNS errors.

erikcorry:
Improvements to DNS errors. ￼…

If the last server gives a timeout then we use
the more useful message that the earlier DNS
servers sent. This hopefully improves the
flaky test on Mac where one of the servers
refuses to answer.

Puts the looked up name in the most common DNS
errors (the ones that are not protocol errors).


floitsch (approved):
LGTM.

PR #1732, opened 2023-08-07
Implement dash-identifiers.

floitsch:


src/compiler/scanner.h
```diff
@@ -129,15 +129,124 @@ static inline bool is_binary_digit(int c) {
...
+    for (int i = 1; i < len - 1; i++) {
+      uint8 c = identifier[i];
+      if (c != '_') continue;
+
+      uint8 previous = identifier[i - 1];
+      uint8 next = identifier[i + 1];
+      if (previous == '-' || previous == '_') continue;
+      if (next == '-' || next == '_') continue;
+
+      copy[i] = '-';
```
kasperl, line 199: You could combine these two loops and have a check to see if you've already created the copy here. 

```
if (!copy) {
  copy = reinterpret_cast<uint8*>(malloc(len + 1));
  memcpy(copy, identifier, len);
  copy[len] = '\0';
}

File: src/compiler/scanner.h [hunk 2]
```diff
@@ -129,15 +129,124 @@ static inline bool is_binary_digit(int c) {
...
+      if (previous == '-' || previous == '_') continue;
+      if (next == '-' || next == '_') continue;
+      needs_canonicalization = true;
+      break;
+    }
+    if (!needs_canonicalization) return identifier;
+
+    uint8* copy = reinterpret_cast<uint8*>(malloc(len + 1));
+    memcpy(copy, identifier, len);
+    // The terminating character might not have been present in the
```
kasperl, line 185: I don't understand this comment. You only get the terminating \0 if the string needs canonicalization, so this seems a bit brittle. Where do you use this from where the input has no terminating \0, but you need the output to have it?

src/compiler/token.h
```diff
@@ -197,10 +197,10 @@ enum Precedence {
   I(List_)                                                     \
   I(ListSlice_)                                                \
   I(Tombstone_)                                                \
-  I(create_array_)                                             \
-  I(create_byte_array_)                                        \
-  I(create_list_literal_from_array_)                           \
-  I(create_cow_byte_array_)                                    \
+  IN(create_array_, "create-array_")                           \
+  IN(create_byte_array_, "create-byte-array_")                 \
+  IN(create_list_literal_from_array_, "create-list-literal-from-array_") \
```
kasperl, line 202: Maybe just bite the bullet and re-align the \'s?

src/compiler/compiler.cc
```diff
@@ -1063,120 +1065,148 @@ static const uint8* wrap_direct_script_expression(const char* direct_script, Dia
...
-      auto note_path1 = build_error_path(toit_path);
-      auto note_path2 = build_error_path(toit_in_dir);
-      note_fun = [=]() {
-        diagnostics()->report_note(note_node,
-                                    "Missing library file. Tried '%s' and '%s'",
-                                    note_path1.c_str(),
-                                    note_path2.c_str());
-      };
+      case AddSegmentResult::NOT_A_DIRECTORY:
+        diagnostics->report_note(note_node, "Not a folder: '%s'", note_path.c_str());
```
kasperl, line 1200: It is not a big deal, but I think the notes could be more uniform. Sometimes the path is at the end, sometimes it is in the middle. Sometimes it starts with 'Not' or 'Missing', sometimes it resembles a proper sentence.

kasperl (approved):
Looks good to me.

PR #1741, opened 2023-08-10
Remove debug print.

floitsch:


kasperl (approved):
Looks good to me, but maybe clean up the "converting" printing in line 99 too, so it is more consistent with the other prints that starts with a capitalized word and end with .?

floitsch:
I removed that line too.
At that point it only looks at the file, but doesn't convert it yet.

PR #1745, opened 2023-08-11
Allow non-Toit files to be renames.

erikcorry:
Remove strange line in help text.


floitsch (approved):
LGTM.

floitsch:
Might be interesting to make this a flag but I think it's OK this way.

PR #1762, opened 2023-08-18
Robustify DNS lookup.

erikcorry:
When there are multiple DNS servers available we
were previously unclear about whether they were
there because they knew about different domains,
or whether they were there in case some servers
were unreachable.

We now follow the way Linux does things, and use
them for redundancy, not because they have different DNS databases.  That simplifies things, because we immediately return a negative answer instead of
waiting for DNS servers that may not be available.

Previously we were getting test flakiness if one
server failed to respond for a few seconds. Now
we keep retrying for the full timeout (20s).  With the previous semantics this change would have caused very slow answers in the case where a non-existent domain was looked up and one of the servers was offline.

### DIFF
--- a/lib/net/modules/dns.toit
+++ b/lib/net/modules/dns.toit
@@ -218,7 +218,7 @@ class DnsClient:
     with-timeout timeout:  // Typically a 20s timeout.
       // We try servers one at a time, but if there was a good error
       // message from one server (eg. no such domain) we let that
-      // error unwind the stack and not try the next server.
+      // error unwind the stack and do not try the next server.
       unwind-block := : | exception |
         not is-server-reachability-error_ exception
       while true:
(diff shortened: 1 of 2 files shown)


lib/net/modules/dns.toit
```diff
@@ -210,41 +215,22 @@ class DnsClient:
...
-      servers-failed := {}
+    with-timeout timeout:  // Typically a 20s timeout.
       // We try servers one at a time, but if there was a good error
-      // message from one server (eg. no such domain) we save that
-      // error and rethrow it if the last one times out with no usable
-      // response.
-      saved-error := null
-      saved-trace := null
+      // message from one server (eg. no such domain) we let that
+      // error unwind the stack and not try the next server.
```
kasperl, line 221:
```suggestion
      // error unwind the stack and do not try the next server.
```

kasperl (approved):
Looks good to me.

PR #1763, opened 2023-08-18
Don't use Go cache on Windows CI runs

erikcorry:
For some reason, restoring the cache, which only takes a few seconds on Linux, is very slow on Windows.

Eg. here it took over 5 minutes, and the build time was 2m3s:
https://github.com/toitlang/toit/actions/runs/5901634021/job/16009569093

Here we skipped it, and the build was still only 3m20s:
https://github.com/toitlang/toit/actions/runs/5902084215/job/16009436231


kasperl (approved):
Looks good to me.

PR #1765, opened 2023-08-22
Don't call Cache_Flush on ESP32.

erikcorry:


src/flash_registry_esp32.cc
```diff
@@ -117,7 +101,6 @@ int FlashRegistry::erase_chunk(int offset, int size) {
 bool FlashRegistry::write_chunk(const void* chunk, int offset, int size) {
   esp_err_t result = esp_partition_write(allocations_partition, offset, chunk, size);
   if (result == ESP_OK) {
```
kasperl, line 103:
```suggestion
  return result == ESP_OK;
```

kasperl (approved):
Looks good to me.

PR #1767, opened 2023-08-22
Round the offsets of metadata sections to cache line.

erikcorry:
This should ensure that different processes are not sharing cache lines when they write to metadata, eg. the remembered set.

### DIFF
--- a/src/third_party/dartino/gc_metadata.cc
+++ b/src/third_party/dartino/gc_metadata.cc
@@ -15,7 +15,7 @@
 namespace toit {
 
 #ifdef TOIT_FREERTOS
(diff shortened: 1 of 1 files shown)


src/third_party/dartino/gc_metadata.cc
```diff
@@ -14,6 +14,15 @@
 
 namespace toit {
 
+#ifdef TOIT_FREERTOS
+// ESP32 has a cache line is 32 bytes, as does Cortex M7.
```
kasperl, line 18: wat

```suggestion
// ESP32 has a cache line is 32 bytes, as does Cortex M7.
```

kasperl (approved):
Looks good to me.

PR #1778, opened 2023-08-25
Update third-parties of compiler.

floitsch:
Newer CMake versions complain if the minimum is too low. Just bumping these up was enough.


kasperl (approved):
Looks good to me.

PR #1780, opened 2023-08-25
Separate downloading from building

erikcorry:
I would really like a faster "make".  This gets it down from about 5s to 3.8s.

Perhaps it's not the right approach - an alternative solution would also be nice.

This also means you can build when the network is down or overloaded, which sadly happens a lot at the office.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -179,6 +179,11 @@ jobs:
           echo "SHARD_START=${{ matrix.shard }}" >> $GITHUB_ENV
           echo "SHARD_STRIDE=${{ steps.constants.outputs.total_shards }}" >> $GITHUB_ENV
 
+      - name: Sync Toit packages
+        shell: bash
+        run: |
+          make sync-packages
+
       # Build using make.
       - name: Build sdk
         shell: bash
@@ -422,6 +427,11 @@ jobs:
         run: |
           sh third_party/esp-idf/install.sh
 
+      - name: Sync Toit packages
+        shell: bash
+        run: |
+          make sync-packages
+
       # Build using make.
       - name: make
         env:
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -140,6 +148,20 @@ tools: host-tools check-env $(BUILD)/$(TARGET)/CMakeCache.txt
 toit-tools: tools download-packages
 	(cd $(BUILD)/$(TARGET) && ninja build_toit_tools)
 
+# Rebuilds the SDK using only Ninja, without rebuilding the
+# Ninja files with Cmake.
+.PHONY: fast
+fast: fast-tools fast-toit-tools
```
floitsch, line 154: typically we have `-no-X` versions of the clauses. For example `esp32-no-env` for the `esp32` rule.
I think `tools-no-sync` would fit in that category. Also makes the command-line completion easier.

tools/toit.cmake
```diff
@@ -120,5 +120,4 @@ macro(toit_project NAME PATH)
     DEPENDS "${TOITPKG}"
   )
   add_dependencies(download_packages "${DOWNLOAD_TARGET_NAME}")
-  add_dependencies("${DOWNLOAD_TARGET_NAME}" sync_packages)
```
floitsch, line 123: Make this dependent on a CMake flag.
You could then just `make disable-auto-sync` in your local check-out, but keep the sync for typical applications.

floitsch: See https://github.com/toitlang/toit/pull/1781

.github/workflows/ci.yml
```diff
@@ -179,6 +179,11 @@ jobs:
           echo "SHARD_START=${{ matrix.shard }}" >> $GITHUB_ENV
           echo "SHARD_STRIDE=${{ steps.constants.outputs.total_shards }}" >> $GITHUB_ENV
 
+      - name: Sync Toit packages
```
floitsch, line 182: These changes should not be necessary if you implement the auto-sync flag.

Makefile
floitsch: Obviously should be `tools-no-cmake`.

PR #1784, opened 2023-08-29
Disable SPIRAM on S3 for now due to #1785

erikcorry:


erikcorry:
See issue #1785

kasperl (approved):
What does this mean for the spiram-octo variant? Does the patch apply?

floitsch:
> What does this mean for the spiram-octo variant? Does the patch apply?

Just tested it, and, unfortunately, no. Does not apply.
This change requires an update to the `envelopes` repository.

floitsch:
That said: is the octo variant even meaningful if there is no spiram?

kasperl:
> That said: is the octo variant even meaningful if there is no spiram?

No, it isn't so maybe we should just drop it until we can bring it back later?

PR #1794, opened 2023-09-05
Reduce only-for-malloc area from 160k to 108k

erikcorry:
For non-SPIRAM ESP32s, this increases the size of the heap metadata from 12k to 16k, but may help for systems that have large Toit heaps and low amounts of malloced memory.


kasperl (approved):
Looks good to me.

PR #1798, opened 2023-09-06
Clean up compression classes.

erikcorry:
Uniformly uses Signal instead of Channel for the
compression channels.

Gives them all the --wait argument on reads and
writes so they can be used from one task in non-blocking mode or from two tasks in blocking mode.

Uniform availability of from= and to= on write methods.

Removes the deprecated read method - now have to use reader.read.

### DIFF
--- a/lib/zlib.toit
+++ b/lib/zlib.toit
@@ -8,99 +8,86 @@ import crypto
 import crypto.adler32 as crypto
 import crypto.crc as crc-algorithms
 
-class RleReader implements reader.Reader:
-  owner_ /ZlibEncoder_? := null
+class CompressionReader implements reader.Reader:
+  owner_ := null
 
   constructor.private_:
 
-  read:
-    return owner_.read_
+  read --wait/bool=true -> ByteArray?:
+    return owner_.read_ --wait=wait
 
   close:
-    // Currently does nothing.
+    owner_.close-read_
 
-// TODO: Should not be a Reader, but we have deprecated the read method,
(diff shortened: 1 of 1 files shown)


lib/zlib.toit
```diff
@@ -8,99 +8,86 @@ import crypto
...
-  read_:
-    result := channel_.receive
-    if not result: channel_.send null  // Once it's closed we should be able to keep reading nulls.
-    return result
+SMALL-BUFFER-DEFLATE-HEADER_ ::= [8, 0x1d]
+MINIMAL-GZIP-HEADER_ ::= [0x1f, 0x8b, 8, 0, 0, 0, 0, 0, 0, 0xff]
 
-class UncompressedDeflateEncoder_ extends ZlibEncoder_:
-  summer_/crypto.Checksum? := ?
+class UncompressedDeflateBackEnd_ implements BackEnd_:
```
floitsch, line 25: Add a bit of documentation, to help the reader.
Basically, pipes the data through in 256byte blocks. Needs to maintain a checksum...

lib/zlib.toit
```diff
@@ -8,99 +8,86 @@ import crypto
 import crypto.adler32 as crypto
 import crypto.crc as crc-algorithms
 
-class RleReader implements reader.Reader:
-  owner_ /ZlibEncoder_? := null
+class CompressionReader implements reader.Reader:
+  owner_ := null
```
floitsch, line 12: I tend to call this `wrapped`.

floitsch (approved):
LGTM.

PR #1801, opened 2023-09-12
Add WiFi tear-down debug printing.

kasperl:


floitsch (approved):
LGTM.
I wonder if we should use the logging mechanism of the esp-idf. It has the advantage that it can be redirected. (Not sure that's also true for `printf`).

PR #1803, opened 2023-09-12
Don't malloc(0)

floitsch:
malloc(0) is implementation dependent and can return a pointer or null. Since we check for 'null' to know whether the malloc failed we are better off not even calling 'malloc(0)'.

### DIFF
--- a/src/messaging.cc
+++ b/src/messaging.cc
@@ -660,6 +660,10 @@ bool MessageDecoder::decode_byte_array_external(void** data, int* length) {
     return true;
   } else if (tag == TAG_BYTE_ARRAY_INLINE) {
     int encoded_length = *length = read_cardinal();
+    if (encoded_length == 0) {
+      *data = null;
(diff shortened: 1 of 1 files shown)


kasperl (approved):
Looks good to me.

src/messaging.cc
```diff
@@ -660,7 +660,8 @@ bool MessageDecoder::decode_byte_array_external(void** data, int* length) {
     return true;
   } else if (tag == TAG_BYTE_ARRAY_INLINE) {
     int encoded_length = *length = read_cardinal();
-    void* copy = malloc(encoded_length);
+    int malloc_length = encoded_length == 0 ? 1 : encoded_length;
```
kasperl, line 663: Maybe add a comment to explain why this is and maybe use `Utils::max(1, encoded_length)`? Not sure the extra `malloc_length` is worth it.

PR #1804, opened 2023-09-13
Rework the WiFi start and stop sequences.

kasperl:
Delay the call to esp_wifi_connect() until we've been told that the WiFi has started.

Remember to call esp_wifi_disconnect() if we're connected when the WiFi is being torn down.

### DIFF
--- a/src/resources/wifi_esp32.cc
+++ b/src/resources/wifi_esp32.cc
@@ -97,16 +97,16 @@ class WifiResourceGroup : public ResourceGroup {
     err = esp_wifi_set_config(WIFI_IF_STA, &config);
     if (err != ESP_OK) return err;
 
-    err = esp_wifi_start();
-    if (err != ESP_OK) return err;
-
     // When connecting to Android mobile hotspot APs, we
     // quite often get WIFI_REASON_AUTH_FAIL followed by
     // WIFI_REASON_CONNECTION_FAIL. The next connect still
     // has a good chance of succeeding, so we allow two
     // reconnect attempts.
     reconnects_remaining_ = 2;
-    return esp_wifi_connect();
+
+    // Request to start the WiFi stack. We will try to connect to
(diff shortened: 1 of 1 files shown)


src/resources/wifi_esp32.cc
```diff
@@ -156,36 +156,17 @@ class WifiResourceGroup : public ResourceGroup {
...
     printf("[wifi] ~WifiResourceGroup()\n");
-    printf("[wifi] esp_wifi_deinit()\n");
-#endif
-    esp_err_t err = esp_wifi_deinit();
-#if CONFIG_WPA_DEBUG_PRINT
-    printf("[wifi] esp_wifi_deinit() -> done\n");
-#endif
-    if (err == ESP_ERR_WIFI_NOT_STOPPED) {
-#if CONFIG_WPA_DEBUG_PRINT
-      printf("[wifi] esp_wifi_stop()\n");
```
kasperl, line 167: Stopping is now taken care of when tearing down the `WifiEvents` resource. That resource only exists if we've tried to start the WiFi stack.

src/resources/wifi_esp32.cc
```diff
@@ -219,21 +200,63 @@ class WifiResourceGroup : public ResourceGroup {
...
+
   TAG(WifiEvents);
   explicit WifiEvents(WifiResourceGroup* group)
       : SystemResource(group, WIFI_EVENT)
-      , disconnect_reason_(WIFI_REASON_UNSPECIFIED) {}
+      , disconnect_reason_(WIFI_REASON_UNSPECIFIED)
+      , state_(STOPPED) {}
 
   ~WifiEvents() {
-    FATAL_IF_NOT_ESP_OK(esp_wifi_stop());
```
kasperl, line 228: We keep track of enough state to allow us to disconnect if necessary.

floitsch (approved):
LGTM.

PR #1808, opened 2023-09-14
Rewrite DNS code to get other record types

erikcorry:


lib/net/modules/dns.toit
```diff
@@ -356,7 +455,8 @@ parts_ packet/ByteArray position/int parts/List [position-block] -> none:
   position-block.call position + 1
 
 /// Limits the size of the cache to avoid using too much memory.
-trim-cache_ cache/Map -> none:
+trim-cache_ map/Map type/int -> none:
```
kasperl, line 458: Could this be static?

lib/net/modules/dns.toit
```diff
@@ -303,31 +380,53 @@ class DnsClient:
...
+          len := type == RECORD-A ? 4 : 16
+          if rd-length != len: throw (DnsException "Unexpected IP address length $rd-length")
+          result.add
+              net.IpAddress
+                  response.copy position position + len
+        else if type == RECORD-CNAME or type == RECORD-PTR:
+          result.add
+              decode-name response position: null
+        else if type == RECORD-TXT:
+          if response[position] != rd-length - 1: throw "WTF"
```
kasperl, line 398: Maybe improve the thrown exception here.

lib/net/modules/dns.toit
```diff
@@ -249,42 +319,49 @@ class DnsClient:
...
-      if cache_.contains name:
-        entry := cache_[name]
-        if entry.valid: return entry.address
-        cache_.remove name
-    if accept-ipv6:
-      if cache-ipv6_.contains name:
-        entry := cache-ipv6_[name]
-        if entry.valid: return entry.address
-        cache-ipv6_.remove name
+  find-in-cache_ name type/int -> List?:
```
kasperl, line 322: static?

lib/net/modules/dns.toit
```diff
@@ -249,42 +319,49 @@ class DnsClient:
...
+    if cache_.contains type:
+      map := cache_[type]
+      if map.contains name:
+        entry := map[name]
+        if entry.valid:
+          return entry.value
+        else:
+          map.remove name
```
kasperl, line 330: Not sure it is much better, but:

```suggestion
    map := cache_.get type --if-absent=: return null
    entry := map.get name --if-absent=: return null
    if entry.valid: return entry.value
    map.remove name
    return null
```

lib/net/modules/dns.toit
```diff
@@ -303,31 +380,53 @@ class DnsClient:
...
-        if case-compare_ r-name q-name:
-          result := net.IpAddress
-              response.copy position position + 4
-          if ttl > 0:
-            trim-cache_ cache_
-            cache_[query.name] = CacheEntry_ result ttl
-          return result
+      if type == overall-type and case-compare_ r-name q-name:
+        if type == RECORD-A or type == RECORD-AAAA:
+          len := type == RECORD-A ? 4 : 16
```
kasperl, line 389: You use `length` in other places.

lib/net/modules/dns.toit
```diff
@@ -303,31 +380,53 @@ class DnsClient:
...
+          if rd-length != len: throw (DnsException "Unexpected IP address length $rd-length")
+          result.add
+              net.IpAddress
+                  response.copy position position + len
+        else if type == RECORD-CNAME or type == RECORD-PTR:
+          result.add
+              decode-name response position: null
+        else if type == RECORD-TXT:
+          if response[position] != rd-length - 1: throw "WTF"
+          len := min (rd-length - 1) response[position]
```
kasperl, line 399: You use `length` in other places.

lib/net/modules/dns.toit
```diff
@@ -303,31 +380,53 @@ class DnsClient:
...
         q-name = decode-name response position: null
       position += rd-length
-    throw (DnsException "Response did not contain matching A record" --name=query.name)
+    if result.size > 0:
+      if min-ttl > 0:
+        trim-cache_ cache_ overall-type
+        type-cache := cache_.get overall-type --init=: {:}
+        type-cache[query.name] = CacheEntry_ result min-ttl
+      return result
+    return []
```
kasperl, line 421:
```suggestion
    if result.size > 0 and min-ttl > 0:
      trim-cache_ cache_ overall-type
      type-cache := cache_.get overall-type --init=: {:}
      type-cache[query.name] = CacheEntry_ result min-ttl
    return result
```

kasperl (approved):
Looks good to me.

lib/net/modules/dns.toit
erikcorry: I moved it into the class.

lib/net/modules/dns.toit
erikcorry: Done.

lib/net/modules/dns.toit
erikcorry: Uses the cache, which is a field.  Same applies to trim-cache_.

lib/net/modules/dns.toit
```diff
@@ -39,7 +60,7 @@ If there are multiple servers then they are tried in rotation until one
   the next one.  This is in line with the way that Linux handles multiple
   servers on the same lookup request.
 */
-dns-lookup -> net.IpAddress
+dns-lookup-ips -> List
```
kasperl, line 63: Maybe `-all` or `-multi` instead of `-ips`?

erikcorry: I like -multi

PR #1809, opened 2023-09-17
Add some util functions to net/modules/dns.toit that will be useful for mdns

erikcorry:
dns.toit is just for simple name lookups.

dns-utils has more advanced features and low level tools that will be useful for mDNS implementation.


lib/net/modules/dns.toit
```diff
@@ -183,7 +186,7 @@ ERROR-NAMES ::= {
 class DnsQuery_:
   base-id/int
   name/string
-  query-packets/Map  // from record type (int) to packet (ByteArray)
+  query-packets/Map  // From record type (int) to packet (ByteArray)
```
kasperl, line 189:
```suggestion
  query-packets/Map  // From record type (int) to packet (ByteArray).
```

lib/net/modules/dns-tools.toit
```diff
@@ -0,0 +1,311 @@
+// Copyright (C) 2023 Toitware ApS. All rights reserved.
```
kasperl, line 1: Not sure I like 'tools' so much. I don't think of such helper functions and classes as tools.

kasperl (approved):
Looks good to me.

toolchains/esp32/sdkconfig
```diff
@@ -1622,6 +1622,7 @@ CONFIG_TOIT_BYTE_DISPLAY=y
 CONFIG_TOIT_BIT_DISPLAY=y
 CONFIG_TOIT_FONT=y
 CONFIG_TOIT_FATFS=y
+CONFIG_TOIT_FULL_ZLIB=y
```
kasperl, line 1625: This doesn't seem right.

toolchains/esp32/sdkconfig.defaults
```diff
@@ -77,4 +77,5 @@ CONFIG_SPI_FLASH_VERIFY_WRITE=y
 CONFIG_SPI_FLASH_LOG_FAILED_WRITE=y
 CONFIG_SPI_FLASH_WARN_SETTING_ZERO_TO_ONE=y
 # CONFIG_WS_TRANSPORT is not set
+CONFIG_TOIT_FULL_ZLIB=y
```
kasperl, line 80: This doesn't seem right.

kasperl (changes requested):

PR #1818, opened 2023-09-19
Fix some situations in which the zlib would deadlock.

erikcorry:


kasperl (approved):
Looks good to me.

Is it hard to write a test for?

PR #1828, opened 2023-09-22
Readdir assumed it was not running on embedded

erikcorry:
In a separate change we should rename primitive_file_posix.cc


floitsch (approved):
LGTM.

PR #1831, opened 2023-09-25
Rename primitive_file_posix.cc to primitive_file_posix_and_esp32.cc

erikcorry:
This file is used on embedded, but its filename was confusing leading to the assumption that code in
this file was not going to run on embedded platforms where malloc can fail.


src/resources/spi_linux.cc
```diff
@@ -27,7 +27,7 @@
 
 namespace toit {
 
-// Defined in primitive_file_posix.cc.
+// Defined in primitive_file_posix_and_esp32.cc.
```
kasperl, line 30: Not super excited about the `and` in the filename, but I can't think of something better.

kasperl: We would probably use `shared` if it was for all platforms. Maybe there is a term we can introduce that covers more here.

Alternatively, we can think about splitting it out (have a dedicated `primitive_file_esp32`) and share code in a more granular way. I suggest we wait with that until we bring up support for another platform like Zephyr.

kasperl (approved):

src/resources/spi_linux.cc
erikcorry: primitive_file_non_windows.cc
?

erikcorry: Or primitive_file_non_win.cc to fit with other windows-related source files.

PR #1832, opened 2023-09-26
Fix bug in our TLS socket creation around OOM.

erikcorry:
We paniced instead of reportint OOM for a malloc
failure at a bad moment.

### DIFF
--- a/src/resources/tls.cc
+++ b/src/resources/tls.cc
@@ -347,11 +347,11 @@ bool is_tls_malloc_failure(int err) {
 }
 
 Object* tls_error(BaseMbedTlsSocket* socket, Process* process, int err) {
-  static const size_t BUFFER_LEN = 400;
-  char buffer[BUFFER_LEN];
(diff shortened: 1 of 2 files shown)


src/resources/tls.cc
```diff
@@ -343,6 +341,15 @@ Object* tls_error(BaseMbedTlsSocket* socket, Process* process, int err) {
...
       hi_error == MBEDTLS_ERR_X509_ALLOC_FAILED) {
+    return true;
+  }
+  return false;
+}
+
+Object* tls_error(BaseMbedTlsSocket* socket, Process* process, int err) {
+  static const size_t BUFFER_LEN = 400;
+  char buffer[BUFFER_LEN];
+  if (is_tls_malloc_failure(err)) {
```
kasperl, line 352: I'd probably do this check before creating the buffer.

kasperl (approved):
Looks good to me.

PR #1837, opened 2023-09-27
Windows: Don't use overlapped mode for subprocess pipes.

erikcorry:
When creating pipes for subprocesses, do not create the end that is for the subprocess in overlapped mode. This is a non- blocking mode that confuses non-Toit subprocesses.


kasperl (approved):
Looks good to me.

Can we enable any tests now?

erikcorry:
This caused other tests to fail, namely the LSP tests.

It looks like this bug was accidentally making Toit child processes work, while breaking non-Toit child processes.

So I'm working on fixing that.

kasperl:
I assume this is dead now.

PR #1839, opened 2023-09-27
Fix BLE advertising

floitsch:
When two devices connect, and a one disconnects the advertising stopped. This patch (thanks @mikkeldamsgaard) fixes this.

### DIFF
--- a/src/resources/ble_esp32.cc
+++ b/src/resources/ble_esp32.cc
@@ -1157,6 +1157,7 @@ int BlePeripheralManagerResource::_on_gap(struct ble_gap_event* event) {
           ESP_LOGW("BLE", "Could not restart advertising: err=%d", err);
         }
       }
+      break;
     case BLE_GAP_EVENT_ADV_COMPLETE:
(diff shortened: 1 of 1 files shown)


src/resources/ble_esp32.cc
```diff
@@ -1124,24 +1128,35 @@ bool BleCharacteristicResource::update_subscription_status(uint8_t indicate, uin
...
         }
       }
       break;
+    case BLE_GAP_EVENT_DISCONNECT:
+      if (advertising_started() && !ble_gap_adv_active()) {
+        int err = do_start_advertising(this);
+        if (err != BLE_ERR_SUCCESS) {
+          ESP_LOGW("BLE", "Could not restart advertising: err=%d", err);
+        }
+      }
```
kasperl, line 1159: No break?

kasperl (approved):
Looks good to me, but I'd add that break.

src/resources/ble_esp32.cc

PR #1841, opened 2023-10-02
Invalidate RTC memory on firmware updates

kasperl:


kasperl:
Looks like `esp_restart()` has changed, so it is understandable that we were surprised.

floitsch (approved):
LGTM.

PR #1843, opened 2023-10-02
Add add-first to Deque

erikcorry:


lib/core/collections.toit
```diff
@@ -3200,10 +3200,24 @@ class Deque implements Collection:
...
     backing[first] = null
-    first_ = first + 1
+    first_++
     shrink-if-needed_
     return result
 
+  add-first element -> none:
+    first := first_
+    if first == 0:
+      first = backing_.size / 2 + 1
```
floitsch, line 3210: I would prefer a different variable here.
`padding_space` or something like this.

floitsch (approved):
LGTM.

PR #1844, opened 2023-10-03
Don't allow mutations of final fields.

floitsch:
Once the 'super' has been called we must not allow changes to final fields.

Fixes https://github.com/toitware/toit/issues/4178


kasperl (approved):
Looks good to me.

PR #1845, opened 2023-10-04
Make field lookups dynamic in the instance section

floitsch:
After a super call (implicit or not), field accesses now do a dynamic lookup. In that case they might find a getter (or a different field) in a subclass.


kasperl (approved):
Looks good to me.

PR #1846, opened 2023-10-04
Disallow capturing of fields before super

floitsch:


tools/lsp/server/multiplex.toit
```diff
@@ -39,11 +39,13 @@ class MultiplexConnection:
   buffered-from-compiler_ / BufferedReader
 
   constructor .from-compiler_:
```
kasperl, line 41: Alternatively, you could just take the parameter and be explicit about the field assignment in the body. I think that is simpler to reason about, but maybe slightly less compact.

kasperl (approved):
Looks good to me.

tools/lsp/server/multiplex.toit
floitsch: I like that more.
thanks and done.

PR #1847, opened 2023-10-04
Fix Windows pipes

erikcorry:
Fix non-blocking on stdio on Windows.

We can't rely on the standard fds being in non-blocking mode on Windows,
so we use a dedicated thread for each of them.

### DIFF
--- a/src/resources/pipe_win.cc
+++ b/src/resources/pipe_win.cc
@@ -47,12 +47,12 @@ class PipeResourceGroup : public ResourceGroup {
 
   bool is_standard_piped(int fd) const {
     if (!(0 <= fd && fd <= 2)) return false;
-    return (standard_pipes_ & ( 1 << fd)) != 0;
+    return (standard_pipes_ & (1 << fd)) != 0;
   }
 
   void set_standard_piped(int fd) {
     if (0 <= fd && fd <= 2) {
-      standard_pipes_ |= ( 1 << fd);
+      standard_pipes_ |= (1 << fd);
     }
   }
   volatile long& pipe_serial_number() { return pipe_serial_number_; }
@@ -313,7 +313,7 @@ class CopyPipeState {
 };
 
 static DWORD copy_pipe_thread(void* data) {
-  auto state = reinterpret_cast<CopyPipeState*>(data)->copy_loop();
+  auto state = reinterpret_cast<CopyPipeState*>(data);
   DWORD result = state->copy_loop();
   delete state;
   return result;
@@ -402,7 +402,7 @@ PRIMITIVE(fd_to_pipe) {
   CopyPipeState* state = for_writing
       ? _new CopyPipeState(read, handle)
       : _new CopyPipeState(handle, write);
-  ASSERT(state);  // Can't fail on Windows.
+  ASSERT(state);  // Can't fail on host platforms.
   HANDLE thread = CreateThread(NULL, 0, copy_pipe_thread, state, 0, NULL);
   if (thread == NULL) {
     close_handle_keep_errno(event);
@@ -417,7 +417,7 @@ PRIMITIVE(fd_to_pipe) {
     pipe_resource = _new ReadPipeResource(resource_group, read, event);
   }
 
-  ASSERT(pipe_resource);  // Can't fail on Windows.
+  ASSERT(pipe_resource);  // Can't fail on host platforms.
 
   resource_group->set_standard_piped(fd);
 
--- a/src/resources/pipe_win.cc
+++ b/src/resources/pipe_win.cc
@@ -45,8 +45,16 @@ class PipeResourceGroup : public ResourceGroup {
   TAG(PipeResourceGroup);
   PipeResourceGroup(Process* process, EventSource* event_source) : ResourceGroup(process, event_source) {}
 
-  bool is_standard_piped(int fd) const { return (standard_pipes_ & ( 1 << fd)) != 0; }
-  void set_standard_piped(int fd) { standard_pipes_ |= ( 1 << fd); }
+  bool is_standard_piped(int fd) const {
+    if (!(0 <= fd && fd <= 2)) return false;
+    return (standard_pipes_ & ( 1 << fd)) != 0;
+  }
+
+  void set_standard_piped(int fd) {
+    if (0 <= fd && fd <= 2) {
+      standard_pipes_ |= ( 1 << fd);
+    }
+  }
   volatile long& pipe_serial_number() { return pipe_serial_number_; }
  protected:
   uint32_t on_event(Resource* resource, word data, uint32_t state) override {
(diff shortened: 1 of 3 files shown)


src/resources/pipe_win.cc
```diff
@@ -45,8 +45,16 @@ class PipeResourceGroup : public ResourceGroup {
   TAG(PipeResourceGroup);
   PipeResourceGroup(Process* process, EventSource* event_source) : ResourceGroup(process, event_source) {}
 
-  bool is_standard_piped(int fd) const { return (standard_pipes_ & ( 1 << fd)) != 0; }
-  void set_standard_piped(int fd) { standard_pipes_ |= ( 1 << fd); }
+  bool is_standard_piped(int fd) const {
+    if (!(0 <= fd && fd <= 2)) return false;
+    return (standard_pipes_ & ( 1 << fd)) != 0;
```
kasperl, line 50:
```suggestion
    return (standard_pipes_ & (1 << fd)) != 0;
```

src/resources/pipe_win.cc
```diff
@@ -45,8 +45,16 @@ class PipeResourceGroup : public ResourceGroup {
...
+  }
+
+  void set_standard_piped(int fd) {
+    if (0 <= fd && fd <= 2) {
+      standard_pipes_ |= ( 1 << fd);
```
kasperl, line 55:
```suggestion
      standard_pipes_ |= (1 << fd);
```

src/resources/pipe_win.cc
```diff
@@ -267,17 +279,50 @@ PRIMITIVE(create_pipe) {
...
+    DWORD read_count;
+    DWORD write_count;
+    while (ReadFile(from_, buffer, sizeof(buffer), &read_count, NULL) && read_count > 0) {
+      if (!WriteFile(to_, buffer, read_count, &write_count, NULL)) {
+        return 1;
+      }
+    }
+    CloseHandle(from_);
+    CloseHandle(to_);
+    delete(this);
```
kasperl, line 304: I'd add a comment here that explains why this is a good idea (I don't really think it is).

```suggestion
    delete this;
```

src/resources/pipe_win.cc
```diff
@@ -267,17 +279,50 @@ PRIMITIVE(create_pipe) {
...
+    return 0;
+  }
+
+ private:
+  HANDLE from_;
+  HANDLE to_;
+};
+
+static DWORD copy_pipe_thread(void* data) {
+  return reinterpret_cast<CopyPipeState*>(data)->copy_loop();
```
kasperl, line 314: I'd make this more sane. Cast and put in a local variable. Run the `copy_loop()` method, then `delete` it.

src/resources/pipe_win.cc
```diff
@@ -288,27 +333,86 @@ PRIMITIVE(fd_to_pipe) {
...
+    WINDOWS_ERROR;
+  }
+  if (for_writing) {
+    pipe_resource = _new WritePipeResource(resource_group, write, event);
+  } else {
+    pipe_resource = _new ReadPipeResource(resource_group, read, event);
   }
 
-  if (!pipe_resource) FAIL(MALLOC_FAILED);
+  ASSERT(pipe_resource);  // Can't fail on Windows.
```
kasperl, line 415:
```suggestion
  ASSERT(pipe_resource);  // Can't fail on host platforms.
```

src/resources/pipe_win.cc
```diff
@@ -288,27 +333,86 @@ PRIMITIVE(fd_to_pipe) {
...
+  );
+  if (write == INVALID_HANDLE_VALUE) {
+    close_handle_keep_errno(event);
+    close_handle_keep_errno(read);
+    WINDOWS_ERROR;
+  }
+  CopyPipeState* state = for_writing
+      ? _new CopyPipeState(read, handle)
+      : _new CopyPipeState(handle, write);
+  ASSERT(state);  // Can't fail on Windows.
```
kasperl, line 400:
```suggestion
  ASSERT(state);  // Can't fail on host platforms.
```

src/resources/pipe_win.cc
```diff
@@ -267,17 +279,50 @@ PRIMITIVE(create_pipe) {
...
+ public:
+  CopyPipeState(HANDLE from, HANDLE to) : from_(from), to_(to) {}
+
+  DWORD copy_loop() {
+    char buffer[4096];
+    DWORD read_count;
+    DWORD write_count;
+    while (ReadFile(from_, buffer, sizeof(buffer), &read_count, NULL) && read_count > 0) {
+      if (!WriteFile(to_, buffer, read_count, &write_count, NULL)) {
+        return 1;
```
kasperl, line 299: So we don't delete it here and we don't close the handles?

src/resources/pipe_win.cc
```diff
@@ -267,17 +279,50 @@ PRIMITIVE(create_pipe) {
...
   array->at_put(1, Smi::from(reinterpret_cast<word>(input ? read : write)));
 
   return array;
 }
 
+class CopyPipeState {
+ public:
+  CopyPipeState(HANDLE from, HANDLE to) : from_(from), to_(to) {}
+
+  DWORD copy_loop() {
```
kasperl, line 293: Do we ever use the return value?

src/resources/pipe_win.cc
```diff
@@ -267,17 +279,50 @@ PRIMITIVE(create_pipe) {
...
+    char buffer[4096];
+    DWORD read_count;
+    DWORD write_count;
+    while (ReadFile(from_, buffer, sizeof(buffer), &read_count, NULL) && read_count > 0) {
+      if (!WriteFile(to_, buffer, read_count, &write_count, NULL)) {
+        return 1;
+      }
+    }
+    CloseHandle(from_);
```
kasperl, line 302: I'd do this in a destructor.

src/resources/pipe_win.cc
```diff
@@ -288,27 +333,86 @@ PRIMITIVE(fd_to_pipe) {
...
-      pipe_resource = _new WritePipeResource(resource_group, error, event);
-      break;
-    }
-    default:
-      FAIL(INVALID_ARGUMENT);
+  HANDLE handle = reinterpret_cast<HANDLE>(_get_osfhandle(fd));
+  if (handle == INVALID_HANDLE_VALUE) WINDOWS_ERROR;
+  int type = GetFileType(handle);
+  if (type != FILE_TYPE_PIPE && type != FILE_TYPE_CHAR) {
+    FAIL(INVALID_ARGUMENT);  // Not a pipe.
```
kasperl, line 340:
```suggestion
    FAIL(INVALID_ARGUMENT);  // Ceci n'est pas une pipe.
```

src/resources/pipe_win.cc
```diff
@@ -288,27 +333,86 @@ PRIMITIVE(fd_to_pipe) {
...
+    close_handle_keep_errno(read);
+    WINDOWS_ERROR;
+  }
+  CopyPipeState* state = for_writing
+      ? _new CopyPipeState(read, handle)
+      : _new CopyPipeState(handle, write);
+  ASSERT(state);  // Can't fail on Windows.
+  HANDLE thread = CreateThread(NULL, 0, copy_pipe_thread, state, 0, NULL);
+  if (thread == NULL) {
+    close_handle_keep_errno(event);
```
kasperl, line 403: Is this the reason you don't want to call CloseHandle in a destructor on CopyPipeState? Can't you just use close_handle_keep_errno from within there?

kasperl (approved):
LGTM with a few comments.

Maybe there is something I'm missing in which case I think more comments are in order.

PR #1851, opened 2023-10-06
Use MALLOC_FAILED on external buffer allocation failures when reading…

kasperl:
… files

### DIFF
--- a/src/primitive_file_non_win.cc
+++ b/src/primitive_file_non_win.cc
@@ -264,7 +264,7 @@ PRIMITIVE(readdir) {
 
   const int MAX_VFAT = 260;  // Max filename length.
   AllocationManager allocation(process);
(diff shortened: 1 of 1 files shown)


erikcorry:
I don't think this is correct, but maybe there's a way to make the correct code more obvious.

For posterity:

The AllocationManager can refuse a malloc for two reasons.  One is that malloc fails.  In this case it sets a flag on the heap that means we do a cross process GC. Therefore there is no need to return MALLOC_FAILED, and ALLOCATION_FAILED is correct.

The second reason is that we have allocated a lot of external memory since the last GC. This can happen on non-embedded platforms. In this case we don't want to trigger a cross-process GC. We are just guarding against the pathological situation where the young space is full of 1000 small external byte array objects that are each keeping 100Mbytes of external memory alive. In this case ALLOCATION_FAILED is also correct.

PR #1852, opened 2023-10-06
Fix case where readdir could trigger cross-process GC for no reason

erikcorry:


src/primitive_file_non_win.cc
```diff
@@ -265,7 +265,7 @@ PRIMITIVE(readdir) {
   const int MAX_VFAT = 260;  // Max filename length.
   AllocationManager allocation(process);
   uint8 *backing = allocation.alloc(MAX_VFAT);
```
kasperl, line 267:
```suggestion
  uint8* backing = allocation.alloc(MAX_VFAT);
```

kasperl (approved):
Looks good to me.

PR #1854, opened 2023-10-06
Add 'io' lib and 'io.Data' interface.

floitsch:
Start using it.


erikcorry:
In general I like it.

Maybe io.Data should be in core and thus just be called Data.  Feels very core and not always related to input/output.

I have a proposal that I think will reduce code size a little and chunk up the output. https://github.com/toitlang/toit/pull/1885

lib/core/utils.toit
```diff
@@ -140,11 +142,17 @@ Seeds the random number generator with the $seed.
 The $seed must be a byte array or a string.
 Currently only the first 16 bytes of the $seed are used.
 */
-set-random-seed seed:
-  #primitive.core.random-seed
-
-random-add-entropy_ data:
-  #primitive.core.add-entropy
+set-random-seed seed -> none:
```
erikcorry, line 145: Comment out of date.

lib/core/utils.toit
```diff
@@ -140,11 +142,17 @@ Seeds the random number generator with the $seed.
...
+  #primitive.core.random-seed:
+    if it == "WRONG_BYTES_TYPE":
+      set-random-seed (ByteArray.from seed 0 (min seed.size 16))
+    else:
+      throw it
+
+random-add-entropy_ data/io.Data -> none:
+  #primitive.core.add-entropy: | error |
+    // TODO(florian): should we do this in a chunked way?
```
erikcorry, line 154: I think we should do it chunked.

erikcorry (commented):
Looks good so far.

lib/core/utils.toit
floitsch: done.

lib/core/utils.toit
floitsch: done.

tests/io-data.toit
```diff
@@ -0,0 +1,21 @@
...
+
+class FakeData implements io.Data:
+  data_ / ByteArray
+
+  constructor .data_:
+
+  constructor.str str/string:
+    data_ = str.to-byte-array
+
+  size -> int: return data_.size
```
kasperl, line 15: I'd probably put this on a new line for consistency.

src/primitive.h
```diff
@@ -951,7 +951,7 @@ Object* get_absolute_path(Process* process, const wchar_t* pathname, wchar_t* ou
   Object* _raw_##name = __args[-(N)];                                        \
   uninitialized_t _u_##name;                                                 \
   Blob name(_u_##name);                                                      \
-  if (!_raw_##name->byte_content(process->program(), &name, STRINGS_OR_BYTE_ARRAYS)) FAIL(WRONG_OBJECT_TYPE);
+  if (!_raw_##name->byte_content(process->program(), &name, STRINGS_OR_BYTE_ARRAYS)) FAIL(WRONG_BYTES_TYPE);
```
kasperl, line 954: So here BYTES / io.Data is the same thing -- or is BYTES a subset of io.Data? It feels like it would be nice if the error message made it more recognizable that this was the wrong kind of io.Data when that is the case.

I am all for cleaning this up further when we look at the exception hierarchy.

lib/io.toit
```diff
@@ -0,0 +1,58 @@
+// Copyright (C) 2023 Toitware ApS. All rights reserved.
+
+/**
+A producer of bytes.
+
+The most important implementations of this interface are
+  $ByteArray and $string. However, any data structure that can be
```
kasperl, line 7: A lot of primitives only accept those. Internally, they are called blobs. Maybe they need a name that users can relate to? Not sure if they also need a type, but probably not.

lib/encoding/hex.toit
```diff
@@ -22,9 +22,16 @@ Takes an input byte array or string
...
 ```
 */
-encode data -> string:
+encode data/io.Data -> string:
   if data.size == 0: return ""
-  if data.size == 1: return "$(%02x data[0])"
+  if data.size == 1:
+    c := ?
+    if data is ByteArray: c = (data as ByteArray)[0]
+    else if data is string: c = (data as string)[0]
```
kasperl, line 30:
```suggestion
    if data is ByteArray: 
      c = (data as ByteArray)[0]
    else if data is string: 
      c = (data as string)[0]
```

kasperl (approved):

src/primitive.h
floitsch: I agree. This should be cleaned up a bit.
In theory, the Toit type check should guarantee that only `io.Data` reaches the primitive functions.
Every primitive function that takes an `io.Data` should then deal with the `WRONG_BYTES_TYPE` immediately and convert the arguments to ByteArrays (potentially in a chunked way).
This means that the user should never see the `WRONG_BYTES_TYPE`. If we have implemented our Toit functions correctly, then it should always be used as a signal to convert, and never as an exception.

tests/io-data.toit
floitsch: done.

PR #1856, opened 2023-10-06
Fix version/sdkversion in toitlsp (and thus toitdocs)

floitsch:
We weren't using 'sdkVersion' anymore:
https://github.com/toitlang/toit/blob/9da69ab57c3f8496261706921b5881a81cc04770/tools/toitlsp/main.go#L26


kasperl (approved):
Looks good to me.

PR #1858, opened 2023-10-07
Add ir class kind.

floitsch:


src/compiler/lsp/protocol_summary.cc
```diff
@@ -396,13 +396,17 @@ void Writer::print_class(ir::Class* klass) {
...
-  if (klass->is_interface()) {
-    kind = "interface";
-  } else if (klass->is_abstract()) {
-    kind = "abstract";
-  } else {
-    kind = "class";
+  const char* kind = "";  // Initialize with value to silence compiler warnings.
+  switch (klass->kind()) {
+    case ir::Class::CLASS:
+      kind = klass->is_abstract() ? "abstract" : "class";
```
kasperl, line 402: It is a bit funky that this doesn't print 'abstract class' for abstract classes, but just 'abstract'.

kasperl (approved):
Looks good to me.

src/compiler/lsp/protocol_summary.cc
floitsch: fixed in https://github.com/toitlang/toit/pull/1863

PR #1859, opened 2023-10-07
Support parsing mixins.

floitsch:


src/compiler/backend.cc
```diff
@@ -56,6 +56,7 @@ class BackendCollector : public ir::TraversingVisitor {
         case ir::Class::MONITOR:
           class_usage_counts_[klass]++;
           break;
+        case ir::Class::MIXIN:
```
kasperl, line 59: It isn't entirely clear to me why you count this as an interface here. Maybe add a comment?

kasperl (approved):
Looks good to me.

src/compiler/backend.cc
floitsch: done.

PR #1863, opened 2023-10-07
Implement mixins for the LSP.

floitsch:


src/compiler/lsp/protocol_summary.cc
```diff
@@ -408,16 +408,18 @@ void Writer::print_class(ir::Class* klass) {
       kind = "interface";
       break;
     case ir::Class::MIXIN:
-      kind = "class";
+      kind = "mixin";
       break;
   }
   this->printf("%s\n", kind);
+  this->printf("%s\n", klass->is_abstract() ? "abstract" : "not-abstract");
```
kasperl, line 415:
```suggestion
  this->printf("%s\n", klass->is_abstract() ? "abstract" : "non-abstract");
```

kasperl (approved):
Looks good to me.

src/compiler/lsp/protocol_summary.cc
floitsch: done.

PR #1864, opened 2023-10-07
Refactor sorting of classes.

floitsch:
Also sort mixins so that their "parents" are always before them.


kasperl (approved):
Looks good to me.

PR #1869, opened 2023-10-07
Update queryables to include mixins.

floitsch:
Queryables are used to check whether a method has a specific method.

### DIFF
--- a/src/compiler/queryable_class.cc
+++ b/src/compiler/queryable_class.cc
@@ -66,47 +66,64 @@ UnorderedMap<Class*, QueryableClass> build_queryables_from_resolution_shapes(Pro
 
   UnorderedMap<Class*, QueryableClass> result;
 
-  for (auto klass : program->classes()) {
-    QueryableClass::SelectorMap methods;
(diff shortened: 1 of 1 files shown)


src/compiler/queryable_class.cc
```diff
@@ -66,47 +66,64 @@ UnorderedMap<Class*, QueryableClass> build_queryables_from_resolution_shapes(Pro
...
-      auto method_shape = method->resolution_shape();
-      auto plain_shape = method_shape.to_plain_shape();
-
-      if (!method_shape.has_optional_parameters()) {
-        // No need to see which versions of the method-shape are needed.
-        Selector<PlainShape> selector(name, plain_shape);
-        methods[selector] = method;
-        continue;
+  for (int i = 0; i < 2; i++) {
+    // We run in two phases:
```
kasperl, line 70: Maybe explain why briefly?

kasperl (approved):
Looks good to me.

src/compiler/queryable_class.cc
floitsch: done.

PR #1871, opened 2023-10-07
Add an additional abstract test.

floitsch:


tests/abstract4-test.toit
```diff
@@ -0,0 +1,29 @@
...
+
+class C extends B:
+  foo x=null:
+    return super
+
+class D:
+  foo:
+    return 33
+
+confuse x: return x
```
kasperl, line 21: I think we have a better way of confusing in `tests/toitp/confuse.toit`. Maybe we should use that in more places?

tests/abstract4-test.toit
```diff
@@ -0,0 +1,29 @@
...
+
+main:
+  c := confuse C
+  expect_equals 1 c.foo
+  expect_equals 1 (c.foo 1)
+
+  d := confuse D
+  expect_equals 33 d.foo
```
kasperl, line 29: What does this test? There is nothing abstract about foo in this case?

kasperl (approved):
Looks good to me.

tests/abstract4-test.toit
floitsch: I think we should eventually go through all `confuse` methods and make it use something similar.
I'm not too keen on using something from the `toitp` directory.
In another PR...

tests/abstract4-test.toit
floitsch: Made `D` extend `C`.
done.

PR #1872, opened 2023-10-07
Change abstract method semantics.

floitsch:
If a superclass has the method implemented then a method isn't actually abstract.


tests/negative/abstract6-test.toit
```diff
@@ -11,5 +11,13 @@ abstract class B extends A:
 
 class C extends B:
 
+mixin M1:
+  foo:
+
+abstract mixin M2 extends M1:
+  abstract foo
+
+
```
kasperl, line 20: Drop one newline.

tests/abstract5-test.toit
```diff
@@ -0,0 +1,30 @@
...
+abstract class D:
+  abstract foo x=null
+  foo: return 1
+
+class E extends D:
+  foo x: return 2
+
+main:
+  c := C
+  expect_equals 1 c.foo
```
kasperl, line 25:
```suggestion
  expect-equals 1 c.foo
```

kasperl (approved):
Looks good to me.

tests/abstract5-test.toit
floitsch: I will migrate all code in a follow-up PR.

tests/negative/abstract6-test.toit
floitsch: done.

PR #1874, opened 2023-10-07
Support mixin interfaces.

floitsch:


tests/interface7-test.toit
```diff
@@ -0,0 +1,27 @@
...
+class A extends Object with M4:
+
+expect-I1 o/I1:
+expect-I2 o/I2:
+expect-I3 o/I3:
+expect-I4 o/I4:
+
+main:
+  a := A
+  expect-I1 a
```
kasperl, line 24: Maybe also check with `is`?

kasperl (approved):
Looks good to me.

PR #1875, opened 2023-10-07
Add mixins to the 'this' scope.

floitsch:


tests/mixin-this-scope-test.toit
```diff
@@ -0,0 +1,38 @@
...
+
+class ClassB extends Object with MixD:
+  c_method: return 42
+
+  check-this-calls:
+    expect-equals 41 a-method
+    expect-equals 42 c-method
+    expect-equals 499 e-method
+
+main:
```
kasperl, line 38: wat?

Is this supposed to test anything?

kasperl (approved):
Looks good to me, but the test seems like a noop.

tests/mixin-this-scope-test.toit
floitsch: At this point the mixin methods aren't implemented yet.
The only thing we want is that there aren't any health-errors. In other words, the static type-checker should be fine with the code.

We will add calls to the methods in a later PR:
https://github.com/toitlang/toit/pull/1883/files#diff-03d6f58e61067730c11924ab650cd88380dfc379a6d4fc690c77236bf103bf5f

PR #1876, opened 2023-10-07
Fix static typecheck for mixins.

floitsch:


tests/mixin-super-test.toit
```diff
@@ -0,0 +1,26 @@
...
+
+class ClassA extends Object with MixC:
+  a-method: return super
+  b-method: return super
+  c-method: return super
+
+main:
+  a := ClassA
+  // The static type check doesn't complain, but we still have a dynamic error.
+  // expect-equals 41 a.a-method
```
kasperl, line 24: Check the dynamic errors? At least don't leave the code in comments here.

kasperl (approved):
Looks good to me.

tests/mixin-super-test.toit
floitsch: The problem is that the function doesn't even throw. It just returns a bad result. (`this`).
I prefer not to check it and fix it when we implement the methods.

PR #1877, opened 2023-10-07
Allow 'is Mixin' checks.

floitsch:
Also use flattened interfaces in tree grower.


src/compiler/mixin.cc
```diff
@@ -77,7 +76,19 @@ static void add_stubs(ir::Class* klass) {
                                               forward_arguments,
                                               range);
 
-      auto stub = _new ir::MixinStub(method_name, klass, shape, method->range());
+      ir::MethodInstance* stub;
+      if (method->is_IsInterfaceOrMixinStub()) {
+        // We copy over the method (used to know if a class is an interface or mixin).
```
kasperl, line 81:
```suggestion
        // We copy over the method (used to determine if a class is an interface or mixin).
```

src/compiler/tree.cc
```diff
@@ -484,11 +484,25 @@ class Fixup : public ReplacingVisitor {
...
+    valid_check_targets_.insert_all(grown_classes);
 
     for (auto klass : grown_classes) {
       for (auto interface : klass->interfaces()) {
-        grown_classes_and_interfaces_.insert(interface);
+        valid_check_targets_.insert(interface);
+      }
+      for (auto current = klass; current != null; current = current->super()) {
+        if (current != klass && valid_check_targets_.contains(current)) {
+          // No need to duplicate work. The current class will (or was already) be
```
kasperl, line 495:
```suggestion
          // No need to duplicate work. The current class will be (or was already)
```

tests/mixin-is-test.toit
```diff
@@ -0,0 +1,55 @@
...
+
+mixin MixC:
+  c-method: return 43
+
+mixin MixD extends MixB with MixC:
+  d-method: return 44
+
+mixin MixE:
+  e-method: return 45
+
```
kasperl, line 21: Drop one newline.

tests/mixin-is-test.toit
```diff
@@ -0,0 +1,55 @@
...
+
+class ClassA extends Object with MixD MixE:
+
+class ClassB extends Object with MixD:
+
+confuse x -> any: return x
```
kasperl, line 27: You may want a better confuse function here. This is a bit too easy for the type propagator.

kasperl (approved):
Looks good to me.

src/compiler/mixin.cc
floitsch: done.

src/compiler/tree.cc
floitsch: done.

tests/mixin-is-test.toit
floitsch: done.

tests/mixin-is-test.toit
floitsch: I think we should do that in a follow-up PR. For all `confuse` methods.

PR #1878, opened 2023-10-07
Simplify is/as optimization.

floitsch:
The classes already have a flattened list of interfaces. No need to run through supers/interfaces.


kasperl (approved):
Looks good to me.

PR #1879, opened 2023-10-07
Implement mixin fields.

floitsch:


src/compiler/mixin.cc
```diff
@@ -56,28 +83,53 @@ static void add_stubs(ir::Class* klass) {
...
-                                              shape.to_equivalent_call_shape(),
-                                              forward_arguments,
-                                              range);
-
+      auto stub_parameters = duplicate_parameters(original_parameters);
       ir::MethodInstance* stub;
-      if (method->is_IsInterfaceOrMixinStub()) {
+
+      if (method->is_FieldStub()) {
+        // Copy of what's happening in `resolver_method`.
```
kasperl, line 90: It isn't very clear why the code cannot be shared better. Maybe refactor this or improve the comment to explain why this is a copy?

src/compiler/mixin.cc
```diff
@@ -56,28 +83,53 @@ static void add_stubs(ir::Class* klass) {
...
+        new_field_stub->set_plain_shape(shape);
+        auto this_ref = _new ir::ReferenceLocal(stub_parameters[0], 0, range);
+        ir::Sequence* body;
+        if (field_stub->is_getter()) {
+          ASSERT(stub_parameters.length() == 1);
+          auto load = _new ir::FieldLoad(this_ref, new_field, range);
+          auto ret = _new ir::Return(load, false, range);
+          body = _new ir::Sequence(ListBuilder<ir::Expression*>::build(ret), range);
+        } else {
+          ASSERT(stub_parameters.length() == 2);
```
kasperl, line 109: No check for `field_stub->field()->is_final()`?

tests/mixin-field-test.toit
```diff
@@ -0,0 +1,125 @@
+// Copyright (C) 2023 Toitware ApS.
+// Use of this source code is governed by a Zero-Clause BSD license that can
+// be found in the tests/LICENSE file.
+
+import expect show *
+
+mixin MixA:
+  field-A/int? := null
+  field-A2 := null
```
kasperl, line 9: No tests with final fields? Maybe we will get them with the constructor change.

kasperl (approved):
Looks good to me.

src/compiler/mixin.cc
floitsch: It's not completely clear-cut.
It would be possible to share the code, but I don't really want to move the code out of the resolver_method file. Almost all of the IR generation is happening there.
On the other hand I don't really want to call from here into the resolver_method either.
Also, they are just slightly different (like where the `this` comes from, ...).
Finally, the resolver_method has some convenience functions that don't exist here.

I tried once to create a shared file and decided that it wasn't worth it. But I could be convinced otherwise.

src/compiler/mixin.cc
floitsch: Good catch.
I'm now forwarding the setter if the field is final.

tests/mixin-field-test.toit
floitsch: also added a test for the type-check of a field.
done.

PR #1880, opened 2023-10-07
Implement correct constructor chain for mixins.

floitsch:


src/compiler/mixin.cc
```diff
@@ -23,6 +23,171 @@
...
+    auto result = ir::ReplacingVisitor::visit_Super(node);
+    in_super_ = false;
+    has_seen_super_ = true;
+    return result;
+  }
+
+  ir::Node* visit_CallStatic(ir::CallStatic* node) override {
+    if (!in_super_) return ir::ReplacingVisitor::visit_CallStatic(node);
+    auto method = node->target()->target();
+    if (method->is_constructor() && method->holder() == super_) {
```
kasperl, line 61: I guess a static call here that isn't to the super constructor is part of an argument? Maybe explain this.

src/compiler/mixin.cc
```diff
@@ -178,11 +345,220 @@ static void apply_mixins(ir::Class* klass) {
...
+    std::vector<ir::Parameter*> this_params(mixins_.length());
+
+    ir::Expression* super_expression = original_super_expression;
+    for (int i = 0; i < mixins_.length(); i++) {
+      // Parameter index 0 is reserved for the implicit block parameter.
+      int parameter_index = 1;
+      auto this_param = _new ir::Parameter(Symbols::this_,
+                                          ir::Type::any(),
+                                          is_block=false,
+                                          parameter_index,
```
kasperl, line 414: Why don't you use the same style for this as for `is_block`?

src/compiler/mixin.cc
```diff
@@ -178,11 +345,220 @@ static void apply_mixins(ir::Class* klass) {
...
+      auto constructor = mixin->constructors()[0];
+      auto constructor_ref = _new ir::ReferenceMethod(constructor, range);
+      auto arguments = ListBuilder<ir::Expression*>::allocate(2);
+      auto outer_this = i == 0
+          ? outermost_this_ref
+          : _new ir::ReferenceLocal(this_params[i - 1], 0, range);
+      arguments[0] = outer_this;
+      arguments[1] = _new ir::ReferenceBlock(block, 0, range);
+      auto call = _new ir::CallStatic(constructor_ref,
+                                             constructor->plain_shape().to_equivalent_call_shape(),
```
kasperl, line 479: Weird indentation.

src/compiler/mixin.cc
```diff
@@ -23,6 +23,171 @@
...
+      node->replace_body(_new ir::Sequence(new_body.build(), node->range()));
+    }
+    ASSERT(node->parameters().length() == 1 && node->parameters()[0]->index() == 0);
+    bool is_block, default_value;
+    int block_index = 1;
+    this_ = node->parameters()[0];
+    next_super_ = _new ir::Parameter(Symbol::synthetic("<next-super>"),
+                                     ir::Type::any(),
+                                     is_block=true,
+                                     block_index,
```
kasperl, line 117: Why don't you use the same style for this as for `is_block`? (assign as part of the call)

src/compiler/mixin.cc
```diff
@@ -23,6 +23,171 @@
...
+  }
+
+  ir::Node* visit_Code(ir::Code* node) override {
+    if (node->is_block()) block_depth_++;
+    auto result = SuperCallVisitor::visit_Code(node);
+    if (node->is_block()) block_depth_--;
+    return result;
+  }
+
+  ir::Node* visit_static_super_call(ir::CallStatic* node, const Source::Range& range) override {
```
kasperl, line 136: I think a comment that explains the translation here would help.

kasperl (approved):
Looks good to me.

PR #1886, opened 2023-10-08
Fix failure to recognize an allocation failure.

erikcorry:
When we were converting a COW byte array to a writable byte array in a primitive we didn't notice a heap full error correctly.


kasperl (approved):
Looks good to me.

PR #1887, opened 2023-10-08
Tighten up limits when drawing bitmaps/bytemaps.

erikcorry:
When drawing bitmaps (including fonts) and bytemaps we have to decode the source from the top, but we don't have to continue at the bottom if we went over the edge of the canvas.  Our calculation of this was incorrect so sometimes we continued to decode the source when it was not needed.  We didn't draw outside the canvas (there's a separate check for that), but we might read beyond the limits of the source.  This could also be an efficiency issue.


kasperl (approved):
Looks good to me.

PR #1891, opened 2023-10-10
Use tail calls for mixin forwarding stubs

kasperl:


floitsch (approved):
LGTM.
I thought I had done that...

PR #1895, opened 2023-10-10
Use ';' to separate list items in env on Windows

floitsch:
Windows uses `;` to separate items in list-values of environment variables.

### DIFF
--- a/src/compiler/filesystem_local.cc
+++ b/src/compiler/filesystem_local.cc
@@ -95,7 +95,11 @@ List<const char*> FilesystemLocal::package_cache_paths() {
     has_computed_cache_paths_ = true;
     char* cache_paths = getenv("TOIT_PACKAGE_CACHE_PATHS");
     if (cache_paths != null) {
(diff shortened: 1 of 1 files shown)


src/compiler/filesystem_local.cc
```diff
@@ -95,7 +95,11 @@ List<const char*> FilesystemLocal::package_cache_paths() {
     has_computed_cache_paths_ = true;
     char* cache_paths = getenv("TOIT_PACKAGE_CACHE_PATHS");
     if (cache_paths != null) {
-      package_cache_paths_ = string_split(strdup(cache_paths), ":");
+      const char* separator = ":";
+      if (strcmp(OS::get_platform(), "Windows") == 0) {
```
kasperl, line 99: I'd move this out and add a `is_windows` local that can be used here and in line 105.

kasperl (approved):
Looks good to me.

src/compiler/filesystem_local.cc
floitsch: done.

PR #1899, opened 2023-10-10
Mark more core functions as SDK protected.

floitsch:
Fixes https://github.com/toitware/toit/issues/4393


lib/core/objects.toit
```diff
@@ -215,7 +215,7 @@ The $id-or-tasks_ can be:
 class LazyInitializer_:
   id-or-tasks_ / any := ?
 
-  constructor .id-or-tasks_:
+  constructor.__ .id-or-tasks_:
```
kasperl, line 218: Can't you make the class protected instead?

kasperl (approved):
Looks good to me.

PR #1900, opened 2023-10-10
Fix off-by-one in the transparency compositing primitive.

erikcorry:


src/primitive_bitmap.cc
```diff
@@ -1145,12 +1145,16 @@ PRIMITIVE(composit) {
...
+            continue;
+          }
           mix = dest_address[i];
         } else {
-          mix = (frame_pixels[i] * frame_factor + dest_address[i] * (255 - frame_factor)) >> 8;
+          mix = (frame_pixels[i] * frame_factor + dest_address[i] * (256 - frame_factor)) >> 8;
         }
         // Now mix shaded background with window area.
-        dest_address[i] = (painting_pixels[i] * painting_factor + mix * (255 - painting_factor)) >> 8;
+        dest_address[i] = (painting_pixels[i] * painting_factor + mix * (256 - painting_factor)) >> 8;
```
kasperl, line 1157: `painting_factor` can be zero here, but I guess that is fine.

kasperl (approved):
Looks good to me.

Is it easy to test?

PR #1902, opened 2023-10-11
Add test that checks we don't lose track of memory.

erikcorry:


tests/directory-accounting-test.toit
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) 2023 Toitware ApS.
+// Use of this source code is governed by a Zero-Clause BSD license that can
+// be found in the tests/LICENSE file.
+
+import host.directory
+import expect show *
+
+main:
+  stats := process-stats --gc=true
```
kasperl, line 9:
```suggestion
  stats := process-stats --gc
```

tests/directory-accounting-test.toit
```diff
@@ -0,0 +1,21 @@
...
+  base := stats[STATS-INDEX-ALLOCATED-MEMORY]
+  10.repeat:
+    stats2 := process-stats --gc=true
+    print "Object heap $stats2[STATS-INDEX-ALLOCATED-MEMORY] bytes on heap."
+    expect:
+      stats2[STATS-INDEX-ALLOCATED-MEMORY] - base <= 64 * 1024
```
kasperl, line 15:
```suggestion
    expect
        stats2[STATS-INDEX-ALLOCATED-MEMORY] - base <= 64 * 1024
```

tests/directory-accounting-test.toit
```diff
@@ -0,0 +1,21 @@
...
+    catch:
```
kasperl, line 16: It feels a bit iffy that this tests pass even if we completely mess up the code in this section. Maybe do that catch around the creation of `dir` only?

tests/directory-accounting-test.toit
```diff
@@ -0,0 +1,21 @@
...
+
+import host.directory
+import expect show *
+
+main:
+  stats := process-stats --gc=true
+  base := stats[STATS-INDEX-ALLOCATED-MEMORY]
+  10.repeat:
+    stats2 := process-stats --gc=true
+    print "Object heap $stats2[STATS-INDEX-ALLOCATED-MEMORY] bytes on heap."
```
kasperl, line 13: Maybe just drop this.

kasperl (approved):
Thanks!

PR #1904, opened 2023-10-11
Use 'github.event.release.tag_name' instead of github.ref

floitsch:
Works around https://github.com/actions/runner/issues/2788


kasperl (approved):
Looks good to me.

There are a couple of steps that aren't guarded with `repository_owner == 'toitlang'`. Is that intentional?

PR #1906, opened 2023-10-11
Add missing null check in primitive_zlib.cc

erikcorry:


kasperl (approved):
Looks good to me.

We mostly create proxies that have the dispose flag cleared (two cases where it is set here). Maybe we should double check that we're using this as intended?

PR #1912, opened 2023-10-16
Much more const correctness

erikcorry:


src/objects.h
```diff
@@ -216,26 +216,37 @@ enum TypeTag {
   PROMOTED_TRACK_TAG,
 };
 
+#define STANDARD_FUNCTIONS(CamelName, snake_name)           \
```
floitsch, line 219: I don't feel like this macro is worth it.

src/utils.h
```diff
@@ -425,6 +425,10 @@ class List {
   List(T* data, int length) : data_(data), length_(length) {
     ASSERT(length >= 0);
   }
+  template <typename U>
```
floitsch, line 428: Maybe try with simple: `List(List<const T> other) ...`.

floitsch (approved):
LGTM.

PR #1915, opened 2023-10-16
More io.Data uses.

floitsch:


lib/crypto/aes.toit
```diff
@@ -2,6 +2,8 @@
 // Use of this source code is governed by an MIT-style license that can be
 // found in the lib/LICENSE file.
 
+import ..io as io
```
kasperl, line 5: Shouldn't this just be `import io` or is this just easier to work with for now?

lib/net/modules/tcp.toit
```diff
@@ -197,7 +197,13 @@ tcp-listen_ socket-resource-group address port backlog:
   #primitive.tcp.listen
 
 tcp-write_ socket-resource-group descriptor data from to:
-  #primitive.tcp.write
+  #primitive.tcp.write: | error |
+    if error != "WRONG_BYTES_TYPE": throw error
```
kasperl, line 201: Add a comment to explain why this isn't using `io.primitive-redo-chunked-io-data_`.

tests/socket-test.toit
```diff
@@ -85,5 +111,5 @@ sleepy-reader port:
         expected = "Message for sleepy reader $index"
       expect str.size < expected.size
   print "End at $index"
-  expect-equals index 100
+  expect-equals index iterations
```
kasperl, line 114:
```suggestion
  expect-equals iterations index
```

kasperl (approved):
Looks good to me.

lib/net/modules/tcp.toit
floitsch: done.

tests/socket-test.toit
floitsch: done.

PR #1918, opened 2023-10-17
Fix interface/mixin checks.

floitsch:
We weren't taking mixins into account when checking whether an interface was completely implemented.


kasperl (approved):
Looks good!

PR #1921, opened 2023-10-18
Add 'in' and 'out' mixins

floitsch:
Objects that support writing should provide an `out` whereas objects that support reading from them should provide an `int`. A mixin makes it easy to support them.

It is also possible to just extend the Reader/Writer classes. For some classes that might be OK. The `io.Buffer` class is one of them. However, most classes should use the `in` and `out` mixins. This also has the advantage that we don't need to make `Reader` and `Writer` an interface.


lib/io.toit
```diff
@@ -734,6 +734,117 @@ class ByteArrayReader_ extends Reader:
   close_ -> none:
     data_ = null
 
+
```
kasperl, line 737: Too many newlines.

lib/io.toit
```diff
@@ -734,6 +734,117 @@ class ByteArrayReader_ extends Reader:
...
+    return strategy_.try-write_ data from to
+
+  close_ -> none:
+    strategy_.close-writer_
+
+abstract mixin OutMixin implements OutStrategy:
+  out_/Out_? := null
+
+  out -> Writer:
+    if not out_: out_ = Out_ this
```
kasperl, line 772: Clearly this is just me, but I'd avoid reading the field twice.

lib/io.toit
```diff
@@ -734,6 +734,117 @@ class ByteArrayReader_ extends Reader:
...
+    return strategy_.consume_
+
+  close_ -> none:
+    strategy_.close-reader_
+
+abstract mixin InMixin implements InStrategy:
+  in_/In_? := null
+
+  in -> Reader:
+    if not in_: in_ = In_ this
```
kasperl, line 827: Clearly this is just me, but I'd avoid reading the field twice.

lib/io.toit
```diff
@@ -734,6 +734,117 @@ class ByteArrayReader_ extends Reader:
...
+  /**
+  Closes this writer.
+
+  See $Writer.close_.
+  */
+  // This is a protected method. It should not be "private".
+  close-writer_ -> none
+
+class Out_ extends Writer:
+  strategy_/OutStrategy
```
kasperl, line 758: It isn't very clear to me why you need this extra level of indirection here. Maybe add a comment on the two strategy interfaces to explain why this is convenient?

kasperl (approved):
Looks good to me.

lib/io.toit

lib/io.toit

lib/io.toit

lib/io.toit
floitsch: You are right. There is no need for this interface.
Removed it, and I am using `OutMixin` instead.

PR #1924, opened 2023-10-18
Split Reader/Writer and CloseableReader/CloseableWriter

floitsch:
It didn't make sense that all readers/writers have a `close` method.


lib/io.toit
```diff
@@ -123,6 +111,19 @@ abstract class Writer:
   // This is a protected method. It should not be "private".
   abstract try-write_ data/Data from/int to/int -> int
 
+
```
kasperl, line 114: Too many newlines.

lib/tls/socket.toit
```diff
@@ -13,7 +13,7 @@ import .certificate
 /**
 TLS Socket implementation that can upgrade a TCP socket to a secure TLS socket.
```
kasperl, line 14:
```suggestion
TLS socket implementation that can upgrade a TCP socket to a secure TLS socket.
```

kasperl (approved):
Looks good to me.

lib/tls/socket.toit
floitsch: done.

lib/io.toit

floitsch (commented):
Also removed the strategies (as suggested in the earlier PR).

PR #1925, opened 2023-10-18
Cache the EndianReader/Writer/Buffer.

floitsch:


lib/io.toit
```diff
@@ -1587,17 +1629,17 @@ class EndianBuffer extends EndianWriter:
 
   /** Writes the given int64 $value to this buffer at the given index $at. */
   put-int64 --at/int value/int:
-    endian_.put-int64 cached-byte-array_ at value
+    byte-order_.put-int64 cached-byte-array_ at value
     buffer_.put --at=at cached-byte-array_ 0 8
```
kasperl, line 1633: This is kinda of a weird pattern. Why don't you just write it into the buffer?

lib/io.toit
```diff
@@ -1545,7 +1587,7 @@ class EndianBuffer extends EndianWriter:
   Writes the given signed int16 $value to this buffer at the given index $at.
   */
   put-int16 --at/int value/int:
-    endian_.put-int16 cached-byte-array_ at value
+    byte-order_.put-int16 cached-byte-array_ at value
```
kasperl, line 1590: Isn't this completely broken? Shouldn't you always write to the cached array at index 0?

lib/io.toit
```diff
@@ -1199,22 +1255,15 @@ class Buffer extends CloseableWriter:
...
-    writer.write-int32 0x12345678
-    result := buffer.bytes
-    ...
+    // The buffer version also supports 'put' operations
+    buffer.little-endian.put-int32 --at=0 0x11223344
   ```
   */
   little-endian -> EndianBuffer:
-    return EndianBuffer --buffer=this --byte-order=binary.LITTLE_ENDIAN
+    if not endian_ or endian_.byte-order_ != binary.LITTLE_ENDIAN:
```
kasperl, line 1264: I wouldn't read the field that many times.

lib/io.toit
```diff
@@ -732,7 +784,9 @@ abstract class Reader implements old-reader.Reader:
   ```
   */
   little-endian -> EndianReader:
-    return EndianReader --reader=this --byte-order=binary.LITTLE_ENDIAN
+    if not endian_ or endian_.byte-order_ != binary.LITTLE_ENDIAN:
```
kasperl, line 787: I wouldn't read the field that many times.

lib/io.toit
```diff
@@ -754,7 +808,9 @@ abstract class Reader implements old-reader.Reader:
   ```
   */
   big-endian -> EndianReader:
-    return EndianReader --reader=this --byte-order=binary.BIG_ENDIAN
+    if not endian_ or endian_.byte-order_ != binary.BIG_ENDIAN:
```
kasperl, line 811: I wouldn't read the field that many times.

lib/io.toit
```diff
@@ -1233,31 +1282,24 @@ class Buffer extends CloseableWriter:
...
-    writer.write-int32 0x12345678
-    result := buffer.bytes
-    ...
+    // The buffer version also supports 'put' operations
+    buffer.big-endian.put-int32 --at=0 0x11223344
   ```
   */
   big-endian -> EndianBuffer:
-    return EndianBuffer --buffer=this --byte-order=binary.BIG_ENDIAN
+    if not endian_ or endian_.byte-order_ != binary.BIG_ENDIAN:
```
kasperl, line 1291: I wouldn't read the field that many times.

kasperl (approved):
Looks good to me.

lib/io.toit

lib/io.toit

lib/io.toit

lib/io.toit

lib/io.toit
floitsch: Good catch.
Fixed and added a test.

lib/io.toit
floitsch: The `buffer_` (of type `Buffer`) has a `buffer_` (of type `ByteArray`).
We could inline the `buffer_.put` call. After all, it's just:
```
  // In the `Buffer` class.
  put --at/int data/Data from/int=0 to/int=data.byte-size:
    if not 0 <= at <= at + (to - from) <= offset_: throw "INVALID_ARGUMENT"
    buffer_.replace at data from to
```

Not sure it's worth it.

PR #1926, opened 2023-10-18
Migrate ubjson.

floitsch:


lib/io.toit
```diff
@@ -75,6 +76,14 @@ abstract class Writer:
     assert: is-closed_
     throw "WRITER_CLOSED"
 
+  /**
+  Writes a single byte.
+  */
+  write-byte b/int -> none:
```
kasperl, line 82:
```suggestion
  write-byte value/int -> none:
```

lib/io.toit
```diff
@@ -75,6 +76,14 @@ abstract class Writer:
...
+    if not byte-cache_: byte-cache_ = ByteArray 1
```
kasperl, line 83: I wouldn't read the field three times here.

kasperl (approved):
Looks good to me.

PR #1930, opened 2023-10-23
Use a bit per object to indicate there is a finalizer.

erikcorry:
Use a bit per object to indicate there is a finalizer.

This means we don't have to iterate the finalizer lists when
adding and removing finalizers.

Also removes some finalizer code from the Process object,
limiting it to the heap classes.

Coming next, but not in this change:
* Remove the unneccesary finalizer-removing code from the finalizers. 
* Perhaps handling new-space finalizers on a separate list from old-space finalisers.
* New kinds of finalizers/weak references.


erikcorry:
Subsumed by https://github.com/toitlang/toit/pull/1937

PR #1931, opened 2023-10-23
Catch some edge cases when comparing ints with floats

erikcorry:


src/interpreter_core.cc
```diff
@@ -212,7 +216,42 @@ int Interpreter::compare_numbers(Object* lhs, Object* rhs) {
...
+      // Things get a bit strange near the limits of the 64 bit signed range.
+      // Decimal             Nearest IEEE value  Hex of IEEE value     JS prints as
+      // 9223372036854775295 9223372036854774784 0x7fff_ffff_ffff_f800 9223372036854775000
+      // 9223372036854775296 9223372036854775808 0x8000_0000_0000_0000 9223372036854776000
+      // 9223372036854776832 9223372036854775808 0x8000_0000_0000_0000 9223372036854776000
+      // 9223372036854776833 9223372036854777856 0x8000_0000_0000_0800 9223372036854778000
+      int equal = COMPARE_RESULT_ZERO | COMPARE_FLAG_LESS_EQUAL | COMPARE_FLAG_EQUAL | COMPARE_FLAG_GREATER_EQUAL | COMPARE_FLAG_LESS_FOR_MIN;
+      int64 SHORTCUT_LIMIT = 0x20000000000000LL;
+      if (rhs_is_int) {
+        if (-SHORTCUT_LIMIT <= rhs_int && rhs_int <= SHORTCUT_LIMIT) return equal; // Optimization.
```
floitsch, line 234: Feels like the lhs and rhs are so similar, that it should be possible to use the same code for it.

floitsch (approved):
LGTM, but the slow-path (`less-from-int`... ) should do the same thing.

PR #1937, opened 2023-10-25
Weak maps

erikcorry:
Adds weak maps, which have their values turned to null
if the values are not otherwise reachable in a mark sweep
GC.

These can be used for caching difficult calculations without
causing out-of-memory errors.  At some point we may want
to restrict the zappng of values to mark-sweep-compact or
malloc-failure-caused GCs.

On the C++ side this is implemented by splitting the Toit finalizer
class into two classes with a common super-class for finalizers
with a callback. Only the map is registered, not individual entries.

A single bit on an object's header indicates that it is either weak or finalizable.
No class can be both, so maps cannot have finalizers.


src/heap_roots.h
```diff
@@ -31,54 +31,71 @@ typedef LinkedFifo<FinalizerNode> FinalizerNodeFifo;
...
  public:
-  ToitFinalizerNode(HeapObject* key, Object* lambda)
-    : key_(key), lambda_(lambda) {}
+  CallableFinalizerNode(HeapObject* key, Object* lambda, ObjectHeap* heap)
+    : FinalizerNode(key, heap), lambda_(lambda) {}
 
-  HeapObject* key() { return key_; }
-  void set_key(HeapObject* value) { key_ = value; }
-  virtual bool has_key(HeapObject* value);
   Object* lambda() { return lambda_; }
```
kasperl, line 55: const

src/objects.h
```diff
@@ -1354,6 +1384,10 @@ class Instance : public HeapObject {
     _at_put(_offset_from(index), value);
   }
 
+  INLINE Object** root_at(int index) {
```
kasperl, line 1387: const

src/heap_roots.cc
```diff
@@ -17,58 +17,179 @@
...
+  cb->do_root(reinterpret_cast<Object**>(&key_));
 }
 
-bool VmFinalizerNode::handle_not_alive(RootCallback* ss, ObjectHeap* heap) {
-  free_external_memory(heap->owner());
-  return true;  // Delete me now.
+bool VmFinalizerNode::weak_processing(bool in_closure_queue, RootCallback* cb, LivenessOracle* oracle) {
+  ASSERT(!in_closure_queue);
+  if (!oracle->has_active_finalizer(key_)) {
+    delete this;
```
kasperl, line 176: So here we don't free the external memory. Maybe add a comment about that?

src/heap.cc
```diff
@@ -333,87 +327,72 @@ void ObjectHeap::install_heap_limit() {
...
+    node = _new WeakMapFinalizerNode(key, lambda, this);
+  } else {
+    ASSERT(key->can_be_toit_finalized(program()));
+    ASSERT(!key->has_active_finalizer());
+    node = _new ToitFinalizerNode(key, lambda, this);
+  }
   if (node == null) return false;  // Allocation failed.
-  registered_toit_finalizers_.append(node);
+  // Add it at the head of the list in case there is an old finalizer lower
+  // down on the list.
```
kasperl, line 368:
```suggestion
  // Add it at the head of the list in case there is an old "removed" finalizer lower
  // down on the list, waiting to be unlinked at the next GC.
```

kasperl (approved):
Looks good to me.

src/heap_roots.h
erikcorry: No, we can modify the object via this call.

src/objects.h
erikcorry: No we can modify the object via this call.

PR #1938, opened 2023-10-25
Add support for watchdog timers.

floitsch:


src/primitive_esp32.cc
```diff
@@ -521,6 +529,51 @@ PRIMITIVE(memory_page_report) {
   return result;
 }
 
+PRIMITIVE(watchdog_init) {
+  ARGS(uint32, ms);
+
+  int watchdog = watchdog_timers.any();
+  if (watchdog == kInvalidWatchdogTimer) FAIL(OUT_OF_RANGE);
```
kasperl, line 536: ALREADY_IN_USE?

src/primitive_esp32.cc
```diff
@@ -521,6 +529,51 @@ PRIMITIVE(memory_page_report) {
...
+  if (err != ESP_OK) {
+    return Primitive::os_error(err, process);
+  }
+  return process->null_object();
+}
+
+PRIMITIVE(watchdog_deinit) {
+  esp_err_t err = esp_task_wdt_deinit();
+  if (err != ESP_OK) {
+    return Primitive::os_error(err, process);
```
kasperl, line 571: If the watchdog is initialized, it almost feels like we need to make this FATAL and force a crash.

src/primitive_esp32.cc
```diff
@@ -521,6 +529,51 @@ PRIMITIVE(memory_page_report) {
...
+    .idle_core_mask = 0,
+    .trigger_panic = true,
+  };
+  esp_err_t err = esp_task_wdt_init(&config);
+  if (err != ESP_OK) {
+    return Primitive::os_error(err, process);
+  }
+
+  SystemEventSource::instance()->run([&]() {
+    err = esp_task_wdt_add(null);  // Add the current thread.
```
kasperl, line 549: I feel 'current' is overloaded here.

```suggestion
    err = esp_task_wdt_add(null);  // Add the SystemEventSource thread.
```

kasperl (approved):
Looks good to me.

src/primitive_esp32.cc
floitsch: done.

src/primitive_esp32.cc
floitsch: done.

src/primitive_esp32.cc
floitsch: I'm not sure it's worth it: the only reason we can fail here is if a task is still registered, or if `deinit` was already called.
The bigger question (to me) is whether we should return the singleton even if `deinit` failed.

PR #1942, opened 2023-11-01
Fix `-Werror`.

floitsch:


src/compiler/diagnostic.h
```diff
@@ -180,8 +185,8 @@ class NullDiagnostics : public Diagnostics {
   void end_group() {}
 
  protected:
-  void emit(Severity severity, const char* format, va_list& arguments) {}
-  void emit(Severity severity, Source::Range range, const char* format, va_list& arguments) {}
+  bool emit(Severity severity, const char* format, va_list& arguments) { return true; }
```
kasperl, line 188: Maybe add a comment to explain why returning true here is better than returning false?

kasperl (approved):
Looks good to me.

src/compiler/diagnostic.h
floitsch: It doesn't really matter, but I agree "false" is better.

floitsch: Turns out, that we check whether we encountered ("printed") errors in asserts. So changing it back to "true".

PR #1945, opened 2023-11-03
Add feature to hold pins during deep sleep.

floitsch:


tests/esp32/README.md
```diff
@@ -22,16 +22,20 @@ On board 1 connect as follows:
...
 3. IO23 - IO22
 
+IO16 must stay unconnected.
+
 ## Running
 
 As of 2023-05-30.
 Run the tests individually. All test run on board 1, except for the
-uart, ultrasound, and wait_for tests. (See their respective files for more info.)
+uart, ultrasound, wait_for tests, and pin-hold tests. (See their respective files for more info.)
```
kasperl, line 38:
```suggestion
uart, ultrasound, wait-for tests, and pin-hold tests. (See their respective files for more info.)
```

src/compiler/propagation/type_primitive_esp32.cc
```diff
@@ -40,6 +40,10 @@ TYPE_PRIMITIVE_ANY(memory_page_report)
 TYPE_PRIMITIVE_ANY(watchdog_init)
 TYPE_PRIMITIVE_ANY(watchdog_reset)
 TYPE_PRIMITIVE_ANY(watchdog_deinit)
+TYPE_PRIMITIVE_ANY(pin_hold_enable)
```
kasperl, line 43: These really only return null.

kasperl (approved):
Looks good to me.

src/compiler/propagation/type_primitive_esp32.cc
floitsch: Agreed.
There wasn't any 'NULL' variant in the file and I didn't look for one. I figured that nobody would use the result anyway.
Changed the watchdog and pin-hold functions to be typed as null.
done.

PR #1951, opened 2023-11-08
Remove deprecated

kasperl:


kasperl:
There is a small number of packages that still depend on the removed functionality:

- [ ] https://github.com/toitware/toit-gnss-location
- [ ] https://github.com/toitware/toit-heading
- [ ] https://github.com/toitware/toit-location
- [ ] https://github.com/toitware/toit-lsm303d
- [ ] https://github.com/kasperl/toit-qubitro

erikcorry:
Almost all the external projects that need updating use the lib/protocols files.  Perhaps we can keep those for now?

kasperl:
There are a number of packages that still depend on deprecated functionality. Maybe we should prioritize 

> Almost all the external projects that need updating use the lib/protocols files. Perhaps we can keep those for now?

Should be really quick to fix though?

floitsch (approved):
LGTM!

PR #1958, opened 2023-11-10
Update external health gold

kasperl:


kasperl:
Jaguar was updated to run on SDK v2.0.0-alpha.120. It now uses `import system`.

floitsch (approved):
LGTM.

PR #1959, opened 2023-11-10
Print diagnostics on stderr when running programs.

floitsch:


src/compiler/compiler.h
```diff
@@ -65,6 +65,8 @@ class Compiler {
     bool werror;
     /// Whether to show warnings in packages.
     bool show_package_warnings;
+    /// Whether to diagnostics on stderr or stdout.
```
kasperl, line 68:
```suggestion
    /// Whether to print diagnostics on stdout (true) or stderr (false).
```

kasperl (approved):
Looks good to me.

PR #1960, opened 2023-11-10
Fix rare case in zlib decompresser

erikcorry:
Review in ignore-whitespace mode.


lib/zlib.toit
```diff
@@ -704,29 +704,29 @@ class InflaterBackend implements Backend_:
...
-          add-symbol.call length-code
-        else:
-          last := 0
-          repeats := 3
-          if length-code == 16:
-            last = lengths_[counter_ - 1].bit-len
-            pending-bits_ = 2
-          else if length-code == 17:
-            pending-bits_ = 3
+        if counter_ != lengths_.size:
```
kasperl, line 707: Should this be an `else:` on the check in line 726?

tests/zlib-glitch-test.toit
```diff
@@ -0,0 +1,48 @@
...
+// be found in the tests/LICENSE file.
+
+import zlib
+
+main:
+  decoder := zlib.BufferingInflater
+
+  decoder.write ZLIB_STREAM
+
+  while read := decoder.reader.read:
```
kasperl, line 13: Maybe explain how this failed before the fix?

kasperl (approved):
Looks good to me.

lib/zlib.toit
erikcorry: No, add-symbol can update counter_.

PR #1961, opened 2023-11-12
encoding.yaml

mikkeldamsgaard:
Added decode and encode of yaml documents. 
The parser follows the spec closely. The spec defines a PEG-like grammar, so the implementation is top-down.

Refactored the json code a bit to enable reuse, on the encoding piece.

I want to apologise for the shear amount of source code in this PR, but handwritten parsers are a bit verbose.


lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1571 @@
...
+        if s-l-comments: empty-node_
+    return null
+
+  l-bare-document -> ValueNode_?:
+    with-rollback:
+      if res := s-l-plus-block-node -1 BLOCK-IN_: return res
+    return null
+
+  l-document-prefix -> bool:
+    c-byte-order-mark
```
floitsch, line 388: I would have expected a roll-back here, since we don't require a c-byte-order-mark.
So something like:
```
l-document-prefix -> bool:
  try-parse c-byte-order-mark
  repeat: l-comment
  return true
```

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1571 @@
...
+    repeat: l-comment
+    return true
```
floitsch, line 390: Why does this always return true?
Is it needed?

floitsch (commented):
First round of comments.

It's not completely clear to me yet, when `with-rollback` should be used and when not.
I expected `foo?` or `bar*` or `toto|titi` to rollback if the call didn't produce a result.
Instead it seems like `foo`/`bar`/`toto` do the rollback themselves.

lib/encoding/json.toit
```diff
@@ -101,128 +98,15 @@ decode-stream reader:
...
+  /** See $EncoderBase_.encode */
   encode obj/any converter/Lambda:
-    encode obj: converter.call obj this
+    return super obj converter
 
-  encode obj/any:
-    encode obj: throw "INVALID_JSON_OBJECT"
+  /** See $Buffer_.put-unquoted */
+  put-unquoted data -> none:
+    super data
```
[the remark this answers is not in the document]
mikkeldamsgaard: delete whet? the comment or the method? The toitdoc compiler was complaining of unresolved referencing, as these are being referenced from json.toit and yaml.toit.

lib/encoding/yaml/encoder.toit
```diff
@@ -0,0 +1,273 @@
...
+    return super obj converter
+
+  /** See $Buffer_.put-unquoted */
+  put-unquoted data -> none:
+    super data
+
+  put-value_ val/string:
+    if enclosed_in_map_: put-byte_ ' '
+    put-unquoted val
+
```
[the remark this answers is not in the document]
mikkeldamsgaard: I am not sure I follow. These methods are needed. Is it the comment that needs to be removed?

lib/encoding/yaml/encoder.toit
```diff
@@ -0,0 +1,273 @@
...
+    escaped := escape-string str
+
+    if enclosed_in_map_: put-byte_ ' '
+    if should_quote: put-byte_ '"'
+    put-unquoted (escape-string str)
+    if should_quote: put-byte_ '"'
+
+  encode-number_ number:
+    str := number is float ? number.stringify 2 : number.stringify
```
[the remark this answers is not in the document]
mikkeldamsgaard: This is also used for the json encoder and changing to arbitrary precession will be a potentail breaking change, which is not the intention of this PR.

lib/encoding/yaml/encoder.toit
```diff
@@ -0,0 +1,273 @@
...
+    old_enclosed_in_map := enclosed_in_map_
+    indent_ = new-indent
+    enclosed_in_map_ = is-map
+    encode value converter
+    indent_ = old_indent
+    enclosed_in_map_ = old_enclosed_in_map
+
+  encode-map_ map [converter]:
+    if map.size == 0: put-value_ "{}"
+    else:
```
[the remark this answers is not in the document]
mikkeldamsgaard: I usually do early returns too, this must have slipped me.

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1571 @@
...
+    if value is string:
+      if NULL.contains value: return null
+      if TRUE.contains value: return true
+      if FALSE.contains value: return false
+
+      catch:
+        return int.parse value
+
+      catch:
+        return float.parse value
```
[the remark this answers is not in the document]
mikkeldamsgaard: I dont see a version of float.parse that takes an --on-error argument. I added a TODO(florian) to fix this when there is such a method

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1571 @@
...
+
+  l-any-document -> ValueNode_?:
+    with-rollback:
+      if document := l-directive-document: return document
+      if document := l-explicit-document: return document
+      if document := l-bare-document: return document
+    return null
+
+  l-directive-document -> ValueNode_?:
+    with-rollback:
```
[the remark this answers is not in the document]
mikkeldamsgaard: Yes, becuase the repeat: could consume l-directive's that would not be rolled back if not followed by an l-explicit-document. One the otherhand, there was an error in the production, as the repeat return value was not checked.

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1571 @@
...
+      repeat --at-least-one: l-directive
+      if res := l-explicit-document:
+        return res
+
+    return null
+
+  l-explicit-document -> ValueNode_?:
+    with-rollback:
```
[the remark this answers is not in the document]
mikkeldamsgaard: Yes, otherwise c-directives-end might consume "---" even if not followed by legal content

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1571 @@
...
+    return offset_
+
+  rollback mark -> none:
+    offset_ = mark
+
+  at-mark mark -> bool:
+    return offset_ == mark
+
+  with-rollback [block] -> none:
+    mark := mark
```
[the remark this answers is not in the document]
mikkeldamsgaard: Kinda. I am open to suggestions for the naming of the variable. mark_ or start-mark or ?

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1571 @@
...
+  mark -> any:
+    return offset_
+
+  rollback mark -> none:
+    offset_ = mark
+
+  at-mark mark -> bool:
+    return offset_ == mark
+
+  with-rollback [block] -> none:
```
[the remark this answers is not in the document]
mikkeldamsgaard: The suggestion will not work, as it rolls back even with a non-local return in the block. Also, why would anyone want a return value of a failed forward parse?
The comment is good, so will add that for clarity. Also, renamed to try-parse

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1571 @@
...
+
+  l-explicit-document -> ValueNode_?:
+    with-rollback:
+      if c-directives-end:
+        if res := l-bare-document: return res
+        if s-l-comments: empty-node_
+    return null
+
+  l-bare-document -> ValueNode_?:
+    with-rollback:
```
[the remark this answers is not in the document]
mikkeldamsgaard: Correct.

lib/encoding/yaml/parser.toit
mikkeldamsgaard: Well, it is optional, and it self-rolls-back if not matching. There is an option method that is better to use for clarity, but it will impact performance. I added optional: c-byte-order-mark for clarity.

lib/encoding/yaml/parser.toit
mikkeldamsgaard: Yes, because it is used in repeat: and that only loops if the block return true.
Otherwise, the usages would be repeat: l-document-prefix; true, and that is not desirable.

mikkeldamsgaard:
> First round of comments.
> 
> It's not completely clear to me yet, when `with-rollback` should be used and when not. I expected `foo?` or `bar*` or `toto|titi` to rollback if the call didn't produce a result. Instead it seems like `foo`/`bar`/`toto` do the rollback themselves.

I tried to make the translation from the grammar to the "DSL" as mechanical as possible. So try-parse should be used when there is a concatenated productions, where the first n productions would consume input even if the n+1 production would fail, then a backtrace is needed to try the next or'ed rule.

Also, I tried to make all productions atomic, in the sense that they either fail and rollback or succeed and consume.

With that in mind, some special cases does not need the try-parse, for example A B* would not need it, but I often put it in to make it consistent with the mechanical translation. 

Another thing that could confuse is with optionals. There is an option method to use for them and the initial thought was to do something like this:

```toit
try-parse:
  return (optional: A) and B and C
```

Often not necessary, per the atomic rule, so A? B C, would become

```toit
try-parse:
  A
  return B and C
```

Unfortunately there is often a need to collect AST information, so the above becomes
```toit
try-parse:
  a := A
  if b := B:
    if c := C:
      return [a, b,c] // or some combination
```

This small DSL would benefit tremendously if := was allowed arbitrarily in logical expressions, a la:
```toit
try-parse:
  if a := (optional: A) and b := B and c := C: return [a, b, c]
```
but that is for another day.

lib/encoding/yaml/encoder.toit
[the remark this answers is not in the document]
floitsch: Not sure why I selected the second method, but the first one is just calling `super` with the same value. It should inherit that method automatically. no?

lib/encoding/yaml/encoder.toit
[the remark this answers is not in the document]
floitsch: Not asking to change. Just add a comment what YAML wants here.
Our double-conversion isn't correct anyway, but it would be nice to know what should happen here.

lib/encoding/yaml/parser.toit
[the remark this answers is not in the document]
floitsch: thanks.

lib/encoding/yaml/parser.toit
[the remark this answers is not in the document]
floitsch: Right. I think I get the need for many of the `with-rollback` now.

lib/encoding/json.toit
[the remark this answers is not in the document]
floitsch: I see. That looks like a bug in the compiler. It should find the superclass's method when referencing a method. Otherwise, a benign change can lead to toitdoc errors.
Maybe add a TODO that these methods should be removed once the compiler has been fixed.

lib/encoding/yaml/encoder.toit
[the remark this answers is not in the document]
floitsch: I'm guessing this is the same toitdoc issue. -> TODO to remove it once the compiler is fixed.

lib/encoding/yaml/parser.toit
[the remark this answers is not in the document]
floitsch: As discussed in chat: there was a confusion wrt who rolls back.
Either:
a.) a production either consumes or doesn't change anything. (Current approach), or
b.) if a subclause is optional, or there are alternatives, then the the caller needs to rollback.

My comments were made under the assumption of (b), but the implementation is using (a).

lib/encoding/yaml/parser.toit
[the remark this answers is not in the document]
floitsch: Maybe `current-mark` or `before-mark`?

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+        if rune := match-chars SIMPLE-ESCAPE-KEYS:
+          return SIMPLE-ESCAPES[rune]
+        if match-char 'x': if res := match-hex 2: return res
+        if match-char 'u':
+          if res := match-hex 4:
+            if 0xd800 <= res <= 0xdbff:
+              if part-2 := c-ns-esc-char:
+                 if not 0xdc00 <= part-2 <= 0xdfff:
+                  set-error "INVALID_SURROGATE_PAIR"
+                  return null
```
floitsch, line 1509:
```suggestion
                   set-error "INVALID_SURROGATE_PAIR"
                   return null
```

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+  c-ns-esc-char -> int?:
+    try-parse:
+      if match-char C-ESCAPE_:
+        if rune := match-chars SIMPLE-ESCAPE-KEYS:
+          return SIMPLE-ESCAPES[rune]
+        if match-char 'x': if res := match-hex 2: return res
+        if match-char 'u':
+          if res := match-hex 4:
+            if 0xd800 <= res <= 0xdbff:
+              if part-2 := c-ns-esc-char:
```
floitsch, line 1506: Where does this come from?
So YAML allows to write 16-bit code-points in a string?
The spec doesn't seem to mention it.
Add a comment?

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+                 if not 0xdc00 <= part-2 <= 0xdfff:
+                  set-error "INVALID_SURROGATE_PAIR"
+                  return null
+                 return 0x10000 + ((res & 0x3ff) << 10) | (part-2 & 0x3ff)
+            else:
+              return res
+        if match-char 'U': if res := match-hex 8: return res
+    return null
+
+  // Space
```
floitsch, line 1516:
```suggestion
  // Space.
```

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+    if c == BLOCK-KEY_ or c == FLOW-KEY_: return s-separate-in-line
+    return s-separate-lines n
+
+  s-separate-lines n:
+    try-parse:
+      if s-l-comments and s-flow-line-prefix n: return true
+    return s-separate-in-line
+
+  s-flow-line-prefix n:
+    return s-indent n and (optional: s-separate-in-line)
```
floitsch, line 1541: I don't think this is correct. It would return false if `s-separate-in-line` is false. However that's only optional.
```suggestion
    if s-indent n:
      optional: s-separate-in-line
      return true
    return null
```
(I'm returning `true`, but I'm not sure that's the correct return value.

lib/encoding/yaml/parser.toit
floitsch, line 1516: Can we add return types to the functions here?

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+
+    if auto-detect-m:
+      m := 0
+      while match-char S-SPACE_: m++
+      return m
+
+    return 0
+
+  s-indent-less-or-equals n:
+    while n-- > 0 and match-char S-SPACE_:
```
floitsch, line 1561:
```suggestion
    while n-- > 0 and match-char S-SPACE_: /* Do nothing. */
```

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+    return true
+
+  s-indent-less-than n:
+    return s-indent-less-or-equals n - 1
```
floitsch, line 1565:
```suggestion
    return s-indent-less-or-equals (n - 1)
```

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+
+  // Lexigraphical-like productions
```
floitsch, line 1567: ?
```suggestion
  // Lexicographical-like productions.
```

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+          match-buffer #[0xFF, 0xFE] or
+          match-char 0:
+        set-error "UNSUPPORTED_BYTE_ORDER"
+        return false
+      try-parse:
+        match-any
+        if match-char 0:
+          set-error "UNSUPPORTED_BYTE_ORDER"
+          return false
+    if can-read 3 and match-buffer #[0xEF, 0xBB, 0xBF]:
```
floitsch, line 1599:
```suggestion
    // Toit only supports UTF-8.
    if can-read 3 and match-buffer #[0xEF, 0xBB, 0xBF]:
```

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+    try-parse:
+      if rune := c-printable:
+        if not is-break rune and not rune == C-BYTE-ORDER-MARK_:
+          return rune
+    return false
+
+  nb-json -> int?:
+    try-parse:
+      rune := consume-rune
+      if rune and (rune == 0x09 or 0x20 <= rune and rune <= 0x10FFFF):
```
floitsch, line 1623:
```suggestion
      if rune and (rune == 0x09 or 0x20 <= rune <= 0x10FFFF):
```

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+        return rune
+    return null
+
+  ns-dec-digit -> int?:
+    return match-one: '0' <= it and it <= '9'
```
floitsch, line 1628:
```suggestion
    return match-one: '0' <= it <= '9'
```

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+    return null
+
+  is-break rune:
+    return rune == B-LINE-FEED_ or rune == B-CARRIAGE_RETURN_
+
+  c-special-printable ::= { S-TAB_, B-CARRIAGE_RETURN_, B-LINE-FEED_, 0x85}
+  c-printable -> int?:
+    try-parse:
+      if rune := consume-rune:
+        if 0x20 <= rune and rune<= 0x7E: return rune
```
floitsch, line 1683:
```suggestion
        if 0x20 <= rune <= 0x7E: return rune
```

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+        if c-special-printable.contains rune: return rune
+        if 0xA0 <= rune and rune <= 0xD7FF: return rune
+        if 0xE000 <= rune and rune <= 0xFFFD: return rune
+        if 0x010000 <= rune and rune <= 0x10FFFF: return rune
```
floitsch, line 1687:
```suggestion
        if 0xA0 <= rune <= 0xD7FF: return rune
        if 0xE000 <= rune <= 0xFFFD: return rune
        if 0x010000 <= rune <= 0x10FFFF: return rune
```

lib/encoding/yaml/yaml.toit
```diff
@@ -0,0 +1,118 @@
...
+  the $converter block can call the $YamlEncoder.encode, $YamlEncoder.put-list,
+  or $YamlEncoder.put_unquoted methods on the encoder.
+Utf-8 encoding is used for strings.
+*/
+encode obj [converter] -> ByteArray:
+  e := YamlEncoder
+  e.encode obj converter
+  return e.to-byte-array
+
+encode obj converter/Lambda -> ByteArray:
```
floitsch, line 35:
```suggestion
/**
Variant of $(encode obj [converter]).
Takes a lambda instead of a block as $converter.
*/
encode obj converter/Lambda -> ByteArray:
```

lib/encoding/yaml/yaml.toit
```diff
@@ -0,0 +1,118 @@
...
+/**
+Decodes the $bytes, which is a ByteArray in YAML format.
+The result is null or an instance of int, bool, float, string, List, or Map.
+  The list elements and map values will also be one of these types.
+If $as-stream, returns a list of included documents.
+*/
+decode --as-stream/bool=false [--on-error] bytes/ByteArray -> any:
+  p := Parser_ bytes
+  result := p.l-yaml-stream --on-error=on-error
+  if not result is ParseResult_: return result // This happens when on-error is invoked
```
floitsch, line 57:
```suggestion
  if not result is ParseResult_: return result // This happens when on-error is invoked.
```

lib/encoding/yaml/yaml.toit
```diff
@@ -0,0 +1,118 @@
...
+Utf-8 encoding is used for strings.
+*/
+encode obj -> ByteArray:
+  return encode obj: throw "INVALID_YAML_OBJECT"
+
+/**
+Decodes the $bytes, which is a ByteArray in YAML format.
+The result is null or an instance of int, bool, float, string, List, or Map.
+  The list elements and map values will also be one of these types.
+If $as-stream, returns a list of included documents.
```
floitsch, line 52:
```suggestion
If $as-stream is true, returns a list of included documents. Otherwise only returns
  the first document or null if none was included.
```

lib/encoding/yaml/yaml.toit
```diff
@@ -0,0 +1,118 @@
...
+*/
+decode --as-stream/bool=false [--on-error] bytes/ByteArray -> any:
```
floitsch, line 54: Maybe we should call `--as-stream` `--as-list` or `--all`.

I think, typewise, it would be better to have:
```
decode bytes -> any
decode --all bytes -> List
```
This would require
```
decode --all bytes -> List:
  if not all: throw "INVALID_ARGUMENT"
  ...
```
We use this pattern in different places.

lib/encoding/yaml/yaml.toit
```diff
@@ -0,0 +1,118 @@
...
+  or Map.
+Maps must have only string keys.  The elements of lists and the values of
+  maps can be any of the above supported types.
+The $converter block is passed an object to be serialized and an instance
+  of the $YamlEncoder class.  If it returns a non-null value, that value will
+  be serialized instead of the object that was passed in.  Alternatively,
+  the $converter block can call the YamlEncoder.encode, YamlEncoder.put-list,
+  or YamlEncoder.put_unquoted methods on the encoder.
+Utf-8 encoding is used for strings.
+Only a subset of YAML is supported.
```
floitsch, line 84: What is not supported?

lib/encoding/yaml/yaml.toit
```diff
@@ -0,0 +1,118 @@
...
+parse --as-stream/bool=false [--on-error] str/string -> any:
+  return decode --as-stream=as-stream --on-error=on-error str.to-byte-array
+
+/**
+Variation of (parse --as-stream --on-error bytes).
+
+Throws on parse error.
+*/
+parse --as-stream/bool=false str/string -> any:
+  return decode --as-stream=as-stream --on-error=(: throw it ) str.to-byte-array
```
floitsch, line 118:
```suggestion
  return decode --as-stream=as-stream --on-error=(: throw it ) str.to-byte-array

```

tests/yaml-test.toit
```diff
@@ -0,0 +1,537 @@
+// Copyright (C) 2018 Toitware ApS.
```
floitsch, line 1:
```suggestion
// Copyright (C) 2023 Toitware ApS.
```

tests/yaml-test.toit
```diff
@@ -0,0 +1,537 @@
...
+        yaml.parse """
+                    "
+                      foo\x20
+                    \x20
+                      \t bar
+
+                      baz "
+                   """
+
+  // Exmple 6.9
```
floitsch, line 432:
```suggestion
  // Example 6.9
```

tests/yaml-test.toit
```diff
@@ -0,0 +1,537 @@
...
+  expect-equals
+      "key: value\n"
+      yaml.stringify
+          yaml.parse """
+                      key:    # Comment
+                        value
+                     """
+
+  // Exmple 6.10
```
floitsch, line 441:
```suggestion
  // Example 6.10
```

tests/yaml-test.toit
```diff
@@ -0,0 +1,537 @@
...
+  expect-equals
+      null
+      yaml.parse """
+                    # Comment
+                     \x20
+
+                 """
+
+  // Exmple 6.11
```
floitsch, line 450:
```suggestion
  // Example 6.11
```

floitsch (approved):
LGTM!
Huge thanks.

lib/encoding/yaml/yaml.toit
mikkeldamsgaard: For now, I will keep --as-stream as it refers to the structure of the input, whereas --as-list refers to the return type. --all is to vague.
I will change to two methods for better typing.

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+  lookbehind rune-count [block] -> bool:
+    old-offset := offset_
+    try:
+      rune-count.repeat:
+        if offset_ == 0:
+          return false
+        offset_--
+        // Skip to the first byte of Unicode surrogates.
+        // Byte 2, 3, and 4 all start with bits 10.
+        // Since the input is a valid string we only need to check for offset_ > 0.
```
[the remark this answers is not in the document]
mikkeldamsgaard: Because if the first rune is a surrogate, you would index -1.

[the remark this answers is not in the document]
floitsch: But then it wouldn't be a valid string. No?
So the question is: do we check before-hand that the input is actually a valid string.
If we do, then the `offset_` check isn't necessary.
If we don't then we need to keep it (and I would add a comment explaining that this might be the case).

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+  ns-l-compact-mapping n/int -> ValueNode_?:
+    try-parse:
+      if first := ns-l-block-map-entry n:
+        rest := repeat:
+          entry := null
+          if s-indent n:
+            if tmp := ns-l-block-map-entry n:
+              entry = tmp
+          entry
+        return ValueNode_.map-from-collection (flatten-list_ [[first], rest])
```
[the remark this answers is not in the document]
mikkeldamsgaard: Actually, no. The map-from-collection expect(ed) a list of lists. As that is very confusing with the flatten-list, I changed that to be a list of alternating key1, value1, key2, value2, ... and made flatten-list recursive.

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+
+      if chomped-empty := l-chomped-empty n t:
+        if content: return flatten-list_ [content, chomped-empty]
+        else: return chomped-empty
+    return null
+
+  b-chomped-last t/int -> string?:
+    if t == STRIP_: if b-non-content: return ""
+    if b-as-line-feed: return "\n"
+    return null
```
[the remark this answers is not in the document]
mikkeldamsgaard: Good catch.

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+    return l-strip-empty n
+
+  l-keep-empty n/int -> List?:
+    empty-lines := repeat: l-empty n BLOCK-IN_
+    optional: l-trail-comments n
+    if empty-lines: return List empty-lines.size "\n"
+    return []
+
+  l-strip-empty n/int -> List?:
+    repeat: s-indent-less-or-equals n and (b-non-content or l-eof)
```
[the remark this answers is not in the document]
mikkeldamsgaard: Some rules depends on repeat to return null when not parsed. This would make it both --ignore-result and --as-null (or not setting --ignore-result) even though it should be ignored. Seems to be to many flags

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1689 @@
...
+        runes.add rune
+      rune
+    return string.from-runes runes
+
+  s-double-next-line n/int -> List?:
+    try-parse:
+      if breaks := s-double-break n:
+        try-parse:
+          if first-rune := ns-double-char:
+            line := nb-ns-double-in-line
```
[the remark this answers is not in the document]
mikkeldamsgaard: nb-ns-double-in-line returns string (not string?) so that should not be necessary.

lib/encoding/yaml/parser.toit
mikkeldamsgaard: YAML is an extension of JSON, so I assume that it should support 16-bit surrogates. The spec is very silent about this.

lib/encoding/yaml/parser.toit
mikkeldamsgaard: optional will return EMPTY_NODE_ (aka a truthful value) when not called with --or-null. So it is correct.

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1603 @@
...
+  // Data part.
+  s-l-plus-block-node n/int c/int -> ValueNode_?:
+    return s-l-plus-block-in-block n c or s-l-plus-flow-in-block n
+
+  s-l-plus-block-in-block n/int c/int -> ValueNode_?:
+    return s-l-plus-block-scalar n c or s-l-plus-block-collection n c
+
+  s-l-plus-flow-in-block n/int -> ValueNode_?:
+    if node := (try-parse: s-separate n + 1 FLOW-OUT_ and ns-flow-node n + 1 FLOW-OUT_):
+      if s-l-comments: return node
```
floitsch, line 639: I might be off, but here we parse the `node`, but if the `s-l-comments` is false, we don't roll back.

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1603 @@
...
+        if res := l-folded-content (spaces - indent) chomp:
+          return res.join ""
+    return null
+
+  c-b-block-header -> List?:
+    try-parse:
+      indent-char := match-range '1' '9'
+      chomp-char := match-chars { '-', '+' }
+      if not indent-char: indent-char = match-range '1' '9'
+      // Even though it might seem from the spec that indent ise not optional. This is an error, see
```
floitsch, line 1099:
```suggestion
      // Even though it might seem from the spec that indent is not optional. This is an error, see
```

floitsch (approved):
Still LGTM.

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1603 @@
...
+
+Naming conventions:
+- `e-`: A production matching no characters.
+- `c-`: A production starting and ending with a special character.
+- `b-`: A production matching a single line break.
+- `nb-`: A production starting and ending with a non-break character.
+- `s-`: A production starting and ending with a white-space character.
+- `ns-`: A production starting and ending with a non-space character.
+- `l-`: A production matching complete line(s).
+- `X-Y-`: A production starting with an `X-~ character and ending with a `Y-` character, where `X-` and `Y-` are any of the above prefixes.
```
floitsch, line 371:
```suggestion
- `X-Y-`: A production starting with an `X-` character and ending with a `Y-` character, where `X-` and `Y-` are any of the above prefixes.
```

lib/encoding/yaml/parser.toit
```diff
@@ -0,0 +1,1603 @@
...
+        if s-flow-line-prefix n:
+          return folded
+    return null
+
+  nb-ns-plain-in-line c/int -> string:
+    mark := mark
+    repeat: (repeat: match-chars S-WHITESPACE_) and ns-plain-char c
+    return string-since mark
+
+  NS-PLAIN-SEMI_SAFE_ ::= { C-MAPPING-KEY_, C-MAPPING-VALUE_, C-SEQUENCE-ENTRY_ }
```
floitsch, line 982:
```suggestion
  static NS-PLAIN-SEMI_SAFE_ ::= { C-MAPPING-KEY_, C-MAPPING-VALUE_, C-SEQUENCE-ENTRY_ }
```

floitsch (commented):
Suggestions to fix two warnings.

lib/encoding/yaml/parser.toit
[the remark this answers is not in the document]
mikkeldamsgaard: Right.

lib/encoding/yaml/parser.toit
mikkeldamsgaard: Good catch. Added a test case as guard.

PR #1968, opened 2023-11-24
Warn if 'if' and 'else' are not at same indentation.

### LINKED ISSUE #1967: `Else` can match an `if` at the wrong level of indentation.
```
main:
  if false:
    if false:
    print "foo"
    else:
      print "bar"
```
Produces no error, no warning, but prints "bar".

floitsch:
Fixes #1967.


kasperl (approved):
Looks good to me.

PR #1969, opened 2023-11-24
Allow ':' of a method/declaration to be at detent level.

floitsch:
This allows us to write something like:
```
foo
    foo
    bar
    --my-named-argument
    --other-argument
    [and-a-block]
:
  body
```

### DIFF
--- a/src/compiler/parser.cc
+++ b/src/compiler/parser.cc
@@ -948,6 +948,7 @@ Declaration* Parser::parse_declaration(bool is_abstract) {
     if (peek_token() == Token::COLON &&
         indentation_stack_.top_indentation() == current_indentation() &&
         indentation_stack_.is_outmost(IndentationStack::DECLARATION_SIGNATURE)) {
+      // Consume the detent.
       consume();
     }
   }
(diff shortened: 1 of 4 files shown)


src/compiler/parser.cc
```diff
@@ -937,6 +937,20 @@ Declaration* Parser::parse_declaration(bool is_abstract) {
...
+  //     body
+  if (current_token() == Token::DEDENT) {
+    if (peek_token() == Token::COLON &&
+        indentation_stack_.top_indentation() == current_indentation() &&
+        indentation_stack_.is_outmost(IndentationStack::DECLARATION_SIGNATURE)) {
+      consume();
+    }
+  }
   Sequence* body;
   if (current_token() == Token::COLON) {
```
kasperl, line 955: Doesn't this allow too much?

```
foo
    x
::
  return x
```

floitsch: It only consumes the DEDENT token.
Added a comment.
done.

kasperl: Thanks!

kasperl (approved):
Looks good to me!

PR #1970, opened 2023-11-24
Added chmod, link and readlink to file primitives

mikkeldamsgaard:
Note: This was not a straightforward implementation on the windows platform. The way windows handles symbolic links made it necessary to reimplement quite a few primitive using the WinAPI functions to get the correct behaviour.

There is also a corresponding PR to pkg-host.


src/primitive_file_non_win.cc
```diff
@@ -438,6 +442,47 @@ PRIMITIVE(chdir) {
...
+  if (!backing) FAIL(ALLOCATION_FAILED);
+
+  int result = FILE_READLINK_(process->current_directory(), pathname,
+                              reinterpret_cast<char*>(backing), PATH_MAX);
+  if (result < 0) return return_open_error(process, errno);
+
+  String* string = process->allocate_string(result);
+  if (!string) FAIL(ALLOCATION_FAILED);
+
+  strncpy(string->as_cstr(), reinterpret_cast<char*>(backing), result);
```
erikcorry, line 481: Please use a String::MutableBytes object here.  The as_cstr is intended for getting a null terminated char* out of a string for passing to C APIs.

src/primitive_file_win.cc
```diff
@@ -309,88 +315,124 @@ PRIMITIVE(read) {
...
+      WINDOWS_ERROR;
     }
     return process->null_object();
   }
 }
 
-Object* time_stamp(Process* process, time_t time) {
-  return Primitive::integer(time * 1000000000ll, process);
+int64 low_high_to_int64(DWORD high, DWORD low) {
+  return (((int64)high) << 32) + low;
```
erikcorry, line 346: We don't allow C-style casts.
```suggestion
  int64 result = high;
  result <<= 32;
  result += low
  return result;
```

src/primitive_file_win.cc
```diff
@@ -309,88 +315,124 @@ PRIMITIVE(read) {
...
+          type = UNIX_CHARACTER_DEVICE;
+          break;
+        case FILE_TYPE_DISK:
+          type = UNIX_REGULAR_FILE;
+          break;
+        case FILE_TYPE_PIPE:
+          type = UNIX_SOCKET;
+          break;
+        default:
+          type = UNIX_REGULAR_FILE;
```
erikcorry, line 406: Please throw something here.

src/primitive_file_win.cc
```diff
@@ -438,6 +480,52 @@ PRIMITIVE(chdir) {
...
+  if (result == 0) WINDOWS_ERROR;
+  return process->null_object();
+}
+
+PRIMITIVE(readlink) {
+  ARGS(WindowsPath, path);
+  HANDLE hFile = CreateFileW(path, 0, 0,
+                             NULL, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS | FILE_ATTRIBUTE_NORMAL, NULL);
+  if (hFile == INVALID_HANDLE_VALUE) WINDOWS_ERROR;
+  DWORD result_length = GetFinalPathNameByHandleW(hFile, NULL, 0, 0);
```
erikcorry, line 507: Please use AutoCloser

src/primitive_file_non_win.cc
```diff
@@ -438,6 +442,47 @@ PRIMITIVE(chdir) {
...
 
+PRIMITIVE(chmod) {
+  ARGS(cstring, pathname, int, mode);
+  int result = FILE_CHMOD_(current_dir(process), pathname, mode, 0);
+  if (result < 0) return return_open_error(process, errno);
+  return process->null_object();
+}
+
+// Notice that link does not use the *at(fd,...) version, as the symlink call does not
+// support it. Therefore, source and target must be absolute, this should be handled in the
```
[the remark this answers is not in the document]
mikkeldamsgaard: I might read the wrongly, but it appears only the target can be relative:
`The symlinkat() system call is equivalent to symlink() except in the case where name2 specifies a relative path.  In this case the symbolic link is
     created relative to the directory associated with the file descriptor fd instead of the current working directory.  If symlinkat() is passed the special
     value AT_FDCWD in the fd parameter, the current working directory is used and the behavior is identical to a call to symlink().`
In all likelihood, both name1 and name2 are then relative to fd.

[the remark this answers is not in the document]
erikcorry: The target (name1) is always relative to the linkpath (name2), never relative to the cwd.

With symlink, the linkpath (name2) is relative to the cwd, which is bad in Toit, this is why we should use symlinkat, where the linkpath is relative to the given file descriptor.

erikcorry (approved):

src/primitive_file_non_win.cc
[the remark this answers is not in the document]
mikkeldamsgaard: Got it (finally)

erikcorry (approved):

PR #1971, opened 2023-11-28
Give project-uri to compiler actions.

floitsch:
Eventually we would like to pass project-uris that are depending on the URIs we are working on. This is a first step in that direction. Instead of having the project-URI in the compiler, we pass it for every action.

### DIFF
--- a/tools/lsp/server/compiler.toit
+++ b/tools/lsp/server/compiler.toit
@@ -44,32 +44,29 @@ class Compiler:
   on-error_            /Lambda?     ::= ?
   timeout-ms_          /int         ::= ?
   protocol             /FileServerProtocol ::= ?
-  project-uri_         /string?     ::= ?
 
   constructor
       .compiler-path_
       .uri-path-translator_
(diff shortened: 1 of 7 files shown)


tools/lsp/server/compiler.toit
```diff
@@ -44,32 +44,29 @@ class Compiler:
...
       --on-crash/Lambda?=null:
     on-crash_ = on-crash
     on-error_ = on-error
-    project-uri_ = project-uri
 
   /**
   Builds the flags that are passed to the compiler.
   */
-  build-run-flags -> List:
+  build-run-flags --project-uri/string -> List:
```
kasperl, line 61:
```suggestion
  build-run-flags --project-uri/string? -> List:
```

kasperl (approved):
Looks good to me.

tools/lsp/server/compiler.toit
floitsch: done.

PR #1972, opened 2023-11-28
Analyze each file in its project root.

floitsch:
Instead of having one shared project root, find the project root for each file before running the analysis on it.

Still missing:
- we shouldn't report diagnostics for files that are in different project roots. Otherwise, we could end up with files that have a warning when they are not opened, and then one when they aren't (depending on the project root that is used to analyze them).
- if a summary has changed, then all projects that contain that file should be analyzed again.

### DIFF
--- a/tools/lsp/server/documents.toit
+++ b/tools/lsp/server/documents.toit
@@ -15,52 +15,157 @@
 
 import host.file
 import tar show *
-import .summary
 
+import .project-uri
+import .summary
 import .uri-path-translator
 import .utils
 
 /**
-Keeps track of unsaved files.
+Keeps track of unsaved files and their dependencies.
 */
 class Documents:
-  documents_ /Map/*<string, Document>*/ ::= {:}
+  /**
+  A map from document uri to opened document.
+  During analysis, this map must be used to find the content of a document.
+  */
+  opened-documents_ /Map/*<string, OpenedDocument>*/ ::= {:}
+
+  /**
+  A map from a document URI to its project URI.
+
+  The project URI is the root of the project where we can find the
+    package lock file and the downloaded packages.
+
+  From the user's point of view a document is only in one project. Diagnostics
+    are only shown for this project.
+
+  Internally, a document might be in more than one project, as local dependencies
+    can lead to a document being referenced from multiple projects.
+  */
+  project-uris_ /Map/*<string, string>*/ ::= {:}
+
+  /**
+  A map from project-uri to a $AnalyzedDocuments object.
+  Note that URIs might be in more than one project.
+  */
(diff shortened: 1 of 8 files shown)


tools/lsp/server/documents.toit
```diff
@@ -15,52 +15,157 @@
...
   error-reporter_ / Lambda ::= ?
 
-  constructor .translator_ --.error-reporter_=(:: /* do nothing */):
+  constructor .translator_ --error-reporter/Lambda=(:: /* do nothing */):
+    error-reporter_ = error-reporter
+
+  /**
+  The project-uri the given document uri belongs to.
+
+  If the document isn't known yet computes its project-uri.
```
kasperl, line 63:
```suggestion
  If the document isn't known yet, computes its project-uri.
```

tools/lsp/server/documents.toit
```diff
@@ -15,52 +15,157 @@
...
+      documents.delete --uri=uri
+
+  get-opened-document_ --uri/string -> OpenedDocument:
+    return opened-documents_.get uri --init=:
+      error-reporter_.call "Document $uri doesn't exist yet"
+      OpenedDocument --uri=uri --revision=-1 --content=""
+
+  get-opened --uri/string -> OpenedDocument?:
+    return opened-documents_.get uri
+  get-opened --path/string -> OpenedDocument?:
```
kasperl, line 116: Add newline for consistency?

tools/lsp/server/documents.toit
```diff
@@ -15,52 +15,157 @@
...
+    open-document := opened-documents_.get uri
+    content-revision := open-document ? open-document.revision : -1
+    return analyzed.update-document-after-analysis
+        --uri=uri
+        --analysis-revision=analysis-revision
+        --summary=summary
+        --content-revision=content-revision
+
+/**
+Keeps track of unsaved files.
```
kasperl, line 150: Is this really only used for unsaved files?

tools/lsp/server/documents.toit
```diff
@@ -15,52 +15,157 @@
...
+  // For each project-uri a set of document URIs that are relevant
+  // for the project.
+  // Note that this information is redundant with the information in
+  // `documents_`, but it is more efficient to keep it here.
+  documents-in-project_ /Map/*<string, Set<string>>*/ ::= {:}
+
+  translator_ /UriPathTranslator ::= ?
+  error-reporter_ / Lambda ::= ?
+
+  constructor .translator_ --error-reporter:
```
kasperl, line 167:
```suggestion
  constructor .translator_ --error-reporter/Lambda:
```

kasperl (approved):
Looks good to me.

PR #1973, opened 2023-11-28
Only emit diagnostics for the current project.

floitsch:
If we have dependent projects, this means that we currently don't see updates to diagnostics if the file is not sharing the same project root. We will fix this in a later PR.


kasperl (approved):
Looks good to me.

PR #1976, opened 2023-11-29
Avoid getting stuck forever on monitor.Signal

kasperl:


lib/monitor.toit
```diff
@@ -169,17 +170,20 @@ monitor Signal:
 
   // Helper method for condition waiting.
   wait_ [condition] -> none:
-    while true:
-      awaited := awaited_
-      if current_ == awaited:
```
kasperl, line 174: This is not enough to guarantee that nobody is waiting. The first `wait` after a signal will always hit this even if there are waiters where the condition hasn't yet been met.

erikcorry (approved):

PR #1978, opened 2023-11-29
Don't sync by default.

floitsch:
If developers want to sync they can just write `toit.pkg sync`.


floitsch:
Unfortunately not as easy as I thought.

PR #1979, opened 2023-12-01
New byte code for branch on not null

erikcorry:


src/compiler/propagation/type_propagator.cc
```diff
@@ -1173,6 +1173,13 @@ static TypeScope* process(TypeScope* scope, uint8* bcp, std::vector<Worklist*>&
     }
   OPCODE_END();
 
+  OPCODE_BEGIN(BRANCH_IF_NOT_NULL);
```
kasperl, line 1176: Maybe this should also be treated as an unconditional branch if the variable cannot be null?

PR #1984, opened 2023-12-04
Added timestamp option to logger.

mikkeldamsgaard:


lib/log/target.toit
```diff
@@ -27,8 +27,13 @@ Standard log service used when the system log service cannot
 class StandardLogService_ implements LogService:
   buffer_/bytes.Buffer ::= bytes.Buffer.with-initial-size 64
 
-  log level/int message/string names/List? keys/List? values/List? -> none:
+  log level/int message/string names/List? keys/List? values/List? timestamp/bool -> none:
     buffer ::= buffer_
+
+    if timestamp:
+      buffer.write "$(Time.monotonic-us / 1000)"
```
floitsch, line 34: The digits should be padded. Space is probably enough, but 0 would work too.
Not sure what the correct digit-amount is.

lib/log/log.toit
```diff
@@ -148,27 +153,50 @@ class Logger:
...
   Adds the tag composed by the $key and the $value to a copy of this logger.
   */
   with-tag key/string value -> Logger:
-    return Logger.internal_ this --tags={key: value}
+    return with --tags={key: value}
+
+  /**
+  Creates a copy where timestamps are enabled.
+  */
+  with-timestamp -> Logger:
```
floitsch, line 197: I could be convinced that `with-timestamps` is better than `with-timestamped`.

PR #1995, opened 2023-12-08
Fix 'glob'.

floitsch:
The documentation wasn't shown because it had a new line. I also created a private glob_ so that the named argument isn't visible to users anymore.

### DIFF
--- a/lib/core/string.toit
+++ b/lib/core/string.toit
@@ -714,30 +714,30 @@ abstract class string implements Comparable:
   ```
   */
   glob pattern/string -> bool:
-    return glob_ pattern --position_=0
+    return glob_ pattern --position=0
 
-  glob_ pattern/string --position_/int -> bool:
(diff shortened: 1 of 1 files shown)


lib/core/string.toit
```diff
@@ -715,8 +713,10 @@ abstract class string implements Comparable:
   "Toad".glob "To\\*d" // => false
   ```
   */
+  glob pattern/string -> bool:
+    return glob_ pattern --position_=0
 
-  glob pattern/string --position_/int=0 -> bool:
+  glob_ pattern/string --position_/int -> bool:
```
kasperl, line 719:
```suggestion
  glob_ pattern/string --position/int -> bool:
```

kasperl (approved):
Looks good to me.

lib/core/string.toit
floitsch: done.

PR #1996, opened 2023-12-11
Add verbatim copy of BufferedReader to lib.io.

floitsch:


lib/io.toit
```diff
@@ -1,5 +1,6 @@
 // Copyright (C) 2023 Toitware ApS. All rights reserved.
 
+import .reader
 /**
```
kasperl, line 4: I'd add a newline here.

kasperl (approved):
Looks good to me.

lib/io.toit
floitsch: done.

PR #2003, opened 2023-12-13
Analyze each URI in its own project.

floitsch:
Same as #1972, but for the Go version of the LSP.

Instead of having one shared project root, find the project root for each file before running the analysis on it.

Still missing:
- we shouldn't report diagnostics for files that are in different project roots. Otherwise, we could end up with files that have a warning when they are not opened, and then one when they aren't (depending on the project root that is used to analyze them).
- if a summary has changed, then all projects that contain that file should be analyzed again.

### DIFF
--- a/tools/toitlsp/lsp/project_uri.go
+++ b/tools/toitlsp/lsp/project_uri.go
@@ -25,20 +25,21 @@ import (
 )
 
 func computeProjectURI(documentUri lsp.DocumentURI) (lsp.DocumentURI, error) {
-	segments := strings.Split(filepath.ToSlash(string(documentUri)), "/")
+	path := uri.URIToPath(documentUri)
+	segments := strings.Split(filepath.ToSlash(path), "/")
 	// Find the first '.packages' segment. We assume that the project root is
 	// the parent of this segment.
 	for i := 0; i < len(segments); i++ {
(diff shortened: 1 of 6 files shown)


tools/toitlsp/lsp/project_uri.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/toitware/toit.git/toitlsp/lsp/uri"
+)
+
+func computeProjectURI(documentUri lsp.DocumentURI) (lsp.DocumentURI, error) {
+	path := uri.URIToPath(documentUri)
+	segments := strings.Split(filepath.ToSlash(path), "/")
+	// Find the first '.packages' segment. We assume that the project root is
+	// the parent of this segment.
+	for i := 0; i < len(segments); i++ {
+		if segments[i] == ".packages" {
```
kasperl, line 33: I wonder if we should have called this something more Toit-specific.

kasperl (approved):
Looks good to me.

tools/toitlsp/lsp/project_uri.go
floitsch: Same for `package.yaml` and `package.lock`.

We should maybe try to rename them...

PR #2004, opened 2023-12-14
Only emit diagnostics for URIs that are in the same project.

floitsch:
This is the Golang port of #1973.

If we have dependent projects, this means that we
currently don't see updates to diagnostics if the file is not sharing the same project root.

We will fix this in the next PR.


kasperl (approved):
Looks good to me.

PR #2005, opened 2023-12-14
Propagate changed summaries to other projects

floitsch:
This is the Go port of #1974.

The test is a copy and should merge cleanly.


kasperl (approved):
Looks good to me.

PR #2006, opened 2023-12-14
Fix local/slash confusion on package-path.

floitsch:
On Windows we were joining a local path (`..\foo`) to a filesystem that was using slashes (the LSP filesystem). This lead to the path not being correctly canonicalized.

### DIFF
--- a/src/compiler/util.h
+++ b/src/compiler/util.h
@@ -124,25 +124,22 @@ class PathBuilder {
 
   // Joins the given sub_path to this builder.
   // The sub_path must use slashes as path separators.
-  void joinSlashPath(const std::string& sub_path) {
+  void join_slash_path(const std::string& sub_path) {
     if (sub_path.empty()) return;
-    size_t last_start = 0;
-    size_t pos = 0;
-    if (sub_path[0] == '/') {
-      join("/");
-      last_start = 1;
-    }
-    while (pos < sub_path.size()) {
-      if (sub_path[pos] == '/') {
-        if (pos > last_start) {
-          join(sub_path.substr(last_start, pos - last_start));
-        }
-        last_start = pos + 1;
+    size_t start_pos = 0;
+    while (start_pos < sub_path.size()) {
+      auto slash_pos = sub_path.find_first_of('/', start_pos);
+      if (slash_pos == std::string::npos) {
+        // No more slashes, use the rest of the string.
+        slash_pos = sub_path.size();
       }
-      pos++;
-    }
-    if (pos > last_start) {
-      join(sub_path.substr(last_start, pos - last_start));
+      // For absolute paths we need to add the leading "/".
+      if (slash_pos == 0 && buffer_.empty()) {
+        buffer_ += fs_->path_separator();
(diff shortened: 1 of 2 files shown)


src/compiler/util.h
```diff
@@ -122,6 +122,30 @@ class PathBuilder {
     join(segment4);
   }
 
+  // Joins the given sub_path to this builder.
+  // The sub_path must use slashes as path separators.
+  void joinSlashPath(const std::string& sub_path) {
```
kasperl, line 127:
```suggestion
  void join_slash_path(const std::string& sub_path) {
```

src/compiler/util.h
```diff
@@ -122,6 +122,30 @@ class PathBuilder {
...
+    if (sub_path.empty()) return;
+    size_t last_start = 0;
+    size_t pos = 0;
+    if (sub_path[0] == '/') {
+      join("/");
+      last_start = 1;
```
kasperl, line 133: It feels a bit odd that `pos` is zero after this. You might as well **not** set `last_start` to 1 here, even though that would be (perhaps) even odder.

src/compiler/util.h
```diff
@@ -122,6 +122,30 @@ class PathBuilder {
     join(segment4);
   }
 
+  // Joins the given sub_path to this builder.
+  // The sub_path must use slashes as path separators.
+  void joinSlashPath(const std::string& sub_path) {
+    if (sub_path.empty()) return;
+    size_t last_start = 0;
```
kasperl, line 129: This seems pretty convoluted. Wouldn't it be easier to write with `sub_path.find_first_of`?

kasperl (approved):
Looks good to me.

src/compiler/util.h
floitsch: Completely rewritten.

src/compiler/util.h
floitsch: Completely rewritten.

src/compiler/util.h
```diff
@@ -122,6 +122,27 @@ class PathBuilder {
...
+    while (start_pos < sub_path.size()) {
+      auto slash_pos = sub_path.find_first_of('/', start_pos);
+      if (slash_pos == std::string::npos) {
+        // No more slashes, use the rest of the string.
+        slash_pos = sub_path.size();
+      }
+      // For absolute paths we need to add the leading "/".
+      if (slash_pos == 0 && buffer_.empty()) {
+        buffer_ += fs_->path_separator();
+      } else if (slash_pos > start_pos) {
```
kasperl, line 139: Maybe explain the guard against `slash_pos == start_pos` in a comment? Are you collapsing repeated slashes?

kasperl (approved):
Much better!

src/compiler/util.h
floitsch: I removed that 'if'.
I don't think we should ever get multiple '/' and there is no reason we should pick one over the other. Also, we do a canonicalize later, which would remove it anyway.

PR #2009, opened 2023-12-15
Fix build not going red with external/flaky.

floitsch:


.github/workflows/ci.yml
```diff
@@ -961,14 +961,23 @@ jobs:
           name: ${{ steps.constants.outputs.external_result_name }}
           path: external-result
 
-      - name: Check
+      - name: Debug out
```
kasperl, line 964: What is this? Maybe improve the name if this is something that we should look at?

.github/workflows/ci.yml
```diff
@@ -961,14 +961,23 @@ jobs:
...
+
+      - name: Check external
+        shell: bash
+        run: |
           cat external-result/*
-          ! grep "fail" external-result/*
+          grep "fail" external-result/* && exit 1
+
+      - name: Check flaky
+        if: always()
```
kasperl, line 976: Can you add a comment that explains why this condition is here, but not on "Check external"?

kasperl (approved):
Looks good to me.

.github/workflows/ci.yml

.github/workflows/ci.yml
floitsch: done.

PR #2014, opened 2023-12-18
Avoid non-zero flaky/external results on success

kasperl:


.github/workflows/ci.yml
```diff
@@ -972,6 +972,11 @@ jobs:
         run: |
           cat external-result/*
           grep "fail" external-result/* && exit 1
+          # If we get here, we didn't find any failures but
+          # the exit code of the grep is still 1 or 2. Echo
+          # something to make it clear and avoid letting the
+          # result of the action be non-zero.
+          echo "success: no failures found"
```
floitsch, line 979: Maybe instead:
```suggestion
          (! grep "fail" external-result/*) || exit 1
```

.github/workflows/ci.yml
```diff
@@ -982,6 +987,11 @@ jobs:
         run: |
           cat flaky-result/*
           grep "fail" flaky-result/* && exit 1
+          # If we get here, we didn't find any failures but
+          # the exit code of the grep is still 1 or 2. Echo
+          # something to make it clear and avoid letting the
+          # result of the action be non-zero.
+          echo "success: no failures found"
```
floitsch, line 994:
```suggestion
          (! grep "fail" flaky-result/*) || exit 1
```

floitsch (approved):
LGTM.
I try to avoid negative command lines, as I would like to make every bash abort on the first failure. I thought this was already the case, but if it isn't we should find a way to make it so.

PR #2015, opened 2023-12-18
Use the last '.packages' as project root.

floitsch:
If we run `toit.pkg install` inside the .packages directory we now see the installed files.

Pretty much a corner case that shouldn't be super important, but this feels more correct.


kasperl (approved):
Looks good to me.

Last is the innermost, which does make more sense.

PR #2017, opened 2023-12-18
Make the settings in the lsp server nicer.

floitsch:


tools/lsp/server/server.toit
```diff
@@ -53,9 +53,7 @@ CRASH-REPORT-RATE-LIMIT-MS ::= 30_000
 monitor Settings:
```
kasperl, line 53: Why is this a monitor?

kasperl (approved):
Looks good to me.

tools/lsp/server/server.toit
floitsch: Because it blocks the settings access if a `request-settings` action is in process.

kasperl: It probably would be more common to just implement this with a mutex, but it makes sense.

PR #2023, opened 2023-12-18
Fix escapes in toitdocs.

floitsch:


src/compiler/toitdoc_parser.cc
```diff
@@ -739,14 +739,18 @@ Symbol ToitdocParser::parse_delimited(int delimiter,
...
       } else {
         buffer += make_string(chunk_start, index_);
         advance();
         chunk_start = index_;
-        advance();
       }
+      // Set 'c' to something that isn't a delimiter, so that
+      // we continue iterating. The escaped character will be
+      // consumed at the top of the loop.
+      c = 'X';
```
kasperl, line 753: Isn't c already `\` here? That isn't a delimiter anywhere and I'm not sure things would work if it actually was a delimiter because of the early check for `c == '\\'`.

tests/toitdoc-escapes-test.toit
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2023 Toitware ApS.
+// Use of this source code is governed by a Zero-Clause BSD license that can
+// be found in the tests/LICENSE file.
+
+/**
+Tests that the toitdoc compiler can deal with escapes..
```
kasperl, line 6:
```suggestion
Tests that the toitdoc compiler can deal with escapes.
```

kasperl (approved):
Looks good to me.

src/compiler/toitdoc_parser.cc
floitsch: Right.
Removed.

PR #2038, opened 2024-01-02
Yaml parse value conversion fix

mikkeldamsgaard:
Fixed a situation where the automatic value conversion would convert "1.1" to a float and "5" to a int and not keep them as string.


floitsch (approved):
LGTM.

PR #2040, opened 2024-01-02
Add 'system.program-path'.

floitsch:
Contrary to 'system.program-name' this gives an absolute resolved path to the executable.


src/os_win.cc
```diff
@@ -33,6 +33,63 @@
...
+}
+
+static char* malloced_narrow_string(const wchar_t* string) {
+  word length = wcslen(string);
+  return malloced_narrow_string(string, length);
+}
+
+char* OS::get_executable_path() {
+  char* path = _new char[MAX_PATH + 1];
+  auto length = GetModuleFileName(NULL, path, MAX_PATH);
```
floitsch, line 60: A follow-up PR changes this to use the wide variant.

lib/system/system.toit
```diff
@@ -209,9 +209,28 @@ object-histogram_ marker/string full-gcs/int? -> ByteArray:
...
+
+If the program is run by the Toit command line tool, this is the name of the
+  source file passed to the tool.
+If the program is run as an executable, this is the name of the executable. In
+  this case it is equivalent to argv[0] in C, or $0 in bash.
+
+May return null if this information is not available.
 */
 program-name -> string?:
   #primitive.core.command
```
kasperl, line 223: I'd probably change the primitive name while you're at it.

src/os_win.cc
```diff
@@ -33,6 +33,63 @@
...
+static wchar_t* malloced_wide_string(const char* string) {
+  word length = Utils::utf_8_to_16(unsigned_cast(string), strlen(string));
+  wchar_t* result = reinterpret_cast<wchar_t*>(malloc((length + 1) * sizeof(wchar_t)));
+  Utils::utf_8_to_16(unsigned_cast(string), strlen(string), result, length);
+  result[length] = '\0';
+  return result;
+}
+
+
+static char* malloced_narrow_string(const wchar_t* string, word w_length) {
```
kasperl, line 45:
```suggestion
}

static char* malloced_narrow_string(const wchar_t* string, word w_length) {
```

src/os_win.cc
```diff
@@ -33,6 +33,63 @@
...
+
+char* OS::get_executable_path() {
+  char* path = _new char[MAX_PATH + 1];
+  auto length = GetModuleFileName(NULL, path, MAX_PATH);
+  path[length] = '\0';
+  return path;
+}
+
+char* OS::get_executable_path_from_arg(const char* source_arg) {
+  wchar_t* w_source_arg = malloced_wide_string(source_arg);
```
kasperl, line 66: I find it kinda odd that these have a `w_` prefix and not a `_w` suffix. YMMV.

src/os_win.cc
```diff
@@ -33,6 +33,63 @@
 
 namespace toit {
 
+static wchar_t* malloced_wide_string(const char* string) {
```
kasperl, line 36: The names of these functions are kinda weird. They are sort of based on the type, because they are conversions. I think it would be more natural to have them named on what they do -- something like convert or allocate instead of converted or allocated.

`malloc_wide_string`
`new_wide_string`
`convert_to_wide_string`
`to_wide_string`

tests/health/gold/sdk/tests__program-path-test.toit.gold
```diff
@@ -0,0 +1,6 @@
+<pkg:pkg-host>/directory.toit:154:6: warning: Deprecated 'platform'
```
kasperl, line 1: Why are we getting warnings for a package that we depend on here?

kasperl (approved):
Looks good to me.

src/os_win.cc
```diff
@@ -33,6 +33,63 @@
...
+  wchar_t* w_source_arg = malloced_wide_string(source_arg);
+
+  DWORD w_result_length = GetFullPathNameW(w_source_arg, 0, NULL, NULL);
+  if (w_result_length == 0) {
+    free(w_source_arg);
+    return null;
+  }
+
+  wchar_t* w_result = unvoid_cast<wchar_t*>(malloc(w_result_length * sizeof(wchar_t)));
+  if (w_result == null) {
```
kasperl, line 75: I don't think you need to worry about malloc failing on windows.

src/os_win.cc
floitsch: Went with `to_wide_string` and `to_narrow_string`.
done.

src/os_win.cc
floitsch: Changed (hopefully consistently) to `_w`.
done.

src/os_win.cc
floitsch: removed.
done.

tests/health/gold/sdk/tests__program-path-test.toit.gold
floitsch: The SDK health-check runs with `--show-package-warnings` so we get warnings from inside the SDK itself (which is treated like a package by the compiler).

PR #2041, opened 2024-01-02
Use wide variant to find executable path.

floitsch:


src/os_win.cc
```diff
@@ -56,10 +55,14 @@ static char* malloced_narrow_string(const wchar_t* string) {
...
-  path[length] = '\0';
-  return path;
+  const int BUFFER_SIZE = 32767 + 1;
+  wchar_t buffer[BUFFER_SIZE];
+  word w_length = GetModuleFileNameW(NULL, buffer, BUFFER_SIZE);
+  // GetModuleFileNameW truncates the path to the buffer size.
+  // If the returned length is equal to the BUFFER_SIZE we assume that the
+  // buffer wasn't big enough.
+  if (w_length == 0 || w_length >= BUFFER_SIZE) return null;
+  return  malloced_narrow_string(buffer, w_length);
```
kasperl, line 65:
```suggestion
  return malloced_narrow_string(buffer, w_length);
```

src/os_win.cc
```diff
@@ -56,10 +55,14 @@ static char* malloced_narrow_string(const wchar_t* string) {
 }
 
 char* OS::get_executable_path() {
-  char* path = _new char[MAX_PATH + 1];
-  auto length = GetModuleFileName(NULL, path, MAX_PATH);
-  path[length] = '\0';
-  return path;
+  const int BUFFER_SIZE = 32767 + 1;
```
kasperl, line 58: This feels like too much to put on the stack. I'd just use malloc and expect it to return a non-null value.

kasperl (approved):
Looks good to me.

src/os_win.cc
floitsch: done.

src/os_win.cc
floitsch: I copied this from another function.
Will do a follow-up PR where I change both to mallocs.

PR #2042, opened 2024-01-03
Don't allocate 32K on the stack.

floitsch:
Use malloc to allocate the temporary buffer.

### DIFF
--- a/src/os_win.cc
+++ b/src/os_win.cc
@@ -62,13 +62,18 @@ static char* to_narrow_string(const wchar_t* string_w) {
 
 char* OS::get_executable_path() {
   const int BUFFER_SIZE = 32767 + 1;
-  wchar_t buffer[BUFFER_SIZE];
+  wchar_t* buffer = unvoid_cast<wchar_t*>(malloc(BUFFER_SIZE));
   word length_w = GetModuleFileNameW(NULL, buffer, BUFFER_SIZE);
   // GetModuleFileNameW truncates the path to the buffer size.
   // If the returned length is equal to the BUFFER_SIZE we assume that the
   // buffer wasn't big enough.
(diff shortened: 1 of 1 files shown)


src/os_win.cc
```diff
@@ -326,11 +331,16 @@ char* OS::getenv(const char* variable) {
   wchar_t* variable_w = to_wide_string(variable);
 
   const int BUFFER_SIZE = 32767;
```
kasperl, line 333: No +1 here?

src/os_win.cc
```diff
@@ -326,11 +331,16 @@ char* OS::getenv(const char* variable) {
...
-  wchar_t buffer[BUFFER_SIZE];
+  wchar_t* buffer = unvoid_cast<wchar_t*>(malloc(BUFFER_SIZE));
   int length_w = GetEnvironmentVariableW(variable_w, buffer, BUFFER_SIZE);
   free(variable_w);
-  if (length_w == 0 || length_w > BUFFER_SIZE) return null;
-  return to_narrow_string(buffer, length_w);
+  if (length_w == 0 || length_w > BUFFER_SIZE) {
```
kasperl, line 337: This is different than the other check.

kasperl (approved):
Looks good to me.

src/os_win.cc
floitsch: Not needed.
The `GetEnvironmentVariable` function behaves differently...
The `GetModuleFileNameW` function truncates and returns the number of written characters.
The `GetEnvironmentVariableW` function returns the needed length.

src/os_win.cc
floitsch: See the other response.

I added a comment.

PR #2045, opened 2024-01-04
Replace some #ifdef TOIT_FREERTOS with #ifdef TOIT_ESP

floitsch:


kasperl (approved):
Looks good to me, but I think I'd prefer `TOIT_ESP32`.

floitsch:
Done. (switched to TOIT_ESP32).

PR #2053, opened 2024-01-11
Allow scanning when already connected to WiFi.

floitsch:


system/extensions/esp32/wifi.toit
```diff
@@ -152,7 +152,13 @@ class WifiServiceProvider extends NetworkServiceProviderBase:
...
   scan config/Map -> List:
     if state_.module:
-      throw "wifi already connected or established"
+      if state_.module is WifiModule:
+        channels := config.get wifi.CONFIG-SCAN-CHANNELS
+        passive := config.get wifi.CONFIG-SCAN-PASSIVE
+        period := config.get wifi.CONFIG-SCAN-PERIOD
+        return (state_.module as WifiModule).scan channels passive period
+      else:
+        throw "wifi already connected or established"
```
floitsch, line 161: Not sure this can happen, or the error message is correct.

system/extensions/esp32/wifi.toit
```diff
@@ -152,7 +152,13 @@ class WifiServiceProvider extends NetworkServiceProviderBase:
 
   scan config/Map -> List:
     if state_.module:
-      throw "wifi already connected or established"
+      if state_.module is WifiModule:
```
kasperl, line 155: It is always null or a WifiModule here.

system/extensions/esp32/wifi.toit
```diff
@@ -152,7 +152,13 @@ class WifiServiceProvider extends NetworkServiceProviderBase:
...
+        channels := config.get wifi.CONFIG-SCAN-CHANNELS
+        passive := config.get wifi.CONFIG-SCAN-PASSIVE
+        period := config.get wifi.CONFIG-SCAN-PERIOD
+        return (state_.module as WifiModule).scan channels passive period
```
kasperl, line 159: I think we need to make it so you can't connect while scanning, but you can scan while connected.

kasperl (approved):
Looks good to me.

system/extensions/esp32/wifi.toit
floitsch: done.

PR #2060, opened 2024-01-19
Add Toit PKG manager

floitsch:
This is a duplicate of https://github.com/toitlang/toit/pull/2044.


lib/core/collections.toit
```diff
@@ -2456,6 +2456,13 @@ class Set extends HashedInsertionOrderedCollection_ implements Collection:
     do: if not predicate.call it: return false
     return true
 
+  /**
+  Copies the set.
+  The set is shallow.
```
kasperl, line 2461: This isn't an awesome comment.

Maybe a slightly better formatted comment that states something about the returned set would be nicer?

kasperl (approved):

lib/core/collections.toit
floitsch: done.

PR #2070, opened 2024-01-31
Add compare-equal-vector

erikcorry:


kasperl (approved):
As discussed, we should probably wait with exposing the method until we know where we want it. You can still use it in the test by just exposing the primitive there.

LGTM.

src/primitive.h
```diff
@@ -114,6 +114,7 @@ namespace toit {
   PRIMITIVE(add_entropy, 1)                  \
   PRIMITIVE(count_leading_zeros, 1)          \
   PRIMITIVE(popcount, 1)                     \
+  PRIMITIVE(compare_equal_vector, 2)         \
```
kasperl, line 117: I'd go ahead with the renaming we discussed. This isn't really a compare function and it doesn't return -1, 0, or 1. 

Something with vector equals something something.

kasperl (approved):
Looks good to me.

PR #2071, opened 2024-02-01
Add a way to check that TLS resume succeeded

erikcorry:
In a different change, the test file will be renamed.


lib/tls/session.toit
```diff
@@ -152,7 +159,20 @@ class Session:
       if symmetric-session_:
         return SESSION-MODE-TOIT
       else:
-        return SESSION-MODE-CLOSED
+        if state-bits_ | HANDSHAKE-ATTEMPTED_ == 0:
```
kasperl, line 162: wat? HANDSHAKE-ATTEMPTED_ is 1.

Did you mean `&`?

kasperl (approved):
Looks good to me.

PR #2076, opened 2024-02-06
Add test that uses regular network and global certs.

erikcorry:


floitsch (approved):
LGTM.
Mark the test preemptively as flaky?

floitsch (approved):

PR #2081, opened 2024-02-07
Optimized SHA256 for platforms with no acceleration.

erikcorry:


src/sha256_alt.h
```diff
@@ -0,0 +1,38 @@
...
+#ifdef MBEDTLS_SHA256_ALT
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define SHA_BLOCK_LEN ((size_t)64)
+
+typedef struct {
+  int bit_len;  // 224 or 256.
```
kasperl, line 27: len vs length?

kasperl (approved):
Looks good to me.

PR #2086, opened 2024-02-14
Keep track of which primitive is currently called.

floitsch:


src/interpreter_run.cc
```diff
@@ -1083,6 +1084,7 @@ Interpreter::Result Interpreter::run() {
 
       // GC might have taken place in object heap but local "method" is from program heap.
       PUSH(result);
+      process_->set_current_primitive_call(-1, -1);
```
kasperl, line 1087: This also has to be done in the non-failure case (around line 1100 post change). You could also argue that you just throw away interesting information here, so maybe you shouldn't reset this and just keep track of the last primitive call?

src/scheduler.cc
```diff
@@ -910,6 +910,12 @@ void Scheduler::tick(Locker& locker, int64 now) {
       // The process is already suppossed to preempt.
       // Check whether it is stuck.
       if (us_since_preemption <= MAX_RUN_WITHOUT_PREEMPTION_US) continue;
+      if (process->current_primitive_index() != -1) {
+        FATAL("Potential dead-lock detected in process %d; primitive call %d-%d\n",
```
kasperl, line 914: Let's make this consistent with https://github.com/toitlang/toit/blob/master/lib/core/exceptions.toit#L191.

```suggestion
        FATAL("Potential dead-lock detected in process %d; primitive call %d:%d\n",
```

src/process.h
```diff
@@ -268,6 +268,19 @@ class Process : public ProcessListFromProcessGroup::Element,
     return run_timestamp_;
   }
 
+  void set_current_primitive_call(int primitive_module, int primitive_index) {
```
kasperl, line 271: You could also just store the bcp and decode it later. It is a little bit more complex, because you have to decode the module and primitive index at the time this fails, but it feels a bit more robust and prepared for other cases (long running intrinsics, perhaps?).

kasperl (approved):
Looks good to me with a few suggestions.

src/interpreter_run.cc
floitsch: I think we want to clear the BCP again, as it has otherwise the potential to give us the wrong impression.
done.

src/process.h
floitsch: done.

src/scheduler.cc
floitsch: done.

src/scheduler.cc
```diff
@@ -910,6 +910,31 @@ void Scheduler::tick(Locker& locker, int64 now) {
...
+            static_cast<int>(Utils::read_unaligned_uint32_be(uuid)),
+            static_cast<int>(Utils::read_unaligned_uint16_be(uuid + 4)),
+            static_cast<int>(Utils::read_unaligned_uint16_be(uuid + 6)),
+            static_cast<int>(Utils::read_unaligned_uint16_be(uuid + 8)),
+            static_cast<int>(Utils::read_unaligned_uint16_be(uuid + 10)),
+            static_cast<int>(Utils::read_unaligned_uint32_be(uuid + 12)));
+        fprintf(stderr, "  Program: %s\n", uuid_buffer);
+        fprintf(stderr, "  BCI: 0x%x\n", bci);
+
+        if (*last_bcp == Opcode::PRIMITIVE) {
```
kasperl, line 931: Can we check if the last bcp is within the program area?

kasperl (approved):
Even better!

src/scheduler.cc
floitsch: done.
I'm only checking if the bcp points into the bytecodes. Further accesses (like getting the primitive parameters) are not checked. But it's unlikely that we have a valid bytecode pointer that points to the edge of the bytecodes.

PR #2088, opened 2024-02-15
Windows roots

erikcorry:


src/resources/tls.cc
```diff
@@ -645,6 +650,60 @@ PRIMITIVE(close) {
...
 }
 
+#define NEEDS_DELETE 1
+#define IN_FLASH 2
+#define IGNORE_UNSUPPORTED_HASH 4
+
+static Object* add_global_root(const uint8* data, size_t length, Object* hash, Process* process, int flags);
+
+#ifdef TOIT_WINDOWS
+static Object* get_roots_from_store(const HCERTSTORE store, Process* process) {
```
floitsch, line 660:
```suggestion
static Object* add_roots_from_store(const HCERTSTORE store, Process* process) {
```

src/resources/tls.cc
```diff
@@ -645,6 +650,60 @@ PRIMITIVE(close) {
   return process->null_object();
 }
 
+#define NEEDS_DELETE 1
```
floitsch, line 653: Not a fan of defines.
either make it a constant, an enum, or just pass the three values instead of the combined flag.

tests/tls-global-cert-simple-test.toit
```diff
@@ -124,6 +124,11 @@ connect-to-site host port expected-certificate-name:
     print "Read $bytes bytes from https://$host$(port == 443 ? "" : ":$port")/"
 
 add-global-certs -> none:
+  if system.platform == "Windows":
```
floitsch, line 127: Make a second test.
We want to test both cases.

lib/tls/certificate.toit
```diff
@@ -102,3 +102,17 @@ Returns the hash of the added certificate, which can be used to add it more
 */
 add-global-root-certificate_ cert hash/int?=null -> int:
   #primitive.tls.add-global-root-certificate
+
+/**
+Adds the trusted root certificates that are installed on the system.
+
+This is only supported on Windows.  On other platforms it currently does
+  nothing.
```
floitsch, line 110:
```suggestion
This is only supported on Windows.  On other platforms it does nothing.
```

floitsch (approved):
LGTM.

lib/tls/certificate.toit
erikcorry: We might add this functionality on MacOS too, so I think the original is better.

floitsch: We can then update the comment.

PR #2092, opened 2024-02-18
Add more syntax.

floitsch:
Add an example of 'as', which was broken: https://github.com/toitware/ide-tools/issues/246


kasperl (approved):
Looks good to me.

PR #2094, opened 2024-02-19
Refactor LSP selection handlers

floitsch:
Create the LSP selection handlers earlier and unify how they are called.


kasperl (approved):
Looks good to me.

PR #2100, opened 2024-02-20
Add a text-range to completions if the prefix ends with '-'

floitsch:
Golang implementation.

### DIFF
--- a/tools/toitlsp/lsp/doc_handlers.go
+++ b/tools/toitlsp/lsp/doc_handlers.go
@@ -131,7 +131,8 @@ func (s *Server) textDocumentDefinition(ctx context.Context, conn *jsonrpc2.Conn
 	return res, nil
 }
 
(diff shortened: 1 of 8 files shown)


tools/toitlsp/lsp/doc_handlers.go
```diff
@@ -142,15 +143,40 @@ func (s *Server) textDocumentCompletion(ctx context.Context, conn *jsonrpc2.Conn
...
 		})
 	}
-	return res, nil
+	if len(items) == 0 {
+		return items, nil
+	}
+	if !strings.HasSuffix(prefix, "-") {
+		return items, nil
+	}
+	// The prefix ends with a '-'. Vscode doesn't like that and assumes that any completion we
```
kasperl, line 160:
```suggestion
	// The prefix ends with a '-'. VS Code doesn't like that and assumes that any completion we
```

kasperl (approved):
Looks good to me.

tools/toitlsp/lsp/doc_handlers.go
floitsch: done.

PR #2101, opened 2024-02-20
Test for completions where the prefix ends with '-'.

floitsch:


tests/lsp/completion-range-compiler-test.toit
```diff
@@ -0,0 +1,81 @@
+// Copyright (C) 2014 Toitware ApS.
```
kasperl, line 1:
```suggestion
// Copyright (C) 2024 Toitware ApS.
```

kasperl (approved):
Looks good to me.

PR #2106, opened 2024-02-26
Increase maximum spread out of memory on MacOS

erikcorry:


kasperl:
I assume this should just be closed, no?

PR #2111, opened 2024-02-28
Replace code accidentally removed from ESP32 mem management

erikcorry:
This just puts exactly the code that was previously there.
There may be room for improvement.


src/third_party/dartino/object_memory.cc
```diff
@@ -231,9 +231,29 @@ void Chunk::find(uword word, const char* name) {
 #endif
 
 Chunk* ObjectMemory::allocate_chunk(Space* owner, uword size) {
+#ifdef TOIT_FREERTOS
```
kasperl, line 234: I think I'd prefer the ifdef to cover the full function, so you end up with two completely separate implementations.

kasperl (approved):
Looks good to me.

PR #2117, opened 2024-03-06
Update system/storage library.

floitsch:


lib/system/storage.toit
```diff
@@ -419,6 +419,9 @@ class RegionReader_ implements Reader:
     from_ = from + n
     return result
 
+  content-size -> int?:
+    return null
```
kasperl, line 423: Maybe add a TODO here? It would be kinda nice to be able to get the content-size for this and we do know it.

kasperl (approved):
Looks good to me.

lib/system/storage.toit
floitsch: done.

PR #2119, opened 2024-03-06
Add 'flush' method and named argument.

floitsch:


lib/io.toit
```diff
@@ -71,32 +75,44 @@ abstract mixin Writer:
...
+    if flush: this.flush
+
+  /**
+  Flushes any buffered data to the underlying resource.
+
+  Often, one can just use the `--flush` flag of the $write, $write-byte or $write-from
+    functions instead.
+  */
+  flush -> none:
+    flush_
```
kasperl, line 115: Why is flush split into two functions?

floitsch: I wanted to have the methods that can be implemented by subclasses next to each other and "protected" (with a trailing "_").

I guess there are two options now:
1. just let people overwrite `flush`.
2. at least call `flush_` from all the other places (instead of `this.flush` in `write-from` ...).

kasperl: I think it is a bit artificial to split it here, but I think both options you provide are sound.

kasperl (approved):
Looks good to me.

PR #2122, opened 2024-03-06
Adapt zlib library.

floitsch:


lib/zlib.toit
```diff
@@ -256,72 +292,136 @@ class ZlibBackend_ implements Backend_:
...
   Writes data to the compressor or decompressor.
   If $wait is false, it may return before all data has been written.
     If it returns zero, then that means the read method must be called
     because the buffers are full.
+
+  Deprecated. Use out.try-write or out.write instead.
   */
   write --wait/bool=true data/io.Data from/int=0 to/int=data.byte-size -> int:
-    if closed-read_: throw "READER_CLOSED"
+    if not wait: return try-write_ data from to
```
erikcorry, line 388: This function no longer throws if the read side was closed?

erikcorry (approved):

lib/zlib.toit
floitsch: The check is in the `try-write_`.

PR #2125, opened 2024-03-06
Fix the consumed count.

floitsch:


lib/io.toit
```diff
@@ -496,7 +496,8 @@ abstract mixin Reader implements old-reader.Reader:
   */
   drain -> none:
     clear
-    while consume_: null // Do nothing.
+    while chunk := consume_:
+      base-consumed_ += chunk.size
```
kasperl, line 500: Why is this called something with base? It feels a bit odd that `consume_` doesn't really consume (at least not in terms of the bookkeeping).

kasperl (approved):
Looks good to me.

PR #2128, opened 2024-03-06
Adapt DNS library.

floitsch:


lib/net/modules/dns.toit
```diff
@@ -640,29 +641,30 @@ class SrvResource extends StringResource:
...
+  parts_ reader packet parts
   return parts.join "."
 
-parts_ packet/ByteArray position/int parts/List [position-block] -> none:
-  while packet[position] != 0:
-    size := packet[position]
+parts_ reader/io.Reader packet/ByteArray parts/List -> none:
+  while true:
+    size := reader.peek-byte
+    if size == 0: break
```
erikcorry, line 656: Why change the code for zero length parts? Now we don't add a zero length string.

lib/net/modules/dns.toit
```diff
@@ -674,58 +676,58 @@ create-dns-packet queries/List records/List -> ByteArray
...
 
   status-bits := is-response ? 0x8000 : 0x0000
   if is-authoritative: status-bits |= 0x0400
 
-  result.write-int16-big-endian id
-  result.write-int16-big-endian status-bits
-  result.write-int16-big-endian queries.size
-  result.write-int16-big-endian records.size
-  result.write-int32-big-endian 0  // Don't support the other things.
+  result.big-endian.write-int16 id
```
erikcorry, line 684: Is it possible to cache result.big-endian in a variable to avoid it being looked up on every single line?

erikcorry (approved):

lib/net/modules/dns.toit
floitsch: as discussed in person. This should still do the same thing.

lib/net/modules/dns.toit
floitsch: done.

PR #2131, opened 2024-03-08
Fix buffer when growing

floitsch:


floitsch:
The test is adapted from the `bytes.Buffer` test.

floitsch:
TBR.

kasperl (commented):
Looks good to me.

PR #2133, opened 2024-03-08
Fix bad dns merge.

floitsch:


floitsch:
TBR.

kasperl (commented):
The naming of the locals is a bit weird now.

Looks good enough.

floitsch:
Will change it in a follow-up PR.

PR #2134, opened 2024-03-11
Replace some of the old readers with io.Readers.

floitsch:


lib/io.toit
```diff
@@ -86,12 +86,12 @@ abstract mixin Writer:
   */
   write data/Data from/int=0 to/int=data.byte-size --flush/bool=false -> int:
     pos := from
-    while not is-closed_:
+    while not is-closed_ and pos < to:
```
kasperl, line 89: Can you explain why you did this change? You went from one to three (pos, to) checks, which just seems super excessive unless there is a good reason for this.

kasperl (approved):
Looks good to me.

lib/io.toit
floitsch: I will change it back once I updated the host package.
I needed to avoid sending empty byte-arrays to the stdout. At the same time I felt like calling flush for empty arrays was still necessary (if `--flush` was set). This is, what I came up with...

Once stdout accepts empty byte-arrays I will revert this code.

floitsch: https://github.com/toitlang/toit/pull/2178

PR #2136, opened 2024-03-11
Make Reader and Writer classes and not mixins.

floitsch:
This helped a lot with the LSP tests.
It's not yet the same performance as before, but much closer.


kasperl (approved):
Looks good to me.

PR #2137, opened 2024-03-12
Rename produced/consumed to 'processed'.

floitsch:


lib/io.toit
```diff
@@ -57,7 +57,7 @@ abstract class Writer:
   is-closed_/bool := false
   endian_/EndianWriter? := null
   byte-cache_/ByteArray? := null
-  consumed_/int := 0
+  processed_/int := 0
```
kasperl, line 60: You omitted the type in the reader. Maybe be consistent about this?

kasperl (approved):
Looks good to me.

lib/io.toit
floitsch: done. (added type to reader).

PR #2139, opened 2024-03-12
Implement content-size for storage.RegionReader.

floitsch:


lib/system/storage.toit
```diff
@@ -403,12 +403,14 @@ class RegionReader_ extends io.Reader:
   from_/int := ?
   to_/int
   max-size_/int
+  content-size/int := ?
```
kasperl, line 406: Immutable.

```suggestion
  content-size/int
```

kasperl (approved):
Looks good to me.

lib/system/storage.toit
floitsch: done.

PR #2140, opened 2024-03-12
Add Reader.skip-up-to.

floitsch:


floitsch:
Not completely sure whether the `skip-up-to` should also consume the delimiter, and whether the result should include the skipped delimiter.

For `read-up-to` we don't return the delimiter (but consume it).

lib/io.toit
```diff
@@ -421,8 +421,46 @@ abstract class Reader implements old-reader.Reader:
...
+  Skips all bytes up to and including the given $delimiter.
+
+  Returns the number of bytes skipped including the $delimiter.
+
+  If $to is given, then the search is limited to the given range.
+  If $throw-if-missing is true and the $delimiter is not in the remaining data, throws.
+  If $throw-if-missing is false and the $delimiter is not in the remaining data, skips
+    all remaining data.
+  */
+  skip-up-to delimiter/int --to/int?=null --throw-if-missing/bool=false -> int:
```
kasperl, line 437: I think it would be easier to remember if this was called `--throw-if-absent` here and in the other places it exists.

tests/io-reader-skip-up-to-test.toit
```diff
@@ -0,0 +1,58 @@
+// Copyright (C) 2024 Toitware ApS.
+// Use of this source code is governed by a Zero-Clause BSD license that can
+// be found in the tests/LICENSE file.
+
+import expect show *
```
kasperl, line 5: Here are a couple of places, where we'd like to use this:

https://github.com/toitware/cellular/blob/main/src/modules/ublox/ublox.toit#L334

https://github.com/toitware/cellular/blob/main/src/base/at.toit#L397

In both cases, we want the delimiter to also be skipped. I think including the delimiter (if found) in the result is fine.

kasperl (approved):
Looks good to me.

lib/io.toit
floitsch: I will change it in a follow-up PR together with the read-up-to.

PR #2147, opened 2024-03-15
Add invoke-size bytecode.

floitsch:
In the LSP server a lot of time was spent in `foo.size` where `foo` was one of our primitive types. This patch adds a bytecode to optimize size accesses.


src/interpreter_core.cc
```diff
@@ -140,6 +140,51 @@ bool Interpreter::fast_at(Process* process, Object* receiver, Object* arg, bool
   return false;
 }
 
+bool Interpreter::fast_size(Process* process, Object* receiver, Smi** result) {
+  if (is_instance(receiver)) {
+    Instance* instance = Instance::cast(receiver);
+    Smi* class_id = instance->class_id();
+    Program* program = process->program();
+    // Note: Assignment in condition.
```
erikcorry, line 148: There's no assignment in the condition.

erikcorry (approved):

PR #2155, opened 2024-03-19
Faster byte array slices

erikcorry:
This is about 20% on top of the faster instance creation.

Perhaps we can do something similar for strings.


floitsch (approved):
LGTM.

kasperl (approved):
Looks good to me.

PR #2156, opened 2024-03-19
Make it easier/possible to change channel for espnow.

floitsch:
We were seeing "Peer channel is not equal to the home channel" errors.

Also see https://esp32.com/viewtopic.php?t=18291


kasperl (approved):
Looks good to me.

PR #2157, opened 2024-03-19
Make uuid more prominent in toitp.

floitsch:
The `--uuid` only prints the UUID, so it can be used in scripts.

### DIFF
--- a/tests/toitp/summary-toitp-test.toit
+++ b/tests/toitp/summary-toitp-test.toit
@@ -8,8 +8,8 @@ import .utils
 main args:
   out := run-toitp args []
   required-output-snippets := [
-    "Uuid: ",
-    "Size: ",
+    "uuid: ",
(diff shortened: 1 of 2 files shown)


tests/toitp/summary-toitp-test.toit
```diff
@@ -8,7 +8,8 @@ import .utils
 main args:
   out := run-toitp args []
   required-output-snippets := [
-    "snapshot: ",
+    "Uuid: ",
```
kasperl, line 11: Why are size and uuid capitalized in the output?

kasperl (approved):
Looks good to me.

tests/toitp/summary-toitp-test.toit
floitsch: changed to lower case.
done.

PR #2159, opened 2024-03-20
Improve performance of reader.skip.

kasperl:


lib/io.toit
```diff
@@ -437,27 +440,29 @@ abstract class Reader implements old-reader.Reader:
   skip-up-to delimiter/int --to/int?=null --throw-if-absent/bool=false -> int:
     skipped := 0
     while true:
-      if not buffered_ or buffered_.size == 0:
+      buffered := buffered_
```
floitsch, line 443: This doesn't work.
`more_` creates the buffered if it doesn't exist yet.

floitsch (approved):
LGTM.

lib/io.toit
kasperl: Fixed.

PR #2162, opened 2024-03-20
Rename 'processed_' field in Reader.

floitsch:
It was too confusing that it didn't include the first buffered array.

### DIFF
--- a/lib/io.toit
+++ b/lib/io.toit
@@ -277,9 +277,12 @@ abstract class Reader implements old-reader.Reader:
   // The position in the first byte array that we got to.
   first-array-position_ := 0
 
-  // The number of bytes in byte arrays that have been used up.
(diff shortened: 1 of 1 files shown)


lib/io.toit
```diff
@@ -280,7 +280,7 @@ abstract class Reader implements old-reader.Reader:
   // The number of bytes in byte arrays that have been used up.
   // Does not yet include the bytes in the first byte array. That is,
   //   the total number that was given to the user is processed_ + first-array-position_.
```
kasperl, line 282: Update comment.

kasperl (approved):
Looks good to me.

lib/io.toit
floitsch: done. Also made it a toitdoc comment so that this can't happen again.

PR #2164, opened 2024-03-20
Move byte-order to io lib.

floitsch:


lib/io/byte-order.toit
```diff
@@ -0,0 +1,477 @@
+// Copyright (C) 2018 Toitware ApS. All rights reserved.
```
floitsch, line 1: Completely unchanged.

kasperl (approved):
Looks good to me.

PR #2167, opened 2024-03-20
Don't get monotonic time on normal GC

erikcorry:
This is > 10% on mandelbrot!


erikcorry:
The measurements were done on Linux desktop.  I don't know why it's so expensive, but a scavenge takes about 1us if almost everything is garbage, so that's the order of magnitude we are dealing with.

src/scheduler.cc
```diff
@@ -449,7 +449,7 @@ bool Scheduler::kill(const Program* program) {
 void Scheduler::gc(Process* process, bool malloc_failed, bool try_hard) {
   bool doing_idle_process_gc = try_hard || malloc_failed || (process && process->system_refused_memory());
   bool doing_cross_process_gc = false;
-  uint64 start = OS::get_monotonic_time();
+  uint64 start = try_hard ? OS::get_monotonic_time() : 0;
```
kasperl, line 452: Add a comment that explains why you do this check.

src/third_party/dartino/two_space_heap.cc
```diff
@@ -124,7 +124,7 @@ void SemiSpace::start_scavenge() {
 GcType TwoSpaceHeap::collect_new_space(bool try_hard) {
   SemiSpace* from = new_space();
 
-  uint64 start = OS::get_monotonic_time();
+  uint64 start = (Flags::tracegc) ? OS::get_monotonic_time() : 0;
```
kasperl, line 127: Add a comment that explains why you do this check.

src/third_party/dartino/two_space_heap.cc
```diff
@@ -308,7 +308,7 @@ void TwoSpaceHeap::validate() {
 
 GcType TwoSpaceHeap::collect_old_space(bool force_compact) {
 
-  uint64 start = OS::get_monotonic_time();
+  uint64 start = (Flags::tracegc) ? OS::get_monotonic_time() : 0;
```
kasperl, line 311: Add a comment that explains why you do this check.

kasperl (approved):
Looks good to me.

PR #2175, opened 2024-03-25
Fix service discovery timeout so it works with tags

erikcorry:
It could be that we want to handle this with retries on the client side instead. Let me know what you think.


erikcorry:
Closed in favour of https://github.com/toitlang/toit/pull/2194

PR #2179, opened 2024-03-27
Fix bad write in pkg manager.

floitsch:
The zlib codec blocks if `write` is used. We need to `try-write` instead.


kasperl (approved):
Looks good to me.

PR #2181, opened 2024-03-27
Start to bring back the 'toit' executable.

floitsch:


src/toit.cc
```diff
@@ -15,375 +15,51 @@
...
-      fprintf(stderr, "Can't have options with '%s'\n", argv[1]);
-      print_usage(1);
-    }
-    print_version();
-  }
-
-  // TODO(2663): remove support for '-r'.
-  if (strcmp(argv[1], "-r") == 0 ||
-      SnapshotBundle::is_bundle_file(argv[1])) {
+  if (argc >= 2 && SnapshotBundle::is_bundle_file(argv[1])) {
```
kasperl, line 38: Do we need all this handling?

tools/toit.toit
```diff
@@ -0,0 +1,204 @@
...
+        cli.Option "sdk-dir"
+            --help="Path to the SDK root."
+            --type="dir"
+            --hidden,
+      ]
+
+  compile-analyze-run-options := [
+    cli.Flag "show-package-warnings"
+        --help="Show warnings from packages.",
+    cli.Flag "Werror" --short-name="Werror"
```
kasperl, line 38: So the short name means you can write `--Werror` and `-Werror`?

tools/toit.toit
```diff
@@ -0,0 +1,204 @@
...
+        --help="Path to the project root. Any package.lock file must be in that folder."
+        --type="dir",
+    cli.Option "X" --short-name="X"
+        --help="Provide a compiler flag."
+        --hidden
+        --multi,
+  ]
+
+  compile-run-options := [
+    cli.OptionInt "optimization" --short-name="O"
```
kasperl, line 50:
```suggestion
    cli.OptionInt "optimization-level" --short-name="O"
```

kasperl (approved):
Looks good to me.

src/toit.cc
floitsch: It's to support fast "run with a snapshot".
For all other cases we need to invoke the compiler anyway, so it feels like running a Toit program first doesn't hurt as much.
We can add more fast cases, if we feel like going through the Toit code adds too much overhead.

tools/toit.toit
floitsch: correct.

tools/toit.toit
floitsch: done.

PR #2183, opened 2024-03-28
Don't print other errors if SDK doesn't match.

floitsch:


src/compiler/compiler.cc
```diff
@@ -1799,9 +1801,11 @@ Pipeline::Result Pipeline::run(List<const char*> source_paths, bool propagate) {
   auto package_lock = load_package_lock(source_paths);
 
   if (package_lock.sdk_constraint() != "") {
-    // TODO(florian): we should be able to continue compiling even with
-    // a wrong SDK.
-    check_sdk(package_lock.sdk_constraint(), diagnostics());
+    bool succeeded = check_sdk(package_lock.sdk_constraint(), diagnostics());
+    if (!succeeded && !configuration_.force && configuration_.lsp == null) {
+      diagnostics()->report_error("Compilation failed.");
```
kasperl, line 1806: Do you typically terminate these errors with .? The one in line 1622 doesn't have one.

kasperl (approved):
Looks good to me.

src/compiler/compiler.cc
floitsch: There are a few other 'report_X' with dots.
I will clean this up in a follow-up PR.

PR #2185, opened 2024-04-02
Don't end errors with "." in compiler.

floitsch:


kasperl (approved):
Looks good to me. I don't mind the terminating dots, but I prefer consistency.

PR #2188, opened 2024-04-03
Add toitp as 'snapshot' subcommand to toit executable.

floitsch:


kasperl (approved):
Looks good to me.

I think `toit firmware` and `toit assets` fit nicely under this scheme too.

floitsch:
> Looks good to me.
> 
> I think `toit firmware` and `toit assets` fit nicely under this scheme too.

I wonder if it should be under `toit esp32 firmware` or something similar, to make it a bit distinct from host operations. But yes: pretty much every tool should be integrated here.

PR #2194, opened 2024-04-03
Don't use tasks to wait for services to appear

erikcorry:


lib/system/services.toit
```diff
@@ -163,18 +172,54 @@ class ServiceClient:
...
+      // service we want has started up.
+      error := catch:
+        with-timeout timeout:
+          while true:
+            discovered = channel.receive
+            result := find-service_ discovered
+            if result: return result
+      if error == "DEADLINE_EXCEEDED":
+        return if-absent.call
+      throw error
```
kasperl, line 217:
```suggestion
      if timeout: 
        // We got back a proxy for a resource, which will notify us when the
        // service we want has started up.
        catch --unwind=(: it != DEADLINE_EXCEEDED_ERROR):
          with-timeout timeout:
            while true:
              discovered = channel.receive
              result := find-service_ discovered
              if result: return result
      return if-absent.call
```

kasperl (approved):
Looks good to me.

Added a suggestion...

PR #2195, opened 2024-04-03
Force the version number.

floitsch:
This way we can start to release packages against the new version number and test it with the produced executable (and more importantly the buildbot).


kasperl (approved):
Looks good to me.

PR #2204, opened 2024-04-05
Improve external-wakeup doc.

floitsch:


lib/esp32/esp32.toit
```diff
@@ -60,10 +60,20 @@ total-deep-sleep-time -> int:
   #primitive.esp32.total-deep-sleep-time
 
 /**
-Sets the ESP32 to wake up from deep sleep if the GPIO pins in pin_mask matches the mode.
+Sets the ESP32 to wake up from deep sleep if the GPIO pins in $pin_mask
```
kasperl, line 63:
```suggestion
Sets the ESP32 to wake up from deep sleep if the GPIO pins in $pin-mask
```

kasperl (approved):
Looks good to me.

PR #2205, opened 2024-04-08
Add pkg-dns-simple-server and fix bug in reader.skip.

floitsch:


lib/io/reader.toit
```diff
@@ -158,6 +158,7 @@ abstract class Reader implements old-reader.Reader:
     buffered := buffered_
     if not buffered:
       if not more_: throw UNEXPECTED-END-OF-READER
+      buffered = buffered_
```
kasperl, line 161: Maybe copy the comment from line 197?

```suggestion
      // The call to $more_ may have changed $buffered_ if
      // it was null. Read the field again.
      buffered = buffered_
```

kasperl (approved):
Looks good to me.

PR #2208, opened 2024-04-09
Make it possible to stream JSON and YAML output onto a writer.

erikcorry:


lib/encoding/json-like-encoder_.toit
```diff
@@ -42,65 +47,23 @@ abstract class EncoderBase_ extends Buffer_:
...
-    new := ByteArray new-size
-    new.replace 0 buffer_
-    buffer_ = new
-
   /**
   Outputs a string or ByteArray directly to the JSON stream.
   No quoting, no escaping.  This is mainly used for things
     that will be parsed as numbers or strings by the receiver.
   */
   put-unquoted data -> none:
```
floitsch, line 65: ?
```suggestion
  put-unquoted data/io.Data -> none:
```

lib/encoding/json-like-encoder_.toit
```diff
@@ -42,65 +47,23 @@ abstract class EncoderBase_ extends Buffer_:
   abstract put-list size/int [generator] [converter]
 
   put-unicode-escape_ code-point/int:
```
floitsch, line 49: Store the writer in a local.

lib/encoding/json.toit
```diff
@@ -40,6 +41,37 @@ Utf-8 encoding is used for strings.
...
+  be serialized instead of the object that was passed in.  Alternatively,
+  the $converter block can call the $Encoder.encode, $Encoder.put-list,
+  or Encoder.put_unquoted methods on the encoder.
+Utf-8 encoding is used on the writer.
+*/
+encode-stream --writer/io.Writer obj [converter] -> none:
+  e := Encoder writer
+  e.encode obj converter
+
+encode-stream --writer/io.Writer obj converter/Lambda -> none:
```
floitsch, line 62:
```suggestion
/**
Variant of $(encode-stream --writer obj).

Takes a lambda as converter.

The $converter is called when the encoder encounters an unsupported value. It
  is called with two arguments: the unsupported object and an object of type $Encoder.
  If the converter returns a non-null value ...
*/
encode-stream --writer/io.Writer obj converter/Lambda -> none:
```

The block version can then be a variant of the lambda version (or inverse).

lib/encoding/json.toit
```diff
@@ -40,6 +41,37 @@ Utf-8 encoding is used for strings.
 encode obj -> ByteArray:
   return encode obj: throw "INVALID_JSON_OBJECT"
 
+/**
```
floitsch, line 44: This documentation should be on `encode-stream o --writer`.
The version with a block|lambda converter should be variants of the one without.

lib/encoding/json.toit
```diff
@@ -100,53 +133,70 @@ decode-stream reader:
   d := StreamingDecoder
   return d.decode-stream reader
 
-
+/**
+Deprecated.  Use the top level json.encode or json.stringify functions
+  instead.
```
floitsch, line 138: Given that we pass the encoder to the converter, we might not want to deprecate it yet.

lib/encoding/json.toit
```diff
@@ -100,53 +133,70 @@ decode-stream reader:
...
-    put-unquoted "false"
+    writer_.write "false"
 
   encode-null_:
-    put-unquoted "null"
+    writer_.write "null"
 
   encode-map_ map [converter]:
-    put-byte_ '{'
+    writer_.write-byte '{'
```
floitsch, line 187: Store the writer in a local.

lib/encoding/yaml/encoder.toit
```diff
@@ -4,26 +4,50 @@
...
+
+  /** See $EncoderBase_.put-unquoted */
   // TODO(florian): Remove when toitdoc compile understands inherited methods
   put-unquoted data -> none:
     super data
 
   put-value_ val/string:
-    if enclosed_in_map_: put-byte_ ' '
-    put-unquoted val
+    if enclosed_in_map_: writer_.write-byte ' '
```
floitsch, line 49: Store writer_ in local?

floitsch (approved):

floitsch (approved):
LGTM.

lib/io/buffer.toit
```diff
@@ -22,6 +22,9 @@ class Buffer extends CloseableWriter:
   buffer_/ByteArray := ?
   is-growable_/bool := ?
 
+  processed -> int:
+    return offset_
```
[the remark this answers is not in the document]
erikcorry: We don't update the processed_ field at all, it's not just when clearing.

lib/io/buffer.toit
```diff
@@ -22,6 +22,9 @@ class Buffer extends CloseableWriter:
   buffer_/ByteArray := ?
   is-growable_/bool := ?
 
+  processed -> int:
```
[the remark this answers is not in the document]
erikcorry: "Amount" is more suitable for amorphous (non-countable) things, whereas "number" is more used for things that can be counted.  Like less and fewer.

lib/encoding/yaml/encoder.toit
```diff
@@ -61,17 +85,17 @@ class YamlEncoder extends EncoderBase_:
     // In yaml, a null value is written as an empty string.
 
   put-new-line:
-    put-byte_ '\n'
-    current-line-start-offset_ = offset_
+    writer_.write-byte '\n'
```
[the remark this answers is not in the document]
erikcorry: I did this for cases where it was used more than twice.

lib/io/buffer.toit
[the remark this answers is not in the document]
floitsch: You don't need to update the `processed_` field otherwise, as the base class is doing it for you.
But by clearing, the buffer class is somehow influencing the "processed" number. You could argue that it shouldn't reset, but it's probably better that it does.
-> If we changed the `processed_` to 0 in the super class, then things should work.
Maybe that's actually better, since the base class could then, in theory, use that field for other things.

lib/encoding/json.toit
```diff
@@ -52,14 +52,14 @@ The $converter block is passed an object to be serialized and an instance
   of the $Encoder class.  If it returns a non-null value, that value will
   be serialized instead of the object that was passed in.  Alternatively,
   the $converter block can call the $Encoder.encode, $Encoder.put-list,
-  or Encoder.put-unquoted methods on the encoder.
+  or $Encoder.put-unquoted methods on the encoder.
 */
 encode-stream --writer/io.Writer obj [converter] -> none:
```
floitsch, line 57: Could be done in a separate PR or here, but just `encode o --writer` and `decode --reader` would actually be nicer than the `-stream` suffix.
Would require to deprecate the existing `decode-stream` method, so maybe a follow-up PR is better.

floitsch (approved):
Still LGTM.

PR #2211, opened 2024-04-10
Remove distinction between update-gold and update-health-gold.

erikcorry:
There's not really any time where you only want to update one set of gold files, leaving the other un-updated.


floitsch (approved):
'make update-gold' was initially primarily used for negative tests, where you wouldn't need to update everything.

maybe:
- update-negative-gold
- update-type-gold
and
`update-gold` depends on these?

PR #2215, opened 2024-04-11
Remove misleading code.

floitsch:
The SDK isn't filtered out in the beginning, but during the export.


kasperl (approved):
Looks good to me.

PR #2218, opened 2024-04-12
Store the "mode" (package vs lib) and sdk-path in json.

floitsch:
The pkg server (pkg.toit.io) currently sets this information through a header. That's inconvenient and unnecessary.


tools/toitlsp/cmd/toitdoc.go
```diff
@@ -55,6 +55,7 @@ func Toitdoc(sdkVersion string) *cobra.Command {
 	cmd.Flags().Bool("exclude-sdk", false, "if set, will remove the sdk libraries from the toitdoc")
 	cmd.Flags().Bool("exclude-pkgs", true, "if set, will remove other packages from the toitdoc")
 	cmd.Flags().Bool("include-private", false, "if set, will include private toitdoc for private elements")
+	cmd.Flags().String("pkg-name", "", "the name of the package. If not set, assumes that the core libs are generated.")
```
kasperl, line 58:
```suggestion
	cmd.Flags().String("pkg-name", "", "the name of the package. If not set, the core libs will be generated")
```

kasperl (approved):
Looks good to me.

PR #2222, opened 2024-04-14
Fix program-path for 'toit'.

floitsch:
When running the toit snapshot (toit.toit) we want the program path to be set to the executable 'toit'. Otherwise we can't find the bundled programs, like `toit.compile`, `toit.pkg`, etc.

Currently `toit run` shells out to the `toit.run` executable which will set the program-path correctly for the executable it is running. If we optimize this by running directly then we will need to update the program_path.


kasperl (approved):
Looks good to me.

PR #2223, opened 2024-04-14
Fix yaml text starting with "Q".

floitsch:
Yaml currently reserves '@' for beginnings of blocks, as in:
```
foo: @
```
However, we accidentally reserved "Q".

Any block that started with "Q" was marked as invalid. Example:
```
description: Qualification of the author.
```

### DIFF
--- a/lib/encoding/yaml/parser.toit
+++ b/lib/encoding/yaml/parser.toit
@@ -34,8 +34,8 @@ C-MAPPING-VALUE_    ::= ':'
 C-ALIAS_            ::= '*'
 C-SINGLE-QUOTE_     ::= '\''
 C-DOUBLE-QUOTE_     ::= '"'
-C-RESERVCED-1_      ::= '@'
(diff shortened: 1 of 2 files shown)


lib/encoding/yaml/parser.toit
```diff
@@ -34,7 +34,7 @@ C-MAPPING-VALUE_    ::= ':'
 C-ALIAS_            ::= '*'
 C-SINGLE-QUOTE_     ::= '\''
 C-DOUBLE-QUOTE_     ::= '"'
-C-RESERVCED-1_      ::= 'Q'
+C-RESERVCED-1_      ::= '@'
 C-RESERVCED-2_      ::= '`'
```
kasperl, line 38:
```suggestion
C-RESERVED-1_      ::= '@'
C-RESERVED-2_      ::= '`'
```

kasperl (approved):
Looks good to me.

PR #2226, opened 2024-04-15
Run 64-bit debug tests when pushing to master.

floitsch:
It's a bit heavy to do them all the time, but just for the releases should be fine.


kasperl (commented):
I feel like this is a little bit too late. When this runs, we've already tagged the release and pinged Discord. Can we turn this into a manual build bot action that we can run before we create a release?

kasperl (approved):
Looks good to me.

PR #2229, opened 2024-04-18
Use a C++-level buffer for outgoing TLS data.

erikcorry:


src/resources/tls.cc
```diff
@@ -482,24 +475,16 @@ Object* tls_error(BaseMbedTlsSocket* socket, Process* process, int err) {
   return Primitive::mark_as_error(str);
 }
 
-PRIMITIVE(get_outgoing_fullness) {
+PRIMITIVE(get_outgoing) {
```
kasperl, line 478: take_outgoing?

lib/tls/session.toit
```diff
@@ -470,43 +468,34 @@ class Session:
...
+    for scan := 0; scan < outgoing-data.size; :
+      if bytes-before-next-record-header_ > 0:
+        skip := min
+            bytes-before-next-record-header_
+            outgoing-data.size - scan
+        scan += skip
+        bytes-before-next-record-header_ -= skip
+      else:
+        addition := min
+            (RECORD-HEADER-SIZE_ - outgoing-partial-header_.size)
```
kasperl, line 484: You don't need the parentheses on these two lines with arguments for `min`.

lib/tls/session.toit
```diff
@@ -470,43 +468,34 @@ class Session:
...
-          // We have the start of the next record available.
-          if fullness - from + RECORD-HEADER-SIZE_ >= bytes-before-next-record-header_:
-            // We have the full record header available.
-            header := RecordHeader_ outgoing-buffer_[from + bytes-before-next-record-header_..]
-            record-size := header.length
+    // Get the outgoing data from the buffer, freeing up space for more data.
+    outgoing-data := tls-get-outgoing_ tls_
+
+    // Scan the outgoing buffer for record headers.
+    for scan := 0; scan < outgoing-data.size; :
```
kasperl, line 475: I think I'd prefer this as:

```
scan := 0
while scan < outgoing-data.size:
  ...
```

I might even want to cache `outgoing-data.size` in a local. It is used in a few places and it doesn't change.

lib/tls/session.toit
```diff
@@ -470,43 +468,34 @@ class Session:
...
+      if bytes-before-next-record-header_ > 0:
+        skip := min
+            bytes-before-next-record-header_
+            outgoing-data.size - scan
```
kasperl, line 479: You could cache this in a local. It is used on both paths.

lib/tls/session.toit
```diff
@@ -470,43 +468,34 @@ class Session:
...
+        scan += skip
+        bytes-before-next-record-header_ -= skip
+      else:
+        addition := min
+            (RECORD-HEADER-SIZE_ - outgoing-partial-header_.size)
+            (outgoing-data.size - scan)
+        if addition != 0:
+          outgoing-partial-header_ += outgoing-data[scan .. scan + addition]
```
kasperl, line 487: I don't like reading the `outgoing-partial-header_` field over and over again. Maybe cache it too? It doesn't change here.

lib/tls/session.toit
```diff
@@ -470,43 +468,34 @@ class Session:
...
-        if outgoing-buffer_.is-empty:
-          outgoing-buffer_ = ByteArray_.external_ 1500
-        // Be sure not to lose the pending bytes.  Instead put them in the
-        // otherwise empty outgoing_buffer_.
-        outgoing-buffer_.replace 0 pending-bytes
-        tls-set-outgoing_ tls_ outgoing-buffer_ pending-bytes.size
-        return
+            bytes-before-next-record-header_ = header.length
+            outgoing-partial-header_ = #[]
+    writer_.write outgoing-data
```
kasperl, line 498: We don't need to skip any bytes from `outgoing-data`? I think that deserves a comment.

kasperl (approved):
Looks good to me.

PR #2230, opened 2024-04-18
Use the 'name' field in the package.lock file.

floitsch:
When having an `import foo` use the name of the package to find the correct library in the import.

This is only relevant if the prefix of the import was changed.


kasperl (approved):
Looks good to me.

PR #2231, opened 2024-04-18
Fix name-usage with dash-identifiers.

floitsch:
The 'add-segment' function first tries the given segment string, and then the underscore version of it. If the name is underscore, but the file is dash-based, then we didn't find the library as it never tried the dash-version.


kasperl (approved):
Even better.

PR #2234, opened 2024-04-18
Add 'reader.do --lines [block]'.

floitsch:


lib/io/reader.toit
```diff
@@ -592,6 +592,16 @@ abstract class Reader implements old-reader.Reader:
       result.add line
     return result
 
+  /**
+  Calls the given $block for each remaining line.
+
+  The $lines argument must be true.
+  See $read-line.
+  */
+  do --lines/bool --keep-newlines/bool=false [block] -> none:
```
kasperl, line 601: Check that `lines` is true.

kasperl (approved):
Looks good to me.

lib/io/reader.toit
floitsch: done.

PR #2241, opened 2024-04-21
Fix implicit yaml key.

floitsch:


lib/encoding/yaml/parser.toit
```diff
@@ -1016,7 +1016,7 @@ class Parser_ extends PegParserBase_:
     return null
 
   ns-plain-safe c/int -> int?:
-    if c == FLOW-OUT_ or c == BLOCK-OUT_: return ns-plain-safe-out
+    if c == FLOW-OUT_ or c == BLOCK-KEY_: return ns-plain-safe-out
```
floitsch, line 1019: https://yaml.org/spec/1.2.2/#rule-ns-plain-safe

kasperl (approved):
Looks good to me.

PR #2245, opened 2024-04-22
Experiment with defragging memory layout on heap compaction.

erikcorry:


erikcorry:
This probably doesn't do much on embedded, because when we really need it there is not a spare page.  We should reuse the spare new-space page for this purpose.

erikcorry:
I could not find a case where this helped.

PR #2249, opened 2024-04-23
Add 'handle' support.

floitsch:


src/resources/ble_esp32.cc
```diff
@@ -1820,6 +1820,12 @@ PRIMITIVE(write_value) {
   return Smi::from(with_response ? 1 : 0);
 }
 
+PRIMITIVE(handle) {
+  ARGS(Resource, resource)
+  auto element = reinterpret_cast<BleReadWriteElement*>(resource);
```
floitsch, line 1825: The reinterpret_cast is horrible, but that's the pattern that is used throughout this file.
Needs to get fixed separately.

src/resources/ble_darwin.mm
```diff
@@ -1207,6 +1207,10 @@ _new BleRemoteDeviceResource(
   UNIMPLEMENTED();
 }
 
+PRIMITIVE(handle) {
+  UNIMPLEMENTED();
```
kasperl, line 1211: Maybe these primitives should be using `FAIL(UNIMPLEMENTED)`, so we don't crash the VM for no good reason?

kasperl (approved):
Looks good to me.

src/resources/ble_darwin.mm
floitsch: I will do this in a follow-up PR.

PR #2253, opened 2024-04-24
Use a global BLE lock.

floitsch:


src/resources/ble_esp32.cc
```diff
@@ -1382,19 +1398,17 @@ PRIMITIVE(create_peripheral_manager) {
...
+    ble_gatts_reset();
+    BleEventSource::instance()->on_event(peripheral_manager, kBleStarted);
   }
 
   return proxy;
 }
 
 PRIMITIVE(close) {
   ARGS(BleResourceGroup, group)
+  // The group is the only primitive that doesn't allocate a locker.
```
kasperl, line 1411:
```suggestion
  // The close primitive is the only primitive that doesn't allocate a locker.
```

kasperl (approved):
Looks good to me.

src/resources/ble_esp32.cc
```diff
@@ -219,16 +248,10 @@ class BleReadWriteElement : public BleErrorCapableResource {
   }
 
   os_mbuf* mbuf_to_send() {
```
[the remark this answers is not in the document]
floitsch: I don't think it's necessary. There was already the `mbuf_to_send` that just took the resource-group mutex.
And it's just important that there is a lock.

src/resources/ble_esp32.cc

PR #2255, opened 2024-04-24
Call the GC from the BLE thread.

floitsch:
When memory is tight just call the GC from the BLE thread.


floitsch:
The test is, unfortunately, still somewhat flaky.
It is a bit better if I remove the parallel task, but even then not completely reliable.

src/resources/ble_esp32.cc
```diff
@@ -880,12 +901,14 @@ void
 BleServiceResource::_on_characteristic_discovered(const BleCallbackScope& scope, const struct ble_gatt_error* error, const struct ble_gatt_chr* chr) {
   switch (error->status) {
     case 0: {
-      auto ble_characteristic =
-          get_or_create_characteristics_resource(
-              chr->uuid, chr->properties, chr->def_handle,
-              chr->val_handle);
-      if (!ble_characteristic) {
-        set_malloc_error(true);
+      for (int i = 0; i < 2; i++) {
```
kasperl, line 904: It is weird that this uses a different pattern (looping and retrying here). Why can't `get_or_create_characteristics_resource` do the retry?

If the first attempt fails and we're going to try again on the second iteration of the loop, is it really correct to call `set_malloc_error(true)` and force an OOM?

src/resources/ble_esp32.cc
```diff
@@ -935,9 +958,16 @@ void BleCentralManagerResource::_on_discovery(const BleCallbackScope& scope, ble
       if (event->disc.length_data > 0) {
```
kasperl, line 958: Maybe put `event->disc.length_data` in a local?

src/resources/ble_esp32.cc
```diff
@@ -901,14 +902,12 @@ void
...
-        auto ble_characteristic =
-            get_or_create_characteristics_resource(
-                chr->uuid, chr->properties, chr->def_handle,
-                chr->val_handle);
-        if (!ble_characteristic) {
-          set_malloc_error(true);
-        }
+      auto ble_characteristic =
+          get_or_create_characteristics_resource(&scope,
+                                                  chr->uuid, chr->properties, chr->def_handle,
```
kasperl, line 907: Isn't the indentation slightly weird here?

kasperl (approved):
Even better!

src/resources/ble_esp32.cc

PR #2257, opened 2024-04-25
Clean tags.h.

floitsch:
There was no need to do special TLS class handling, and the LeakyDirectory has been removed a while ago.


kasperl (approved):
Looks good to me.

PR #2258, opened 2024-04-25
Add 'write --flush' to BLE.

floitsch:


lib/ble.toit
```diff
@@ -278,13 +278,18 @@ class RemoteCharacteristic extends RemoteReadWriteElement_ implements Attribute:
 
   /**
   Writes the value of the characteristic on the remote device.
+
+  If $flush is true, waits until the data has been written. This is the default for
```
kasperl, line 282: Flush seems to just default to false all the way down to the primitive. It seems quite convoluted. Is it really necessary?

src/resources/ble_esp32.cc
```diff
@@ -1889,7 +1889,7 @@ PRIMITIVE(get_value) {
 }
 
 PRIMITIVE(write_value) {
-  ARGS(BleReadWriteElement, element, Object, value, bool, with_response)
+  ARGS(BleReadWriteElement, element, Object, value, bool, with_response, bool, flush)
```
kasperl, line 1892: This feels weird. Why do you take two bool arguments and only act on the logical or of them?

lib/ble.toit
```diff
@@ -278,13 +278,18 @@ class RemoteCharacteristic extends RemoteReadWriteElement_ implements Attribute:
...
+    characteristics that require a response.
   */
-  write value/ByteArray -> none:
+  write value/ByteArray --flush/bool?=null -> none:
     if (properties & (CHARACTERISTIC-PROPERTY-WRITE
                       | CHARACTERISTIC-PROPERTY-WRITE-WITHOUT-RESPONSE)) == 0:
       throw "Characteristic does not support write"
 
-    write_ value --expects-response=(properties & CHARACTERISTIC-PROPERTY-WRITE) != 0
+    if not flush: flush = false
```
kasperl, line 290: This feels like a bad smell too. Why can't flush just have a proper default value (false)?

kasperl (approved):
Looks good to me, but I don't quite understand why you implemented it the way you did. Maybe there is something subtle that I missed?

PR #2260, opened 2024-04-25
Throw when trying to send data that doesn't fit into the MTU.

floitsch:


src/resources/ble_esp32.cc
```diff
@@ -868,6 +874,17 @@ BleCharacteristicResource::~BleCharacteristicResource() {
   service_->remove_characteristic(this);
 }
 
+uint16 BleCharacteristicResource::get_mtu() {
+  int min_sub_mtu = -1;
```
floitsch, line 878: Code that was copied from the `get_att_mtu` primitive.
Tbh, I don't really understand it.

src/resources/ble_esp32.cc
```diff
@@ -1343,24 +1360,28 @@ int BlePeripheralManagerResource::_on_gap(const BleCallbackScope& scope, struct
   return BLE_ERR_SUCCESS;
 }
 
-static Object* object_to_mbuf(Process* process, Object* object, os_mbuf** result) {
```
floitsch, line 1346: Split into two functions on the right. One that takes an object and one that takes a blob.

src/resources/ble_esp32.cc
```diff
@@ -1894,14 +1915,23 @@ PRIMITIVE(get_value) {
...
 
 PRIMITIVE(write_value) {
-  ARGS(BleReadWriteElement, element, Object, value, bool, with_response, bool, flush, bool, allow_retry)
+  ARGS(BleReadWriteElement, element, Blob, value, bool, with_response, bool, flush, bool, allow_retry)
 
   Locker locker(BleResourceGroup::instance()->mutex());
 
   if (!element->service()->device()) FAIL(INVALID_ARGUMENT);
 
+  uint16 mtu;
```
kasperl, line 1924: Not sure it is really beneficial to have this as a 16 bit local, but it might be the style around here.

kasperl (commented):
Looks good to me.

kasperl (approved):

src/resources/ble_esp32.cc

PR #2262, opened 2024-04-27
Keep track of how many BLE devices exist.

floitsch:
We will use this to correctly deallocate the central when all devices have disappeared.

### DIFF
--- a/src/resources/ble_esp32.cc
+++ b/src/resources/ble_esp32.cc
@@ -512,7 +512,8 @@ class BleCentralManagerResource : public BleErrorCapableResource {
 
   explicit BleCentralManagerResource(BleResourceGroup* group, BleAdapterResource* adapter)
       : BleErrorCapableResource(group, CENTRAL_MANAGER)
-      , adapter_(adapter) {}
+      , adapter_(adapter)
+      , device_count_(0) {}
 
   ~BleCentralManagerResource() override;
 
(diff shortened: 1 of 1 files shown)


src/resources/ble_esp32.cc
```diff
@@ -532,9 +533,18 @@ class BleCentralManagerResource : public BleErrorCapableResource {
...
+  }
+
+  void decrease_device_count() {
+    device_count_--;
+  }
+
  private:
   void _on_discovery(const BleCallbackScope& scope, ble_gap_event* event);
   BleAdapterResource* adapter_;
+  int device_count_;
```
kasperl, line 547: Maybe just initialize it here?

src/resources/ble_esp32.cc
```diff
@@ -608,11 +618,18 @@ class BlePeripheralManagerResource : public ServiceContainer<BlePeripheralManage
 class BleRemoteDeviceResource : public ServiceContainer<BleRemoteDeviceResource> {
  public:
   TAG(BleRemoteDeviceResource);
-  explicit BleRemoteDeviceResource(BleResourceGroup* group, bool secure_connection)
-    : ServiceContainer(group, REMOTE_DEVICE)
-    , handle_(kInvalidHandle)
-    , secure_connection_(secure_connection)
-    , connected_(false) {}
+  explicit BleRemoteDeviceResource(BleResourceGroup* group, BleCentralManagerResource* central_manager, bool secure_connection)
```
kasperl, line 621: No need for explicit, but it's fine to keep it.

kasperl (approved):
Looks good to me.

src/resources/ble_esp32.cc

src/resources/ble_esp32.cc
floitsch: Keeping it, as it doesn't hurt.

PR #2266, opened 2024-04-29
Use word instead of int.

erikcorry:
Anything that relates to the size of an array, an offset to a pointer, or the size of memory should be a word, not an int.  This avoids 64+32 add operations.

Anything extracted from a Smi should be word because int will silently overflow.

Many primitives changed from int (the smaller of int and Smi) to word (which actually means the argument must be Smi).


kasperl (approved):
Looks good to me.

PR #2272, opened 2024-04-30
Use more io.Data in the ble library.

floitsch:


lib/ble.toit
```diff
@@ -132,7 +133,7 @@ class AdvertisementData:
     if name: size += 2 + name.size
     service-classes.do: | uuid/BleUuid |
       size += 2 + uuid.to-byte-array.size
-    if not manufacturer-data.is-empty: size += 2 + manufacturer-data.size
+    if manufacturer-data.byte-size != 0: size += 2 + manufacturer-data.byte-size
```
kasperl, line 136: Not sure why I prefer this.

```suggestion
    if manufacturer-data.byte-size > 0: size += 2 + manufacturer-data.byte-size
```

kasperl (approved):
Nice!

PR #2273, opened 2024-04-30
Store NimBLE service-definition in adapter.

floitsch:
We are not allowed to free the data as long as the GATT server is up and running. Since the server is shut down by the adapter, we now store the data in the adapter and free it only once the NimBLE thread has been stopped.


kasperl (approved):
Looks good to me.

PR #2275, opened 2024-05-01
Remove some 'close' methods in the BLE code.

floitsch:
The underlying system can't just close services or characteristics. It doesn't make sense to give these 'close' methods to users if it doesn't do anything.


lib/ble.toit
```diff
@@ -247,12 +247,12 @@ class RemoteCharacteristic extends RemoteReadWriteElement_ implements Attribute:
   Automatically removes the characteristic from the list of discovered characteristics of
     the service.
   */
-  close:
+  close_:
```
kasperl, line 250:
```suggestion
  close_ -> none:
```

kasperl (approved):
Looks good to me.

PR #2279, opened 2024-05-02
Fix RMT on esp32s3 (and other variants).

floitsch:


lib/rmt.toit
```diff
@@ -232,6 +232,10 @@ class Channel:
   /**
   Constructs a channel using the given $num using the given $pin.
 
+  Note: only the ESP32 and the ESP32S2 support configuring the channel direction at a later
+    time. For all other platforms this constructor will give a TX channel, unless
```
kasperl, line 236:
```suggestion
    time. For all other platforms, this constructor will give a TX channel, unless
```

src/resources/rmt_esp32.cc
```diff
@@ -31,17 +31,74 @@ namespace toit {
...
 #ifndef SOC_RMT_CHANNELS_PER_GROUP
 #error "SOC_RMT_CHANNELS_PER_GROUP not defined"
 #endif
 
-ResourcePool<rmt_channel_t, kInvalidChannel> rmt_channels(
-    RMT_CHANNEL_0, RMT_CHANNEL_1, RMT_CHANNEL_2, RMT_CHANNEL_3
+#if SOC_RMT_CHANNELS_PER_GROUP == SOC_RMT_TX_CANDIDATES_PER_GROUP
+// All channels are TX channels. We assume that they also are RX channels.
+// This is the setup on the ESP32 and ESP32S2 chips.
+ResourcePool<rmt_channel_t, kInvalidChannel> rmt_channels_(
```
kasperl, line 49: Should these be static (pools and pointers)?

kasperl (approved):
Looks good to me.

lib/rmt.toit
floitsch: done.

src/resources/rmt_esp32.cc
floitsch: done.

PR #2282, opened 2024-05-02
Fix add_characteristic primitive.

floitsch:
It was taking one too many parameters which made it flaky.


kasperl (approved):
Looks good to me.

PR #2284, opened 2024-05-02
Fix more resource management

floitsch:
We still had use-after-free cases.

### DIFF
--- a/src/resources/ble_esp32.cc
+++ b/src/resources/ble_esp32.cc
@@ -79,7 +79,8 @@ class DiscoveredPeripheral : public DiscoveredPeripheralList::Element {
   uint8 event_type_;
 };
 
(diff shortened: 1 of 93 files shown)


src/resources/ble_esp32.cc
```diff
@@ -294,13 +295,15 @@ class BleCallbackResource : public BleResource {
 
   void* token() const {
     if (token_ == kInvalidToken) FATAL("Ble Resource token wasn't set");
+    if (token_ == kDeletedToken) FATAL("Ble Resource token taken for deleted object");
```
kasperl, line 298: I think I'd prefer "BLE resource token ..." or "BleResource token ..." here and in other places.

kasperl (approved):
Looks good to me.

src/resources/ble_esp32.cc

PR #2285, opened 2024-05-03
Allow to handle write requests.

floitsch:


lib/ble.toit
```diff
@@ -778,19 +778,68 @@ class LocalCharacteristic extends LocalReadWriteElement_ implements Attribute:
...
+            // Get the received value, and call the block with it.
+            received-value := ble-get-value_ resource_
+            // Typically, we only get a 'null' here if it was a data-received-event.
+            // TODO(florian): why is this possible?
+            if received-value:
+              block.call received-value
+        finally:
+          // Always reply.
+          // If no value was set we store null.
+          // TODO(florian): would be nice to call "CANCELLED" if the block throws.
```
kasperl, line 839:
```suggestion
          // TODO(florian): would be nice to call "CANCELED" if the block throws.
```

lib/ble.toit
```diff
@@ -778,19 +778,68 @@ class LocalCharacteristic extends LocalReadWriteElement_ implements Attribute:
...
+          resource-state_.clear-state event
```
kasperl, line 840: If the current task is canceled or if you're out of time, you will not deliver a reply. Maybe use `critical-do` to avoid this?

kasperl (approved):
Looks good to me.

lib/ble.toit
floitsch: changed the comment slightly, but
done.

lib/ble.toit
floitsch: I think the clearing of the state is more important than the missed callback-reply (since the callback has already a timeout).
done.

PR #2287, opened 2024-05-03
Split ble into local and remote.

floitsch:


lib/ble/ble.toit
```diff
@@ -0,0 +1,503 @@
...
+  #primitive.ble.set-preferred-mtu
+
+ble-get-error_ characteristic is-oom:
+  #primitive.ble.get-error
+
+ble-clear-error_ characteristic is-oom:
+  #primitive.ble.clear-error
+
+ble-platform-requires-uuid-as-byte-array_:
+  return system.platform == system.PLATFORM-FREERTOS
```
floitsch, line 474: Only change: went from `platform` to `system.platform` as I would otherwise need an `import system show platform`, which would then be exported due to the `export *`.

kasperl (approved):
Looks good to me.

PR #2303, opened 2024-05-15
Switch to uint8_t* for data in the C service.

floitsch:


include/toit/toit.h
```diff
@@ -4,10 +4,13 @@
 
 #pragma once
 
+
```
kasperl, line 7: Why another one?

kasperl (approved):
Looks good to me.

include/toit/toit.h
floitsch: done.

PR #2307, opened 2024-05-16
Don't use w5500 structs if they aren't enabled

floitsch:
The eth_w5500_config_t struct is only available if the sdkconfig has support for w5500 enabled.


kasperl (approved):
Looks good to me.

PR #2309, opened 2024-05-16
Add missing write barrier in array-replace

erikcorry:


src/primitive_core.cc
```diff
@@ -1708,6 +1708,7 @@ PRIMITIVE(array_replace) {
   if (index < 0 || from < 0 || from > to || to > source_length) FAIL(OUT_OF_BOUNDS);
   word len = to - from;
   if (index + len > dest_length) FAIL(OUT_OF_BOUNDS);
+  if (dest != source) GcMetadata::insert_into_remembered_set(dest);
```
kasperl, line 1711: Add a comment to explain why you can avoid the write barrier if dest == source.

Also, fix the comment in line 1703: betwen -> between.

kasperl (approved):
Looks good to me.

Can you provoke the issue (before fixing it) with a regression test?

PR #2310, opened 2024-05-16
Messaging: check the class of the list wrapped by ListSlice_

erikcorry:
ListSlice_ can wrap anything that implements List, not just List_.  So we can't assume it's a List_.

### DIFF
--- a/src/messaging.cc
+++ b/src/messaging.cc
@@ -179,21 +179,26 @@ bool MessageEncoder::encode_any(Object* object) {
       if (!is_smi(size)) return false;
       return encode_list(instance, 0, Smi::value(size));
     } else if (class_id == program->list_slice_class_id()) {
-      Object* list = instance->at(Instance::LIST_SLICE_LIST_INDEX);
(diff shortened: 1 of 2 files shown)


src/messaging.cc
```diff
@@ -186,6 +186,10 @@ bool MessageEncoder::encode_any(Object* object) {
       word from = Smi::value(from_object);
       word to = Smi::value(to_object);
       if (is_array(list)) return encode_array(Array::cast(list), from, to);
+      Smi* perhaps_list_class_id = list->class_id();
+      if (perhaps_list_class_id != program->list_class_id()) {
+        return false;
+      }
```
kasperl, line 192: I'd move this up to right after you fetch the `list` from the slice (line 182).

```suggestion
      // We currently do not support serializing slices of non-default lists.
      if (list->class_id() != program->list_class_id()) return false;
```

kasperl (approved):
Looks good to me.

Maybe add a regression test?

PR #2318, opened 2024-05-17
Fix compilation warning.

floitsch:
We had the following errors:
```
possibly dangling reference to a temporary [-Wdangling-reference]
 537 |     for (auto name : selector.shape().names()) {
     |                                             ^
```


kasperl (approved):
Looks good to me.

PR #2321, opened 2024-05-21
Add firmware.envelope to host systems.

floitsch:


tools/firmware.toit
```diff
@@ -122,10 +125,10 @@ write-file-or-print --path/string? output/string -> none:
     print output
 
 main arguments/List:
-  firmware-cmd := build-command
+  firmware-cmd := build-command --create-esp32-only
```
kasperl, line 128: So this way 'tools/firmware' keeps it old interface, but 'toit firmware' is updated?

kasperl (approved):
Looks good to me.

tools/firmware.toit
floitsch: correct.

kasperl (approved):
Still looks good!

PR #2325, opened 2024-05-22
Non-quadratic list, set stringify (like map)

erikcorry:


lib/core/collections.toit
```diff
@@ -539,12 +539,16 @@ abstract class List extends CollectionBase:
     (to - from).repeat: this[it + from] = block.call it
 
   /** See $super. */
-  stringify -> string:
-    str := "["
-    size.repeat:
-      if it != 0: str = str + ", "
-      str = str + this[it].stringify
-    return str + "]"
+  stringify:
```
kasperl, line 542: Mixin for the win?

kasperl (approved):
Looks good to me.

lib/core/collections.toit
```diff
@@ -539,12 +539,16 @@ abstract class List extends CollectionBase:
...
+    key-strings := []
+    size := 0
+    do: | key |
+      key-string := key.stringify
+      size += key-string.size + 2
+      if size > MAX-PRINT-STRING_:
+        return "[$(key-strings.join ", ")..."
```
kasperl, line 549:
```suggestion
        return "[$(key-strings.join ", ")...]"
```

lib/core/collections.toit
```diff
@@ -2626,12 +2630,15 @@ class Set extends HashedInsertionOrderedCollection_ implements Collection:
...
-      if first: first = false else: str = str + ", "
-      str = str + it.stringify
-    return str + "}"
+    key-strings := []
+    size := 0
+    do: | key |
+      key-string := key.stringify
+      size += key-string.size + 2
+      if size > MAX-PRINT-STRING_:
+        return "{$(key-strings.join ", ")..."
```
kasperl, line 2639:
```suggestion
        return "{$(key-strings.join ", ")...}"
```

lib/core/collections.toit
kasperl: I see that this isn't what we do for maps.... Hmm.

lib/core/collections.toit
floitsch, line 542: Keep `-> string`

And add it below.

PR #2327, opened 2024-05-22
Simplify toit-run image.

floitsch:
Start an embedded image and let the Toit code execute the given image.

### DIFF
--- a/tools/run-image.toit
+++ b/tools/run-image.toit
@@ -1,0 +1,18 @@
+// Copyright (C) 2024 Toitware ApS.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; version
+// 2.1 only.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


tools/run-image.toit
```diff
@@ -0,0 +1,77 @@
...
+main arguments:
+  if arguments.is-empty:
+    print_ "Usage: run-image <image|directory>"
+    exit 1
+
+  registry ::= FlashRegistry.scan
+  container-manager ::= initialize-system registry [
+  ]
+  container-manager.register-system-image (SystemImage container-manager)
+
```
kasperl, line 59: Drop one newline.

tools/run-image.toit
```diff
@@ -0,0 +1,77 @@
...
+    unreachable  // Not implemented yet.
+
+add-image path/string -> none:
+  image-data := file.read-content path
+  writer := containers.ContainerImageWriter image-data.size
+  writer.write image-data
+  writer.commit
+
+main arguments:
+  if arguments.is-empty:
```
kasperl, line 51: I would probably delay as much as possible of this logic until we're in a state where things look more normal. So maybe postpone these checks until after the system has been initialized?

kasperl (approved):
Looks good to me.

PR #2329, opened 2024-05-23
Don't check the address for non-programs.

floitsch:


src/flash_allocation.cc
```diff
@@ -44,7 +44,10 @@ uint32 FlashAllocation::Header::compute_checksum(const void* memory) const {
   // is useful if the allocation contains relocated pointers to parts
   // of itself. In that case, those pointers are only correct if the
   // allocation is always access from the same virtual memory address.
-  uint32 initial = Utils::crc32(FORMAT_MARKER, reinterpret_cast<uint8*>(&memory), sizeof(memory));
+  // We only do that for programs, as data doesn't have pointers in it.
+  uint32 initial = type_ != FLASH_ALLOCATION_TYPE_PROGRAM
+      ? FORMAT_MARKER
```
kasperl, line 49:
```suggestion
  // We don't need to do that for data as it doesn't have pointers in it.
  uint32 initial = (type_ == FLASH_ALLOCATION_TYPE_REGION)
      ? FORMAT_MARKER
```

PR #2330, opened 2024-05-23
Don't install images if they already exist.

floitsch:


tools/run-image.toit
```diff
@@ -57,16 +68,20 @@ main arguments:
     print_ "Usage: run-image <image|directory>"
     exit 1
 
+  existing-uuids/Set := {}
```
kasperl, line 71: Maybe this part of the setup should be in a separate method, so we don't keep the new set alive while we're calling `boot`?

tools/run-image.toit
```diff
@@ -41,7 +43,16 @@ class SystemImage extends ContainerImage:
   delete -> none:
     unreachable  // Not implemented yet.
 
-add-image path/string -> none:
+add-image path/string existing-uuids/Set -> none:
```
kasperl, line 46: I'd probably move this below main.

kasperl (approved):
Looks good to me.

tools/run-image.toit
floitsch: done.

tools/run-image.toit
floitsch: done.

PR #2336, opened 2024-05-23
Make some mixins available for those implementing collections

erikcorry:


floitsch:
We can't really use mixins in performance-critical classes. Some optimizations are disabled.

PR #2339, opened 2024-05-24
Continue working on run-image.

floitsch:
Now assumes that the given dir uses the OTA structure.


tools/firmware.toit
```diff
@@ -700,32 +700,32 @@ extract-host parsed/cli.Parsed envelope/Envelope --config-encoded/ByteArray:
 
   // Reserve space for the header.
   header-size := 6 * 4
-  bits.reserve header-size
+  bits.grow-by header-size
```
kasperl, line 703: expand?

tools/run-image.toit
```diff
@@ -43,20 +57,179 @@ class SystemImage extends ContainerImage:
...
+
+class FirmwareWriter_ extends services.ServiceResource implements FirmwareWriter:
+  static image/ByteArray := #[]
+  view_/ByteArray? := null
+  cursor_/int := 0
+  ota-dir-inactive_/string
+
+  constructor provider/FirmwareServiceProvider client/int from/int to/int --ota-dir-inactive/string:
+    ota-dir-inactive_ = ota-dir-inactive
+    if to > image.size: image = image + (ByteArray to - image.size: random 0x100)
```
kasperl, line 111: Add a comment to explain why you do this random padding?

kasperl (approved):
Looks good to me!

tools/firmware.toit
floitsch: I think "grow-by" is fine, and I would prefer not to change the `io.Writer`.

tools/run-image.toit
floitsch: done.

PR #2344, opened 2024-05-24
Add boot.sh as toit file

floitsch:
Pretty much verbatim copy of https://github.com/toitware/ota-linux/blob/main/skeleton/boot.sh.

I changed the license (since there isn't any LICENSE file included in the tarball), and then escaped all `$`.


tools/firmware-boot-sh.toit
```diff
@@ -0,0 +1,132 @@
+// Copyright (C) 2024 Toitware ApS.
```
kasperl, line 1: Should the two licenses just be MIT? It makes sense to include it in full in the bash script.

kasperl (approved):
Looks good to me.

tools/firmware-boot-sh.toit
floitsch: As discussed in person: outer license stays LPGL and inner one is MIT.
I added a comment that states that the inner license is MIT.

PR #2355, opened 2024-05-28
Add storage provider and improve testing.

floitsch:


tools/run-image-exit-codes.toit
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2024 Toitware ApS.
+// Use of this source code is governed by a Zero-Clause BSD license that can
+// be found in the tests/LICENSE file.
+
+/** A new firmware has been written and the system should switch to it. */
+EXIT-CODE-UPGRADE ::= 17
+/** A rollback has been requested. */
```
kasperl, line 7: I would probably add a few more newlines between these.

tools/run-image-exit-codes.toit
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2024 Toitware ApS.
+// Use of this source code is governed by a Zero-Clause BSD license that can
+// be found in the tests/LICENSE file.
+
```
kasperl, line 4: Maybe add a comment here that you can't just change these, because we may have already deployed boot.sh scripts and they will all break.

kasperl (approved):
Looks good to me.

tools/run-image-exit-codes.toit
floitsch: done.

tools/run-image-exit-codes.toit
floitsch: done.

PR #2360, opened 2024-05-29
Add a few more flash operations.

kasperl:


src/primitive_flash_registry.cc
```diff
@@ -124,6 +124,39 @@ PRIMITIVE(get_header_page) {
...
+PRIMITIVE(write_non_header_pages) {
+  PRIVILEGED;
+  ARGS(word, offset, Blob, content);
+  for (auto reservation : reservations) {
+    int reserved_offset = reservation->left();
+    if (reserved_offset < offset) continue;
+    if (reserved_offset > offset) break;
+    ASSERT(reserved_offset == offset);
+
+    int length = Utils::min(reservation->size() - FLASH_PAGE_SIZE, content.length());
```
floitsch, line 150: It looks like we "reserving" FLASH_PAGE_SIZE bytes.
Add a comment why.

floitsch (approved):
LGTM, although I didn't get the FLASH_PAGE_SIZE thing.

src/primitive_flash_registry.cc
kasperl: This is for writing everything but the header page (`write_non_header_pages`). The header page is FLASH_PAGE_SIZE bytes.

PR #2370, opened 2024-05-30
Improve firmware to show which esp32 chip an envelope belongs to.

floitsch:


floitsch:
Sample output:
```
Envelope format version: 8
Kind: esp32
SDK version: v2.0.0-alpha.148.46+floitsch-firmware-esp32.bc9cabf9
Chip: esp32
Containers: 
  system: 
    Kind: snapshot
    Id: a58e22e6-2995-bff4-d19e-9dee7fca86c3
    Size: 232226
    Flags: 
      - trigger=boot
      - critical
```

kasperl (approved):
Looks good to me.

PR #2372, opened 2024-05-30
Allow io.Data for the storage lib.

floitsch:
Also deprecate `write --from` in favor of `write --at`.

### DIFF
--- a/lib/system/storage.toit
+++ b/lib/system/storage.toit
@@ -382,8 +382,10 @@ class Region extends ServiceResourceProxy:
   /**
   Writes the given $data into the region at the given offset $at.
 
-  If the region has already data, it is bit-anded with the
-    data to be written. Use $erase to reset areas to 1.
+  If the region has already data, the new data might be combined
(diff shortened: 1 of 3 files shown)


lib/system/storage.toit
```diff
@@ -372,10 +372,23 @@ class Region extends ServiceResourceProxy:
...
+  Deprecated. Use $(write --at bytes) instead.
+  */
   write --from/int bytes/ByteArray -> none:
     if not resource_: throw "ALREADY_CLOSED"
     flash-region-write_ resource_ from bytes
 
+  /**
+  Writes the given $data into the region at the given offset $at.
+
+  If the region has already data, it is bit-anded with the
```
kasperl, line 385: This isn't necessarily true. It depends on $erase-value, etc.

kasperl (approved):
Looks good to me.

lib/system/storage.toit
floitsch: done.

PR #2377, opened 2024-06-03
Avoid always resetting monotonic time at wakeup.

kasperl:


lib/core/time.toit
```diff
@@ -803,15 +803,16 @@ class Time implements Comparable:
     return parse str --on-error=: throw "INVALID_ARGUMENT"
 
   /**
-  Returns a monotonic microsecond value.
-  This clock source only increments and never jumps.
-  The clock is not anchored and thus has no fixed relationship to the current
-    time.
+  Returns a monotonically increasing microsecond value.
+
+  If $since-wakeup is false, the clock can be used across deep sleeping.
```
floitsch, line 808:
```suggestion
  If $since-wakeup is false, the clock can be used across deep sleeps.
```

lib/core/time.toit
```diff
@@ -803,15 +803,16 @@ class Time implements Comparable:
...
+    Otherwise, the clock is reset when waking up and starts from zero.
```
floitsch, line 809:
```suggestion
    Otherwise, the clock is reset to zero when waking up.
```

floitsch (approved):
LGTM.

PR #2383, opened 2024-06-05
Search for snapshot files when decoding system messages.

floitsch:
Also add 'toit decode'.

### DIFF
--- a/tools/system_message.toit
+++ b/tools/system_message.toit
@@ -111,10 +111,17 @@ main args:
   command.run args
 
 compute-snapshot-dirs -> List:
-  result := []
-  result.add (fs.join xdg.state-home "toit" "snapshots")
-  result.add (fs.join xdg.cache-home "jaguar" "snapshots")
-  return result
+  return [
+    fs.join xdg.state-home "toit" "snapshots",
+    fs.join xdg.cache-home "jaguar" "snapshots",
+  ]
+
+find-snapshot id/uuid.Uuid -> string?:
+  compute-snapshot-dirs.do: | dir/string |
+    path := fs.join dir "$(id).snapshot"
+    if file.is-file path:
+      return path
+  return null
 
 decode parsed/cli.Parsed -> none:
   if not parsed: exit 1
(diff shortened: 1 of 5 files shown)


tools/system_message.toit
```diff
@@ -42,52 +47,109 @@ handle-system-message encoded-system-message snapshot-content -> none
...
+main args:
+  command := build-command --use-old-api
   command.run args
 
-decode-system-message parsed command -> none:
+compute-snapshot-dirs -> List:
+  result := []
+  result.add (fs.join xdg.state-home "toit" "snapshots")
+  result.add (fs.join xdg.cache-home "jaguar" "snapshots")
+  return result
```
kasperl, line 117:
```suggestion
  return [
    fs.join xdg.state-home "toit" "snapshots",
    fs.join xdg.cache-home "jaguar" "snapshots",
  ]
```

tools/system_message.toit
```diff
@@ -42,52 +47,109 @@ handle-system-message encoded-system-message snapshot-content -> none
...
   encoded-system-message := base64.decode parsed["message"]
+  // The decoded message still has its payload encoded in JSON.
+  decoded-system-message := decode-system-message encoded-system-message --on-error=:
+    pipe.print-to-stderr it
+    exit 1
+
+  snapshot-path := parsed["snapshot"]
+  if not snapshot-path:
+    // Try to find the snapshot based on the UUID.
+    dirs := compute-snapshot-dirs
```
kasperl, line 135: Not sure if this is the best way of factoring this. Maybe it would be nicer if you just had a function to find the snapshot-path and inlined the compute-snaphots-dir into that. Then you could use return instead of break from there.

kasperl (approved):
Looks good to me.

tools/system_message.toit
floitsch: That's clearly better...
done.

tools/system_message.toit
floitsch: Still need the `compute-snapshot-dirs` for the help, but created a function.
done.

PR #2385, opened 2024-06-06
Kill the run-image executable if the boot.sh is killed.

floitsch:


tests/envelope/boot-kill-test.toit
```diff
@@ -0,0 +1,66 @@
...
+import net.tcp
+import .exit-codes
+import .boot-kill-source show TEST-PORT-ENV TEST-INTERVAL-MS
+import .util show EnvelopeTest with-test
+
+/**
+Test that we can kill the boot script and that it kills the nested Toit program
+*/
+
+
```
kasperl, line 18: Drop one newline.

tests/envelope/boot-kill-test.toit
```diff
@@ -0,0 +1,66 @@
...
+main args:
+  network := net.open
+
+  started-latch := monitor.Latch
+  stopped-latch := monitor.Latch
+  server := network.tcp-listen 0
+  task --background::
```
kasperl, line 25: Why background?

system/extensions/host/run-image-boot-sh.toit
```diff
@@ -92,8 +92,15 @@ for (( ; ; )); do
   # Run the image.
   export TOIT_FLASH_UUID_FILE=\$PREFIX/\$current/uuid
   export TOIT_FLASH_REGISTRY_FILE=\$PREFIX/flash-registry
-  (cd \$PREFIX/\$current; ./run-image \$PREFIX/\$current \$PREFIX/scratch)
+  pushd \$PREFIX/\$current
+  ./run-image \$PREFIX/\$current \$PREFIX/scratch &
+  popd
+  RUN_IMAGE_PID=\$!
```
kasperl, line 98: I think this should be `run_image_pid` to fit with `exit_code`. Maybe even rename `exit_code` to `run_image_exit_code`?

system/extensions/host/run-image-boot-sh.toit
```diff
@@ -92,8 +92,15 @@ for (( ; ; )); do
...
+  # Make sure to kill the run-image process if we are killed.
+  trap "kill \$RUN_IMAGE_PID; exit" TERM
```
kasperl, line 100: Isn't there a race condition here if we get killed at an unfortunate moment?

Also, doesn't this mean that we will not kill the current process on TERM, because we just handle that instead?

system/extensions/host/run-image-boot-sh.toit
```diff
@@ -89,13 +89,37 @@ for (( ; ; )); do
...
   export TOIT_FLASH_REGISTRY_FILE=\$PREFIX/flash-registry
-  (cd \$PREFIX/\$current; ./run-image \$PREFIX/\$current \$PREFIX/scratch)
-  exit_code=\$?
 
-  if [ \$exit_code -eq $EXIT-CODE-UPGRADE ]; then
+  run_image_pid=
+  should_exit=false
+  # Set a trap for TERM and INT signals, so we can kill the run-image child.
+  # If the run_image_pid was already set, kill it and exit immediately.
+  # Otherwise we just set the should_exit bool, and wait to get the image_pid, before
```
kasperl, line 99:
```suggestion
  # Otherwise we just set the should_exit bool, and wait to get the run_image_pid, before
```

or 

```suggestion
  # Otherwise we just set the should_exit bool, and wait to get the pid, before
```

kasperl (approved):
Looks good to me.

system/extensions/host/run-image-boot-sh.toit
floitsch: done.

PR #2388, opened 2024-06-10
Add a way to nicely set the value of a characteristic.

floitsch:


lib/ble/local.toit
```diff
@@ -341,18 +341,38 @@ class LocalCharacteristic extends LocalReadWriteElement_ implements Attribute:
...
+  */
+  set-value value/io.Data:
+    ble-set-value_ resource_ value
+
+  /**
+  Sets the value of this characteristic and sends a notification or indication if supported.
 
   If the characteristic supports both indications and notifications, then a notification is sent.
+
+  If $update-value is true, sets the value of the characteristic to $value. Any read requests
```
kasperl, line 360: Should `--update-value` be `--set-value` or `set-value` be `update-value`?

kasperl (approved):
Looks good to me.

lib/ble/local.toit
floitsch: switched to `--set-value`.
done.

PR #2390, opened 2024-06-10
Don't emit the test-dir in the boot-output.

floitsch:
The pushd/popd commands were printing their paths.


kasperl (approved):
Looks good to me!

PR #2391, opened 2024-06-10
Fix io.Reader.index-of with a '--to' parameter.

floitsch:


lib/io/reader.toit
```diff
@@ -309,19 +309,25 @@ abstract class Reader implements old-reader.Reader:
...
         index := it.index-of byte --from=start --to=end
         if index >= 0: return offset + (index - start)
-        if to: to -= it.size - start
+        if to:
+          to -= it.size - start
+          if to <= 0: return -1
         offset += it.size - start
         start = 0
 
+    if to and to <= 0: return -1
```
kasperl, line 318: Maybe move this check up before the check for buffered_? It feels like you don't really have to check if there were any elements in buffered_, because you will have checked this after looking in the last buffered segment.

Also, it is a fairly simple check and it just seems nicer to get it out of the way early on.

tests/io-utils.toit
```diff
@@ -20,3 +20,19 @@ class FakeData implements io.Data:
...
+
+  read_ -> ByteArray?:
+    if index_ >= chunks_.size: return null
+    chunk := chunks_[index_++]
+    return chunk is ByteArray
+        ? chunk
+        : chunk.to-byte-array
+
+  close_:
+
```
kasperl, line 38: Maybe add a comment here that explains that the `close_` method is deliberately left empty.

tests/io-reader-index-of-test.toit
```diff
@@ -0,0 +1,35 @@
...
+import .io-utils
+
+main:
+  test "foobarX"
+  test "Xfoobar"
+  test "fooXbar"
+  test "foobar"
+  test ""
+
+test input/string:
```
kasperl, line 16: Does this cover all the new code? All `to <= 0` checks?

kasperl (approved):
Looks good to me!

lib/io/reader.toit
```diff
@@ -309,19 +309,25 @@ abstract class Reader implements old-reader.Reader:
...
+
     while true:
       if not more_:
         if throw-if-absent: throw UNEXPECTED-END-OF-READER
```
kasperl, line 322: It isn't very clear what is supposed to happen if we can't find the $byte because $to restricts our search. Should that throw too? The `--throw-if-absent` cases are not very well covered by tests if at all.

PR #2392, opened 2024-06-11
Fix --throw-if-absent behavior

kasperl:


tests/io-reader-index-of-test.toit
```diff
@@ -15,21 +15,34 @@ main:
...
+            else:
+              expect-equals -1 pos
+              expect-throw io.Reader.UNEXPECTED_END_OF_READER:
+                (reader chunks test-buffered).index-of 'X' --to=to --throw-if-absent
+
+reader chunks/List buffered/bool -> io.Reader:
+  result := TestReader chunks
+  // If requested, we make sure all the reader data is buffered
+  // by looking for a character that isn't there.
+  if buffered: expect-equals -1 (result.index-of ' ')
```
floitsch, line 47:
```suggestion
  if buffered: result.buffer-all
```

lib/io/reader.toit
```diff
@@ -302,6 +302,14 @@ abstract class Reader implements old-reader.Reader:
   Returns -1 if $throw-if-absent is false and the $byte is not in the remaining data.
   */
   index-of byte/int --to/int?=null --throw-if-absent/bool=false -> int:
+    absent := :
```
floitsch, line 305: Maybe we should then just provide a method?
For example?
```
index-of byte/int --to/int?=null [--if-absent] -> int?:
```

But probably for a different PR.

floitsch (approved):
LGTM (without the debug printing).

tests/io-reader-index-of-test.toit
kasperl: Done.

PR #2394, opened 2024-06-11
Fix stripping on macos.

floitsch:


kasperl (approved):
Looks good to me.

kasperl (approved):
Looks good to me.

PR #2398, opened 2024-06-12
Use the config-size when extracting a config in run-image.

floitsch:


system/extensions/host/run-image.toit
```diff
@@ -188,8 +188,8 @@ class Firmware:
 
   config -> ByteArray:
     config-part := part_ PART-CONFIG_
-    // Skip the size header.
-    return config-part[4..]
+    config-size := io.LITTLE-ENDIAN.int32 config-part 0
```
kasperl, line 191:
```suggestion
    config-size := io.LITTLE-ENDIAN.uint32 config-part 0
```

kasperl (approved):
Looks good to me.

PR #2414, opened 2024-06-28
Don't add the version to the released toitdoc file.

floitsch:
By not using a version in the filename it's easier to just fetch the latest released file.


kasperl (approved):
Looks good to me.

PR #2420, opened 2024-07-04
Make deprecation a part of the external summary.

floitsch:


tests/lsp/depB-compiler-test.toit
```diff
@@ -0,0 +1,113 @@
...
+import .lsp-client show LspClient run-client-test
+import expect show *
+import system
+import system show platform
+
+main args:
+  run-client-test args: test it
+  // run-client-test --use-toitlsp args: test it
+
+// Make sure that the `abstract` keyword is correctly handled.
```
kasperl, line 14: ?

tests/lsp/depB-compiler-test.toit
```diff
@@ -0,0 +1,113 @@
...
+// be found in the tests/LICENSE file.
+
+import .lsp-client show LspClient run-client-test
+import expect show *
+import system
+import system show platform
+
+main args:
+  run-client-test args: test it
+  // run-client-test --use-toitlsp args: test it
```
kasperl, line 12: In or out?

src/compiler/lsp/protocol_summary.cc
```diff
@@ -318,6 +318,7 @@ void Writer::print_field(ir::Field* field) {
   print_range(field->range());
 
   this->printf("%s\n", field->is_final() ? "final" : "mutable");
+  this->printf("%s\n", field->is_deprecated() ? "deprecated" : "not deprecated");
```
kasperl, line 321: Maybe we should try to avoid having the "non-abstract" and "not deprecated" strings contain the positive variant?

kasperl (approved):
Looks good to me.

src/compiler/lsp/protocol_summary.cc
floitsch: done. Now simply sending "-" in both cases for the negative case.

tests/lsp/depB-compiler-test.toit
floitsch: For now: in. Will be removed in a different PR.
done.

tests/lsp/depB-compiler-test.toit
floitsch: done.

PR #2424, opened 2024-07-05
Pass the timeout to handle-request instead of to add-characteristic.

floitsch:


lib/ble/local.toit
```diff
@@ -154,6 +155,39 @@ class LocalService extends Resource_ implements Attribute:
   remove-characteristic_ characteristic/LocalCharacteristic -> none:
     characteristics_.remove characteristic
 
+
+  /**
+  Deprecated. Pass the timeout to the $LocalCharacteristic.handle-read-request function instead.
+  */
+  // TODO(florian): when removing this function, also remove the argument to the constructor.
+  add-characteristic -> LocalCharacteristic
```
floitsch, line 163: A bit of code-duplication, but I didn't want to call the deprecated method from ourselves (as that would lead to deprecation warnings), and adding a third function just didn't feel worth it.

kasperl (approved):
Looks good to me.

PR #2440, opened 2024-07-19
Make LSP translation functions global.

floitsch:
No need to make the LSP uri-path translator a strategy.


kasperl (approved):
Looks good to me.

PR #2445, opened 2024-07-19
Revive the Toit version of the repro server.

floitsch:
The Toit version was removed in https://github.com/toitware/toit/pull/3972.
A lot of the code here is just reverting that commit.


tools/lsp/server/repro.toit
```diff
@@ -156,3 +157,73 @@ create-archive project-uri/string? compiler-path/string entry-path/string out-pa
       --protocol=protocol
       --cwd-path=cwd
       --include-sdk
+
```
kasperl, line 160: Drop one newline?

kasperl (approved):
Looks good to me.

PR #2449, opened 2024-07-19
Replace "contents" with "content"

floitsch:


floitsch:
As discussed: this PR was wrong.
It was changed to have no changes.

PR #2452, opened 2024-07-19
Handle parameter ref-kinds

floitsch:
Parameter references shouldn't just be "other" references.
No functional change.

The corresponding viewer PR: https://github.com/toitware/web-toitdocs/pull/268


kasperl (approved):
Looks good to me.

PR #2453, opened 2024-07-19
Be better at escaping in Toitdocs

floitsch:
When users escaped '\$' or similar that didn't always work.

Now the six characters \ \` ' $ " and \[ are escaped when they are following a backslash, but all other \X are left intact. This allows to write \n without needing to write it with two backslashes.


kasperl (approved):
Looks good to me.

PR #2455, opened 2024-07-19
Update some of the LSP server toitdocs

floitsch:
They were still using `[foo]` for references instead of `$foo`.

There are still some left...


kasperl (approved):
Looks good to me.

PR #2457, opened 2024-07-19
Introduce full-ranges.

floitsch:
The AST nodes now have a `full_range()` function, which gives the range of the node including all its children and delimiters.


kasperl (approved):
Looks good to me.

PR #2458, opened 2024-07-19
Remove work-around for toitdocs

floitsch:
We can now use the full-range instead of hacking our way around it.


kasperl (approved):
Looks good to me.

PR #2460, opened 2024-07-19
Fix the outline ranges

floitsch:
The LSP didn't have the correct outline ranges. Now that we have "full_ranges" we can finally fix it.


src/compiler/outline_ranges.cc
```diff
@@ -0,0 +1,61 @@
...
+
+#include "ast.h"
+#include "comments.h"
+
+namespace toit {
+namespace compiler {
+
+static Source::Range compute_outline_range(ast::Node* node, CommentsManager& manager) {
+  int earliest_comment = manager.find_closest_before(node);
+  auto full_range = node->full_range();
```
kasperl, line 26: Tada.

kasperl (approved):
Looks good to me.

PR #2462, opened 2024-07-19
Prefix operators with "operator" in the summary.

floitsch:
This also means that they are prefixed with operator in the toitdocs.


kasperl (approved):
Looks good to me.

PR #2463, opened 2024-07-19
Include the selector in Dot selection range.

floitsch:
It's not perfect (and sometimes selects the wrong thing), but better
than just selecting the ".".


kasperl (approved):
Nice!

PR #2464, opened 2024-07-22
Disable boot-upgrade-test with type propagation

floitsch:
This test ended up taking more than 5 minutes.


floitsch:
TBR.

kasperl (commented):
Looks good to me.

PR #2467, opened 2024-07-22
Make url encoding|decoding more predictible.

floitsch:
This is a breaking change:
- `encode` now always returns a string.
- `decode` always returns a string, and `decode-binary` always returns a ByteArray.

### DIFF
--- a/lib/encoding/url.toit
+++ b/lib/encoding/url.toit
@@ -59,9 +59,6 @@ Decodes the given $data using URL-encoding, also known as percent encoding.
 The function is liberal, accepting unencoded characters that should be
   encoded, with the exception of '%'.
 
-Returns a string if the input is a string and no decoding is needed.
-Returns a byte array otherwise.
-
 Plus signs (+) are not decoded to spaces.
 
 # Example
--- a/lib/encoding/url.toit
+++ b/lib/encoding/url.toit
@@ -78,12 +78,12 @@ The encoded $data may contain characters that are not valid UTF-8.
 */
 decode-binary data/string -> ByteArray:
   count := 0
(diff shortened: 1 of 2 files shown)


lib/encoding/url.toit
```diff
@@ -37,55 +45,55 @@ url-encode_ from -> any:
...
 /**
 Decodes the given $data using URL-encoding, also known as percent encoding.
 The function is liberal, accepting unencoded characters that should be
   encoded, with the exception of '%'.
-Takes a string or a byte array, and may return a string or a ByteArray.
-  (Both string and ByteArray have a to-string method.)
-Does not check for malformed UTF-8, but calling to-string on the return
-  value will throw on malformed UTF-8.
+
+Returns a string if the input is a string and no decoding is needed.
```
kasperl, line 62: wat

floitsch: Removed.
done.

lib/encoding/url.toit
```diff
@@ -37,55 +45,52 @@ url-encode_ from -> any:
...
-  else if data is ByteArray:
-    if (data.index-of '%') == -1: return data
-  else:
-    throw "WRONG_OBJECT_TYPE"
+decode-binary data/string -> ByteArray:
   count := 0
-  data.do: | c |
-    if c == '%': count++
-  result := ByteArray data.size - count * 2
+  for i := 0; i < data.byte-size; i++:
```
kasperl, line 81: Why not use `data.byte-size.repeat` here too?

kasperl (approved):
Looks good to me.

lib/encoding/url.toit
floitsch: done.

PR #2468, opened 2024-07-22
Type data-bytes of UUIDs.

floitsch:
Fixes https://github.com/toitware/toit/issues/4197.

This is a breaking change as the function used to accept anything that had a `to-byte-array` (like UUIDs themselves), but the documentation always stated "strings or bytearrays".


lib/uuid.toit
```diff
@@ -97,8 +93,7 @@ UUIDs are equivalent to a 128-bit number. Through the use of
 See https://en.wikipedia.org/wiki/Universally_unique_identifier.
 */
 class Uuid:
-  // TODO(4196): the field should be types as `ByteArray`.
-  bytes_ ::= ?
+  bytes_/io.Data ::= ?
```
kasperl, line 96: Why is this not a byte array?

floitsch: Changed to ByteArray.
done.

lib/uuid.toit
```diff
@@ -97,8 +93,7 @@ UUIDs are equivalent to a 128-bit number. Through the use of
 See https://en.wikipedia.org/wiki/Universally_unique_identifier.
 */
 class Uuid:
-  // TODO(4196): the field should be types as `ByteArray`.
-  bytes_ ::= ?
+  bytes_/ByteArray ::= ?
```
kasperl, line 96:
```suggestion
  bytes_/ByteArray
```

lib/uuid.toit
```diff
@@ -107,7 +102,7 @@ class Uuid:
   The given parameter must be a byte array of size 16.
   */
   constructor .bytes_:
-    if bytes_.size != SIZE: throw "INVALID_UUID"
+    if bytes_.byte-size != SIZE: throw "INVALID_UUID"
```
kasperl, line 105: Why are all these changes to the Uuid class necessary or useful?

lib/uuid.toit
floitsch: done.

kasperl (approved):
Looks good to me.

PR #2469, opened 2024-07-22
Move uuid functions into class.

### LINKED ISSUE #479: Rename SIZE constant in uuid library
We should consider renaming it to something less ambiguous.

floitsch:
This makes the class (`Uuid`) work like a namespace, allowing the library to be importade with `import uuid show *`.

Fixes #479.


lib/uuid.toit
```diff
@@ -93,6 +42,76 @@ UUIDs are equivalent to a 128-bit number. Through the use of
...
+  /** Bytesize of a UUID. */
+  // TODO(4193): should the name of the constant be less ambiguous?
+  static SIZE ::= 16
+
+  /**
+  The Nil UUID.
+  This UUID is composed of all bits set to zero.
+  */
+  static NIL ::= Uuid
+    ByteArray SIZE
```
kasperl, line 55:
```suggestion
      ByteArray SIZE
```

kasperl (approved):
Looks good to me.

PR #2473, opened 2024-07-23
Improve documentation wrt default values.

### LINKED ISSUE #2437: Toitdoc should always state the case of non-default values
As indicated in https://github.com/toitlang/toit/issues/2436.

When a comment states "If parameter x is false", then the reader assumes that "false" is not the default value. We should either say "false (the default)", or state the opposite "If parameter x is true, then... . Otherwise, ..".

- [ ] go through the documentation and fix any wording that does this wrong.
- [ ] update the toitdoc guidelines (help) to include this.

floitsch:
When mentioning a parameter with a default value we should either say what the default value is, or mention the case when passing the value makes a difference. For example, "If foo is true, ...", then the default value should be false.

Partially fixes #2437.


kasperl (approved):
Looks good to me.

PR #2475, opened 2024-07-29
Fix install-range constraint.

floitsch:
Just using the given term as prefix doesn't work. For example, 'pkg install foo@1' would match 'v10.0.1'.

Also some refactorings to use conventions that are more in line with the rest of the repository.

### DIFF
--- a/tools/pkg/constraints.toit
+++ b/tools/pkg/constraints.toit
@@ -15,19 +15,28 @@
 
 import .semantic-version
 import .parsers.constraint-parser
+import .parsers.semantic-version-parser
 
 class Constraint:
-  constraints/List
+  simple-constraints/List
   source/string
 
-  constructor .source/string:
+  constructor --.simple-constraints --.source:
+
+  static parse source/string -> Constraint:
     parsed := (ConstraintParser source).constraints --consume-all
-    constraints = []
+    constraints := []
     parsed.do: | constraint/ConstraintParseResult |
       version := SemanticVersion.from-parse-result constraint.semantic-version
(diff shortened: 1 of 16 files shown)


tools/pkg/constraints.toit
```diff
@@ -15,29 +15,60 @@
...
       else if constraint.prefix == "":
         constraints.add (SimpleConstraint "=" version)
       else:
         constraints.add (SimpleConstraint constraint.prefix version)
+    return Constraint --simple-constraints=constraints --source=source
+
+  static parse-range source/string -> Constraint:
+    parser := SemanticVersionParser source --allow-missing-minor
+    parse-result := parser.semantic-version --consume-all
+    // triple/TripleParseResult
```
kasperl, line 52: What are these comments?

tools/pkg/constraints.toit
```diff
@@ -15,29 +15,60 @@
...
+    // pre-releases/List
+    // build-numbers/List
+    triple := parse-result.triple.triple
```
kasperl, line 55: This looks kinda silly. Why not have major, minor, and patch on the TripleParseResult class?

kasperl (approved):
Looks good to me.

tools/pkg/constraints.toit
floitsch: Removed the comments.
done.

tools/pkg/constraints.toit
floitsch: I'm not super happy with the structure either.
Will put it on my list of refactoring TODOs.

PR #2480, opened 2024-07-29
Implement preferred versions.

floitsch:
This makes it possible to minimize changes to lock files when installing packages.


tools/pkg/solver/solver.toit
```diff
@@ -152,7 +152,7 @@ class SolverPackage:
 /**
 A (lazy) database of packages that are available.
 */
-class SolverDb:
+class Db_:
```
kasperl, line 155: Maybe just call this `Database_` now?

kasperl (approved):
Looks good to me.

tools/pkg/solver/solver.toit
floitsch: done.

PR #2481, opened 2024-07-29
Rename "package-file" to "specification"

floitsch:
It's just too confusing to call 'package.yaml' the package file.


kasperl (approved):
Looks good to me.

PR #2491, opened 2024-07-31
Warn when conditions are always true due to nullables.

floitsch:
Catches cases like `if not x:` where `x` is non-nullable.


kasperl (approved):
Looks good to me.

PR #2492, opened 2024-07-31
Expose True and False and use it for flags that must be true.

floitsch:
We sometimes use boolean flags to dispatch to different methods (like
`string.do --raw`). Now we can type these as `/True`.


tests/boolean-types-test.toit
```diff
@@ -0,0 +1,28 @@
...
+
+global1/True := true
+global2/False := false
+
+main:
+  x := true
+  y := false
+  expect-throw "AS_CHECK_FAILED": foo y
+  expect-throw "AS_CHECK_FAILED": bar x
+  // TODO(florian): the following two tests currently fail.
```
kasperl, line 24: Still?

kasperl (approved):
Looks good to me.

tests/boolean-types-test.toit
floitsch: done.

PR #2494, opened 2024-08-06
Enable asserts by default for -O0 and -O1 and add flag.

### LINKED ISSUE #2493: assert: how-to ?
```
richards.toit

  assert: scheduler.queue-count == EXPECTED-QUEUE-COUNT
  assert: scheduler.hold-count == EXPECTED-HOLD-COUNT
```

I haven't been able to figure-out how to trigger the assert or where to see what happened.

floitsch:
Partially fixes #2493.
We still need to expose it in Jaguar.

### DIFF
--- a/tools/toit.toit
+++ b/tools/toit.toit
@@ -519,7 +519,7 @@ compile-or-analyze-or-run --command/string parsed/cli.Parsed:
 
     enable-asserts := optimization < 2
     if parsed.was-provided "enable-asserts":
-      // An explicit --assert, or --no-assert, overrides the default.
+      // An explicit --enable-asserts, or --no-enable-asserts, overrides the default.
       if parsed["enable-asserts"]:
         enable-asserts = true
       else:
--- a/tools/toit.toit
+++ b/tools/toit.toit
@@ -512,10 +512,8 @@ compile-or-analyze-or-run --command/string parsed/cli.Parsed:
   if command == "analyze":
     args.add "--analyze"
   else:
-    optimization/int := 0
-    if parsed["optimization-level"]:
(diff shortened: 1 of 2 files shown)


tools/toit.toit
```diff
@@ -510,10 +512,21 @@ compile-or-analyze-or-run --command/string parsed/cli.Parsed:
...
     args.add "--analyze"
   else:
+    optimization/int := 0
     if parsed["optimization-level"]:
-      optimization/int := parsed["optimization-level"]
+      optimization = parsed["optimization-level"]
       if not 0 <= optimization <= 2: error "Invalid optimization level"
-      args.add "-O$optimization"
+
+    args.add "-O$optimization"
```
kasperl, line 520: So this changes things a bit. The compiler can no longer pick the default and you've changed it zero.

kasperl: So why do we have an if when we have a default value?

tools/toit.toit
```diff
@@ -510,10 +512,21 @@ compile-or-analyze-or-run --command/string parsed/cli.Parsed:
...
+
+    enable-asserts := optimization < 2
+    if parsed.was-provided "enable-asserts":
+      // An explicit --assert, or --no-assert, overrides the default.
```
kasperl, line 524:
```suggestion
      // An explicit --enable-asserts, or --no-enable-asserts, overrides the default.
```

tools/toit.toit
floitsch: Because I forgot to push my changes after the feedback...

kasperl (approved):
Looks good to me.

PR #2499, opened 2024-08-08
Report deprecation warnings for imports.

floitsch:


src/compiler/resolver.cc
```diff
@@ -132,6 +132,23 @@ ir::Program* Resolver::resolve(const std::vector<ast::Unit*>& units,
...
+  // losing import information.
+  for (int i = -1; i < static_cast<int>(modules.size()); i++) {
+    if (i == entry_index) continue;
+    Module* module = i == -1 ? entry_module : modules[i];
+    for (auto imported : module->imported_modules()) {
+      if (imported.module->is_deprecated()) {
+        auto import_node = imported.import;
+        if (import_node) {
+          auto range = import_node->range().extend(import_node->segments().last()->range());
+          diagnostics()->report_warning(range, "Deprecated library");
```
kasperl, line 146: "Importing deprecated library" might be slightly better.

kasperl (approved):
Looks good to me.

src/compiler/resolver.cc
floitsch: done.

PR #2501, opened 2024-08-08
Don't include '\r' in Toitdoc.

floitsch:


src/compiler/toitdoc_parser.cc
```diff
@@ -873,6 +873,11 @@ std::string ToitdocParser::make_string(int from, int to) {
       if (i >= to) break;
     }
     int c = text[i];
+    if (c == '\r' && i < to - 1 && text[i + 1] == '\n') {
```
kasperl, line 876: Back in the day, you would also need to collapse \n\r to \n.

kasperl (approved):
Looks good to me.

src/compiler/toitdoc_parser.cc
floitsch: Fortunately these machines have died out...

PR #2502, opened 2024-08-08
Add default packages.

floitsch:


Makefile
```diff
@@ -284,6 +284,10 @@ install-sdk:
 	mkdir -p "$(DESTDIR)$(prefix)"/lib
 	cp -R "$(CURDIR)"/lib/* "$(DESTDIR)$(prefix)"/lib
 	find "$(DESTDIR)$(prefix)"/lib -type f -exec chmod 644 {} \;
+	mkdir -p "$(DESTDIR)$(prefix)"/default
```
kasperl, line 287: I think I'd prefer packages-default for this to group more nicely with packages/.packages.

src/compiler/compiler.cc
```diff
@@ -131,10 +131,6 @@ class Pipeline {
   virtual ast::Unit* parse(Source* source);
   virtual void setup_lsp_selection_handler();
 
-  // Gives the Pipeline the opportunity to change the program once it was
```
kasperl, line 134: Maybe merge from main to make this easier to review?

floitsch: done.

PR #2503, opened 2024-08-09
Add way to list all transitive source files

floitsch:
We can then later use this to remove packages that aren't used.


kasperl (approved):
Looks good to me.

PR #2513, opened 2024-08-29
Fix BLE on macos.

### LINKED ISSUE #2512: BLE Example scan duration never times out on MacOS with Toit SDK alpha 159
Running the example ble example from https://github.com/toitlang/toit/blob/master/examples/ble/scan.toit just results in a hanging process that never times out on MacOS. 

The BLE Central scan never times out, as though it's not respecting the scan duration param? 

```
sw_vers
ProductName:            macOS
ProductVersion:         14.5
BuildVersion:           23F79


jag version
Version:         v1.39.0
SDK version:     v2.0.0-alpha.159
Build date:      2024-08-27T19:37:04Z
```

I have noticed this specifically with the v2.0.0 alpha 159 SDK but it was working fine in the past.

floitsch:
Fixes #2512.

### DIFF
--- a/src/resources/ble_darwin.mm
+++ b/src/resources/ble_darwin.mm
@@ -757,7 +757,7 @@ PRIMITIVE(create_adapter) {
   group->register_resource(adapter_resource);
   proxy->set_external_address(adapter_resource);
 
-  // On macos, the adapter is immediately available.
+  // On macOS, the adapter is immediately available.
   HostBleEventSource::instance()->on_event(adapter_resource, kBleStarted);
 
   return proxy;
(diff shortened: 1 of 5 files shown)


src/resources/ble_darwin.mm
```diff
@@ -757,6 +756,10 @@ - (id)initWithResource:(toit::BleResource*)resource {
   BleAdapterResource* adapter_resource = _new BleAdapterResource(group);
   group->register_resource(adapter_resource);
   proxy->set_external_address(adapter_resource);
+
+  // On macos, the adapter is immediately available.
```
kasperl, line 760:
```suggestion
  // On macOS, the adapter is immediately available.
```

kasperl (approved):
Looks good to me.

PR #2514, opened 2024-08-29
Fix BLE on macos. (#2513)

### LINKED ISSUE #2512: BLE Example scan duration never times out on MacOS with Toit SDK alpha 159

floitsch:
Backport of BLE macos fix.

### DIFF
--- a/lib/ble/local.toit
+++ b/lib/ble/local.toit
@@ -24,6 +24,7 @@ class Peripheral extends Resource_:
   constructor .adapter bonding/bool secure-connections/bool:
     resource := ble-create-peripheral-manager_ adapter.resource_ bonding secure-connections
     super resource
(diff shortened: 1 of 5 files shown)


kasperl (approved):
Looks good to me.

PR #2518, opened 2024-09-05
Add a way to use the i2c pullup resistors.

floitsch:


src/resources/i2c_esp32.cc
```diff
@@ -69,7 +69,7 @@ class I2cResourceGroup : public ResourceGroup {
 MODULE_IMPLEMENTATION(i2c, MODULE_I2C);
 
 PRIMITIVE(init) {
-  ARGS(int, frequency, int, sda, int, scl);
+  ARGS(int, frequency, int, sda, int, scl, bool, sda_pullup, bool, scl_pullup);
```
kasperl, line 72: We should move the type to after the parameter name :)

kasperl (approved):
Looks good to me.

PR #2520, opened 2024-09-05
Add generic Locker/Unlocker

kasperl:


floitsch (approved):
LGTM!

src/mutex.cc
```diff
@@ -0,0 +1,49 @@
...
+    int previous_level = previous_locker->mutex_->level();
+    if (level <= previous_level) {
+      FATAL("trying to take lock of level %d (%s) while holding lock of level %d (%s)",
+          level, mutex_->name(), previous_level, previous_locker->mutex_->name());
+    }
+  }
+  // Lock after checking the precondition to avoid deadlocking
+  // instead of just failing the precondition check.
+  mutex_->lock();
+  // Only update variables after we have the lock - that grants right
```
kasperl, line 43: This feels a little bit misguided. Both the Locker and the Thread are thread-local, unless Locker is used in a bizarre non-stack-allocated way.

floitsch: I agree. We could remove the comment.

floitsch (approved):
Still LGTM.

PR #2521, opened 2024-09-05
Allow reentrant Lockers

kasperl:


src/mutex.cc
```diff
@@ -21,27 +21,47 @@ namespace toit {
 void Locker::leave() {
   Thread* thread = Thread::current();
   if (thread->locker_ != this) FATAL("unlocking would break lock order");
-  thread->locker_ = previous_;
-  // Perform the actual unlock.
-  mutex_->unlock();
+  thread->locker_ = previous();
```
kasperl, line 24: I'm considering setting `previous_` to `NULL` here, just because it feels weird that we keep the state in the locker and only get rid of it when we ever re-enter.

floitsch: setting to NULL works for me.

tests/ctest/vm-functions-test.cc
```diff
@@ -35,11 +35,29 @@ static void many_pages() {
   }
 }
 
+void reentrant_locking() {
```
floitsch, line 38: Add tests with unlockers.
For example:
```
  { Locker outer(m1);
    { Locker inner(m1);
    }
    Locker inbetween(m3);
    { Locker inner(m1);
      { Unlocker u(inbetween);
        Locker innermost(m2);
      }
    }
  }
  { Locker outer(m1);
  }
  ```

File: src/mutex.cc [hunk 3]
```diff
@@ -21,27 +21,47 @@ namespace toit {
...
+  // Perform the actual unlock unless it was a reentrant locking.
+  bool unlock = !reentrant_;
+  reentrant_ = false;
```
floitsch, line 27: I think I would not reset this field.
If I'm not wrong the field should only be useful when the lock is taken.
You can then remove the `ASSERT` in line 42.

src/mutex.cc
```diff
@@ -21,27 +21,47 @@ namespace toit {
...
+    int previous_level = previous->mutex()->level();
     if (level <= previous_level) {
-      FATAL("trying to take lock of level %d (%s) while holding lock of level %d (%s)",
-          level, mutex_->name(), previous_level, previous_locker->mutex_->name());
+      reentrant = is_reentrant(previous, mutex);
+      if (!reentrant) {
+        FATAL("trying to take lock of level %d (%s) while holding lock of level %d (%s)",
+            level, mutex->name(), previous_level, previous->mutex()->name());
+      }
+      reentrant_ = true;
```
floitsch, line 60: I would set the `reentrant_` to false otherwise.

floitsch (approved):
LGTM!

PR #2524, opened 2024-09-09
Add 'BleUuid.to-string'.

floitsch:


lib/ble/ble.toit
```diff
@@ -55,6 +58,14 @@ class BleUuid:
     else:
       return data_
 
+  /**
+  Returns a string representation of this UUID.
+
+  If a deterministic UUID string representation is needed, prefer to use $to-string.
```
kasperl, line 64:
```suggestion
  If a deterministic UUID string representation is needed, prefer using $to-string.
```

kasperl (approved):
Looks good to me.

PR #2525, opened 2024-09-09
Improve error message when using a restricted pin.

floitsch:


lib/gpio/pin.toit
```diff
@@ -391,7 +393,10 @@ gpio-init_:
   #primitive.gpio.init
 
 gpio-use_ resource-group num allow-restricted:
-  #primitive.gpio.use
+  #primitive.gpio.use:
+    if it == RESTRICTED-PIN-ERROR_:
```
kasperl, line 397: What is going on with this "RESTRICTED" <-> "RESTRICTED_PIN" dance?

src/resources/gpio_esp32.cc
```diff
@@ -38,6 +38,8 @@
 
 namespace toit {
 
+static const char* const RESTRICTED_PIN_ERROR = "RESTRICTED";
```
kasperl, line 41: Can we use one of the predefined / preallocated ones instead?

lib/gpio/pin.toit
floitsch: removed.
done.

src/resources/gpio_esp32.cc
floitsch: Switched to "PERMISSION_DENIED".
done.

src/primitive.h
```diff
@@ -1225,6 +1225,7 @@ class Primitive {
   static HeapObject* mark_as_error(HeapObject* object) { return object->mark(); }
   static Object* unmark_from_error(Program* program, Object* object);
   static Object* os_error(int error, Process* process);
+  static Object* error(const char* error, Process* process);
```
kasperl, line 1228: This is not needed anymore.

kasperl (approved):
Looks good to me.

src/primitive.h
floitsch: done.

PR #2526, opened 2024-09-09
Fix unused warning and add negative test for locks.

floitsch:


tests/ctest/CMakeLists.txt
```diff
@@ -50,58 +50,102 @@ add_custom_target(
     DEPENDS ${CTEST_DIR}/lib
```
floitsch, line 50: I recommend to review this file without whitespace changes.

kasperl (approved):
Looks good to me.

PR #2531, opened 2024-09-10
Infer literal types.

floitsch:
Also add test for removal of type-check for global assignments.

### DIFF
--- a/src/compiler/optimizations/utils.cc
+++ b/src/compiler/optimizations/utils.cc
@@ -38,7 +38,15 @@ bool is_This(Node* node, Class* holder, Method* method) {
   return target->as_Parameter()->index() == 0;
 }
 
-Type compute_guaranteed_type(Expression* node, Class* holder, Method* method) {
+static Type find_literal_type(List<Type> literal_types, Symbol symbol) {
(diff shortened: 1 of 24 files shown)


src/compiler/optimizations/utils.cc
```diff
@@ -38,7 +38,15 @@ bool is_This(Node* node, Class* holder, Method* method) {
   return target->as_Parameter()->index() == 0;
 }
 
-Type compute_guaranteed_type(Expression* node, Class* holder, Method* method) {
+static Type find_literal_type(List<Type> literal_types, Symbol symbol) {
+  for (auto literal_type : literal_types) {
+    if (literal_type.klass()->name() == symbol) return literal_type;
+  }
+  FATAL("Couldn't find literal type");
+  return Type::invalid();
```
kasperl, line 46: Is this return necessary? FATAL is fatal.

kasperl (approved):
Looks good to me.

src/compiler/optimizations/utils.cc
floitsch: I copied it from a different place, so I think yes. Most likely to silence the compiler.

PR #2534, opened 2024-09-17
Fix UART flush on the ESP32.

floitsch:


src/resources/uart_esp32.cc
```diff
@@ -178,6 +178,7 @@ class UartResource : public EventQueueResource {
   UART_ISR_INLINE void enable_tx_interrupts(bool begin);
   UART_ISR_INLINE void disable_tx_interrupts(bool done);
 
+  UART_ISR_INLINE bool has_flushed_tx_fifo();
```
kasperl, line 181: Perhaps `is_empty_tx_fifo()`? Somehow I think the "has flushed" could mean that it only returns true if you actively asked for flushing.

src/resources/uart_esp32.cc
```diff
@@ -461,10 +462,14 @@ UART_ISR_INLINE void UartResource::disable_tx_interrupts(bool done) {
   }
 }
 
+UART_ISR_INLINE bool UartResource::has_flushed_tx_fifo() {
+  return get_tx_fifo_free() == SOC_UART_FIFO_LEN && uart_toit_hal_is_tx_idle(hal_);
```
kasperl, line 466: Maybe add a comment that explains why we need to check for idle? It wasn't clear to the original writer of the code :)

kasperl (approved):
Looks good to me.

src/resources/uart_esp32.cc
floitsch: done.

src/resources/uart_esp32.cc
floitsch: done.

PR #2535, opened 2024-09-17
Improve and expose SPI devices for Linux.

floitsch:


src/resources/spi_linux.cc
```diff
@@ -32,12 +33,34 @@ extern Object* return_open_error(Process* process, int err);
...
+  ARGS(cstring, pathname, int, frequency, int, mode);
+  if (frequency <= 0) FAIL(INVALID_ARGUMENT);
   // We always set the close-on-exec flag otherwise we leak descriptors when we fork.
   // File descriptors that are intended for subprocesses have the flags cleared.
   int fd = open(pathname, O_CLOEXEC | O_RDWR);
   if (fd < 0) return return_open_error(process, errno);
+  int result = ioctl(fd, SPI_IOC_WR_MAX_SPEED_HZ, &frequency);
+  if (result < 0) {
+    close(fd);
+    return return_open_error(process, errno);
```
floitsch, line 47: Not sure about using "return_open_error" here.
Do we have something that gets the perror?

kasperl: `Primitive::os_error`

kasperl (approved):
Looks good to me.

src/resources/spi_linux.cc
floitsch: done.

PR #2543, opened 2024-09-23
Introduce contents.json

floitsch:


src/compiler/lock.cc
```diff
@@ -935,20 +986,24 @@ PackageLock PackageLock::read(const std::string& lock_file_path,
...
   }
 
+  Map<std::string, Map<std::string, std::string>> mappings;
+  fill_package_mappings(&mappings, package_dirs, fs);
+
   Map<std::string, std::string> path_to_package;
   for (auto package_id : lock_content.packages.keys()) {
     auto entry = lock_content.packages.at(package_id);
 
     // For simplicity we are using a loop from -1 to length, so we don't
```
kasperl, line 996:
```suggestion
    // For simplicity we are using a loop from -2 to length, so we don't
```

kasperl: Would it be terrible to factor this out and introduce a helper that you can call twice before the loop?

kasperl (approved):
Looks good to me.

src/compiler/lock.cc
floitsch: https://github.com/toitlang/toit/pull/2545

PR #2547, opened 2024-10-01
Use toit for building

floitsch:
Instead of using 'toit.compile', 'toit.run' etc, use 'toit' as much as possible.

Update README to use the 'toit' binary.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,10 @@
 version: 2
 updates:
-  - package-ecosystem: "github-actions" # Necessary to update action hashs
-    directories: |
-      "/"
-      "/actions/setup-build"
+  - package-ecosystem: "github-actions" # Necessary to update action hashes.
+    directories:
+      - "/"
+      - "/actions/setup-build"
     schedule:
       interval: "weekly"
-    # Allow up to 3 opened pull requests for github-actions versions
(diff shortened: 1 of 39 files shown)


tools/pkg-install.cmake
```diff
@@ -0,0 +1,111 @@
...
+set(PACKAGES_DIR ".packages" CACHE PATH "Directory to store cloned packages.")
+set(QUIET FALSE CACHE BOOL "Suppress output.")
+
+if (QUIET)
+  # Run this script again but with OUTPUT_QUIET set.
+  execute_process(
+    COMMAND "${CMAKE_COMMAND}"
+      "-DPACKAGE_LOCK_PATH=${PACKAGE_LOCK_PATH}"
+      "-DPACKAGES_DIR=${PACKAGES_DIR}"
+      "-DQUIET=FALSE"
```
kasperl, line 14: This looks a little bit odd.

kasperl (approved):
Looks good to me.

tools/pkg-install.cmake
floitsch: It's a recursive call to itself (thus setting the 'QUIET' to false to avoid an infinite loop.
Note that it has the "OUTPUT_QUIET" option set (line 16).
Added a comment.

PR #2551, opened 2024-10-02
Add do-release work-flow dispatch.

floitsch:


.github/workflows/ci.yml
```diff
@@ -52,6 +52,15 @@ on:
         required: false
         type: boolean
         default: true
+      fix-release:
```
kasperl, line 55: Maybe use rebuild rather than fix consistently? Not really sure what fix means.

.github/workflows/ci.yml
```diff
@@ -468,46 +485,46 @@ jobs:
...
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: build/firmware-esp32c3.gz
-          tag: ${{ github.event.release.tag_name }}
+          tag: ${{ env.TOIT_VERSION }}
           overwrite: true
 
       - name: Upload ESP32-S2 release firmware artifacts
         if: |
-          github.event_name == 'release' &&
+          (github.event_name == 'release' || env.IS_FIX_RELEASE) &&
```
kasperl, line 510: Could this be combined so we always have something like `env.BUILD_RELEASE` set when we should build the release? If the `github.event_name` is `'release'` it should default to true -- otherwise, it can be overwritten.

.github/workflows/ci.yml
```diff
@@ -1089,21 +1108,14 @@ jobs:
     name: Update AUR package
     runs-on: ubuntu-latest
     needs: [combine]
-    if: github.event_name == 'release' &&
+    if: (github.event_name == 'release' || startsWith(github.event.inputs.fix-release, 'v')) &&
```
kasperl, line 1111: Why is this different?

.github/workflows/ci.yml
floitsch: Changed to "do-release"

.github/workflows/ci.yml

.github/workflows/ci.yml
floitsch: wondered the same thing.
Changed it.

.github/workflows/ci.yml
```diff
@@ -52,6 +52,15 @@ on:
         required: false
         type: boolean
         default: true
+      do-release:
+        description: "Build release artifacts and upload to this version"
```
kasperl, line 56:
```suggestion
        description: "Build release artifacts and upload with this version"
```

kasperl (approved):
Looks good to me.

PR #2568, opened 2024-10-04
Switch from "qemu" format to "full-image" format.

floitsch:
The full-image is also working for wokwi and for flashing.

### DIFF
--- a/tools/firmware.toit
+++ b/tools/firmware.toit
@@ -603,7 +603,7 @@ extract-cmd -> cli.Command:
           for debugging.
         - ubjson: a UBJSON encoding of the sections of the image.
         - image: a binary image suitable for running on QEMU,
-          wokwi, or for flashing the entire flash. See below.
+          Wokwi, or for flashing a device. See below.
         - qemu: a deprecated alias for 'image'.
         For host:
         - tar: a tar ball with a bash script to run the extracted firmware.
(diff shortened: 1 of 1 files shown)


tools/firmware.toit
```diff
@@ -602,13 +602,18 @@ extract-cmd -> cli.Command:
         - elf: the ELF file of the executable. This is typically used
           for debugging.
         - ubjson: a UBJSON encoding of the sections of the image.
-        - qemu: a full binary image suitable for running on QEMU.
+        - image: a binary image suitable for running on QEMU,
+          wokwi, or for flashing the entire flash. See below.
```
kasperl, line 606:
```suggestion
          Wokwi, or for flashing a device. See below.
```

tools/firmware.toit
```diff
@@ -678,8 +693,10 @@ extract-esp32 invocation/cli.Invocation envelope/Envelope --config-encoded/ByteA
     write-file output-path --ui=ui: it.write firmware-bin
     return
 
-  if format == "qemu":
-    write-qemu_ output-path firmware-bin envelope --ui=ui
+  if format == "qemu" or format == "image":
+    if format == "qemu":
+      ui.emit --warning "qemu format is deprecated, use 'image' instead"
```
kasperl, line 698:
```suggestion
      ui.emit --warning "The 'qemu' format is deprecated, use 'image' instead."
```

kasperl (approved):
Looks good to me.

PR #2569, opened 2024-10-06
Check the commands when asserts are on.

floitsch:


tools/toit.toit
```diff
@@ -500,6 +500,10 @@ main args/List:
 
   root-command.add system-message.build-command
 
+  assert:
+    // Only run the check in assert mode.
+    root-command.check
```
kasperl, line 505: What does check do? Maybe extend the comment to explain it a tiny bit, because assert, check, assert, and check doesn't tell the reader much here.

kasperl (approved):
Looks good to me.

tools/toit.toit
floitsch: done.

PR #2571, opened 2024-10-07
Support creating toitdocs for older packages.

floitsch:
We didn't require to have the name of a package in the package.yaml file.

### DIFF
--- a/tools/toitdoc/toitdoc.toit
+++ b/tools/toitdoc/toitdoc.toit
@@ -241,9 +241,25 @@ toitdoc invocation/cli.Invocation --toitc/string --sdk-path/string? --output/str
     if not file.is-file package-yaml-path:
       ui.abort "No package.yaml found at $package-yaml-path."
     content := yaml.decode (file.read-content package-yaml-path)
-    if content is not Map or not content.contains "name":
+    if content is Map:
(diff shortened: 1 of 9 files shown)


tools/toitdoc/toitdoc.toit
```diff
@@ -241,9 +241,25 @@ toitdoc invocation/cli.Invocation --toitc/string --sdk-path/string? --output/str
...
+      readme-path := fs.join source "README.md"
+      if file.is-file readme-path:
+        readme := (file.read-content readme-path).to-string
+        new-line-pos := readme.index-of "\n"
+        first-line/string := ?
+        if new-line-pos == -1:
+          first-line = readme
+        else:
+          first-line = readme[..new-line-pos]
+        if first-line.starts-with "#":
```
kasperl, line 259: Maybe skip all #s by looking for the first whitespace character?

kasperl (approved):
Looks good to me.

tools/toitdoc/toitdoc.toit
floitsch: There is no guarantee that there is a whitespace (`#foo`), but added code to skip over all `#` and added a check that the title isn't "".

PR #2576, opened 2024-10-09
Fix NetworkService.SELECTOR assert

kasperl:


floitsch (approved):
LGTM!

lib/system/api/network.toit
```diff
@@ -110,65 +111,8 @@ interface NetworkService:
   socket-mtu handle/int -> int
   static SOCKET-MTU-INDEX /int ::= 306
 
-class NetworkServiceClient extends ServiceClient implements NetworkService:
+class NetworkServiceClient extends NetworkServiceClientBase:
```
kasperl, line 114: This is only used from lib/net/net.toit and a single test now. All other uses have been changed to `NetworkServiceClientBase`.

PR #2580, opened 2024-10-16
Make it more obvious when we are hitting a 'new'.

floitsch:


src/top.cc
```diff
@@ -105,7 +105,7 @@ void* operator new[](size_t size) {
   // pointer exception when allocation fails.  Use _new which will call the
   // nothrow version instead, which skips the constructor when allocation
   // fails.
-  if (!toit::throwing_new_allowed) UNREACHABLE();
+  if (!toit::throwing_new_allowed) FATAL("Throwing new[] not allowed: %zu\n", size);
```
kasperl, line 108:
```suggestion
  if (!toit::throwing_new_allowed) FATAL("Throwing new[] not allowed: %zu", size);
```

kasperl (approved):
Looks good to me.

PR #2582, opened 2024-10-18
Fix examples that were using camelcase in set-timezone.

floitsch:


kasperl (approved):
Looks good, but that was never camelCase :)

PR #2585, opened 2024-10-21
Refactor the epoll event source so it can be subclassed.

floitsch:


src/event_sources/epoll_linux.cc
```diff
@@ -86,39 +80,35 @@ EpollEventSource::EpollEventSource() : EventSource("Epoll")
     FATAL("failed to register close fd: %d\n", errno);
   }
 
-  spawn();
+  return spawn();
 }
 
-EpollEventSource::~EpollEventSource() {
+void EpollEventSourceBase::stop() {
   close(control_write_);
```
kasperl, line 87: Maybe reset `control_write_` and `epoll_fd_`? You could even check that they have been reset in `start`.

kasperl (approved):
Looks good to me.

PR #2588, opened 2024-10-21
Minor cleanups for SPI.

floitsch:


src/resources/spi_linux.cc
```diff
@@ -46,21 +46,11 @@ PRIMITIVE(open) {
     close(fd);
     return Primitive::os_error(errno, process);
   }
-  result = ioctl(fd, SPI_IOC_RD_MAX_SPEED_HZ, &frequency);
```
kasperl, line 49: Not entirely clear why these two RD settings went away. Maybe add a comment to explain why we don't mess with them?

kasperl (approved):
Looks good to me.

src/resources/spi_linux.cc
floitsch: The `SPI_IOC_RD_MAX_SPEED_HZ` is not a setting, but a way to read the current MAX_SPEED_HZ (basically a setter).
added a comment:
```
// "WR"ite the max speed and mode.
```

PR #2590, opened 2024-10-21
Support keeping SPI.cs-active on Linux.

floitsch:


lib/spi.toit
```diff
@@ -281,6 +295,9 @@ class DevicePath_ extends DeviceBase_:
 
   resource_/ByteArray? := null
   state_/monitor.ResourceState_?
+  // We can't enforce that the bus is enforced, but the `keep-cs-active` is only
```
kasperl, line 298: can't enforce that the bus is enforced?

floitsch: done.

PR #2592, opened 2024-10-21
Add hardware tests.

floitsch:


tests/hw/shared/gpio-open-drain.toit
```diff
@@ -0,0 +1,79 @@
+// Copyright (C) 2024 Toitware ApS.
+// Use of this source code is governed by a Zero-Clause BSD license that can
+// be found in the tests/LICENSE file.
+
+/**
+Tests changing the open-drain setting of gpio pins.
+
+Setup:
+Connect test-pin and level-pin with a 330 Ohm (or any other 300-1K) resistor.
```
kasperl, line 9: Any particular reason you're not using $test-pin, etc. in here?

kasperl (approved):
Looks good to me.

tests/hw/shared/gpio-open-drain.toit
floitsch: The comment was originally a lib toitdoc without access to the arguments. Now that there is a function it makes sense to move the toitdoc down and use `$foo` references.
done.

PR #2596, opened 2024-10-21
Remove warning when compiling with clang.

floitsch:


src/ar.cc
```diff
@@ -23,7 +23,7 @@ namespace toit {
 namespace ar {
 
 static const char* const AR_HEADER = "!<arch>\x0A";
```
kasperl, line 25: Is it not possible to share the string constant?

kasperl (approved):
Looks good to me.

PR #2598, opened 2024-10-23
Allow to set the BLE GAP name.

floitsch:


lib/ble/ble.toit
```diff
@@ -252,9 +252,14 @@ class Adapter extends Resource_:
     reconnects more efficient.
 
   If $secure-connections is true then the peripheral is enabling secure connections.
+
+  If $name is provided it is used for the GAP name. The GAP name can be different from the
```
kasperl, line 256:
```suggestion
  If $name is provided, it is used for the GAP name. The GAP name can be different from the
```

lib/ble/ble.toit
```diff
@@ -252,9 +252,14 @@ class Adapter extends Resource_:
...
+    advertised name in the advertisement data. On some platforms the GAP name is stored
```
kasperl, line 257:
```suggestion
    advertised name in the advertisement data. On some platforms, the GAP name is stored
```

src/compiler/propagation/type_primitive_ble.cc
```diff
@@ -61,6 +61,7 @@ TYPE_PRIMITIVE_ANY(get_bonded_peers)
 TYPE_PRIMITIVE_ANY(toit_callback_init)
 TYPE_PRIMITIVE_ANY(toit_callback_deinit)
 TYPE_PRIMITIVE_ANY(toit_callback_reply)
+TYPE_PRIMITIVE_ANY(set_gap_device_name)
```
kasperl, line 64: Always returns null.

kasperl (approved):
Looks good to me.

lib/ble/ble.toit
floitsch: done.

lib/ble/ble.toit
floitsch: done.

src/compiler/propagation/type_primitive_ble.cc
floitsch: done.

PR #2599, opened 2024-10-23
Add support for changing the hostname

floitsch:


lib/system/system.toit
```diff
@@ -247,3 +253,28 @@ If the program is run as an executable, this is the fully resolved path to the
...
+Sets the hostname of the machine running the program.
+
+This operation is not supported on all platforms.
+
+Only new network connections will use the new hostname. Also, some
+  routers may cache the old hostname for a while.
+*/
+hostname= hostname/string -> none:
+  if platform != PLATFORM-FREERTOS:
+    throw "Setting hostname is not supported on this platform"
```
kasperl, line 276: We typically use much shorter exceptions in the core libraries. 

```suggestion
    throw "UNSUPPORTED"
```

src/primitive_core.cc
```diff
@@ -2571,4 +2577,36 @@ PRIMITIVE(rtc_user_bytes) {
...
+  // The HOST_NAME_MAX variable isn't defined on all platforms, so we
+  // just hardcode a good default. On my machine HOST_NAME_MAX is set to 64.
+  // So 256 should be enough.
+  char buffer[256];
+  int result = gethostname(buffer, sizeof(buffer));
+  if (result != 0) {
+    return Primitive::os_error(errno, process);
+  }
+  buffer[sizeof(buffer) - 1] = '\0';
+  // On macos, the hostname might include the .local suffix, which we don't want.
```
kasperl, line 2601:
```suggestion
  // On macOS, the hostname might include the .local suffix, which we don't want.
```

src/compiler/propagation/type_primitive_ethernet.cc
```diff
@@ -27,6 +27,7 @@ TYPE_PRIMITIVE_ANY(connect)
 TYPE_PRIMITIVE_ANY(setup_ip)
 TYPE_PRIMITIVE_ANY(disconnect)
 TYPE_PRIMITIVE_ANY(get_ip)
+TYPE_PRIMITIVE_ANY(set_hostname)
```
kasperl, line 30: Always returns null?

src/compiler/propagation/type_primitive_wifi.cc
```diff
@@ -31,6 +31,7 @@ TYPE_PRIMITIVE_ANY(get_ip)
 TYPE_PRIMITIVE_ANY(init_scan)
 TYPE_PRIMITIVE_ANY(start_scan)
 TYPE_PRIMITIVE_ANY(read_scan)
+TYPE_PRIMITIVE_ANY(set_hostname)
```
kasperl, line 34: Always returns null?

kasperl (approved):
Looks good to me.

lib/system/system.toit
floitsch: done.

src/compiler/propagation/type_primitive_ethernet.cc
floitsch: done.

src/compiler/propagation/type_primitive_wifi.cc
floitsch: done.

src/primitive_core.cc

PR #2601, opened 2024-10-24
Don't accept strings with '\0' for C operations.

floitsch:


src/primitive_core.cc
```diff
@@ -69,15 +69,23 @@ namespace toit {
...
   return process->null_object();
 }
 
 PRIMITIVE(write_string_on_stderr) {
-  ARGS(cstring, message, bool, add_newline);
-  fprintf(stderr, "%s%s", message, add_newline ? "\n" : "");
+  ARGS(Blob, message, bool, add_newline);
+  auto bytes = message.address();
+  for (int i = 0; i < message.length(); i++) {
+    putchar(bytes[i]);
```
kasperl, line 86: How is this stderr? This looks like a pretty significant regression.

I'd take something like my suggestion above and maybe put it in a static helper function parameterized with the stream.

Also, consider making both member variables of `Blob` const like:

```
 private:
  const uint8* const address_;
  const word length_;
```

src/primitive_core.cc
```diff
@@ -69,15 +69,23 @@ namespace toit {
...
 
 PRIMITIVE(write_string_on_stdout) {
-  ARGS(cstring, message, bool, add_newline);
-  fprintf(stdout, "%s%s", message, add_newline ? "\n" : "");
+  ARGS(Blob, message, bool, add_newline);
+  auto bytes = message.address();
+  for (int i = 0; i < message.length(); i++) {
+    putchar(bytes[i]);
+  }
+  if (add_newline) putchar('\n');
```
kasperl, line 77:
```suggestion
  flockfile(stdout);
  for (int i = 0; i < message.length(); i++) {
    putc_unlocked(bytes[i], stdout);
  }
  if (add_newline) putc_unlocked('\n', stdout);
  funlockfile(stdout);
```

src/primitive_core.cc
floitsch: Good catch...
I will update the blob fields in a separate PR.

floitsch:
Good feedback.

src/primitive_core.cc
```diff
@@ -68,17 +68,24 @@ namespace toit {
...
+  write_to(stdout, message.address(), message.length());
   return process->null_object();
 }
 
 PRIMITIVE(write_string_on_stderr) {
-  ARGS(cstring, message, bool, add_newline);
-  fprintf(stderr, "%s%s", message, add_newline ? "\n" : "");
-  fflush(stderr);
+  ARGS(Blob, message, bool, add_newline);
+  write_to(stderr, message.address(), message.length());
```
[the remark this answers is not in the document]

src/primitive_core.cc
```diff
@@ -68,17 +72,88 @@ namespace toit {
...
+    WriteFile(console, buffer.data(), (DWORD)buffer.size(), &written, NULL);
+
+    if (newline) {
+      WriteFile(console, "\r\n", 2, &written, NULL);
+    }
+  }
+
+  return process->null_object();
+}
+#elif (_POSIX_C_SOURCE >= 199309L || _BSD_SOURCE) && defined(_POSIX_THREAD_SAFE_FUNCTIONS)
```
kasperl, line 120:
```suggestion

#elif (_POSIX_C_SOURCE >= 199309L || _BSD_SOURCE) && defined(_POSIX_THREAD_SAFE_FUNCTIONS)

```

src/primitive_core.cc
```diff
@@ -68,17 +72,88 @@ namespace toit {
...
+  fwrite_unlocked(bytes, 1, length, stream);
+  if (newline) {
+    fputc_unlocked('\n', stream);
+  } else {
+    fflush_unlocked(stream);
+  }
+  funlockfile(stream);
+  return process->null_object();
+}
+#else
```
kasperl, line 133:
```suggestion

#else

```

src/primitive_core.cc
```diff
@@ -68,17 +72,88 @@ namespace toit {
...
+  FILE* stream = is_stdout ? stdout : stderr;
+  fwrite(bytes, 1, length, stream);
+  if (newline) {
+    fputc('\n', stream);
+  } else {
+    fflush(stream);
+  }
+  return process->null_object();
+}
+#endif
```
kasperl, line 144:
```suggestion

#endif
```

src/primitive_core.cc
```diff
@@ -68,17 +72,88 @@ namespace toit {
 
 MODULE_IMPLEMENTATION(core, MODULE_CORE)
 
-PRIMITIVE(write_string_on_stdout) {
-  ARGS(cstring, message, bool, add_newline);
-  fprintf(stdout, "%s%s", message, add_newline ? "\n" : "");
-  fflush(stdout);
+#if defined(TOIT_WINDOWS)
```
kasperl, line 75:
```suggestion
#if defined(TOIT_WINDOWS)

```

src/primitive_core.cc
```diff
@@ -68,17 +72,88 @@ namespace toit {
...
+#include <windows.h>
+#include <vector>
+
+static void replace_line_endings(const uint8_t* bytes, size_t length, std::vector<uint8_t>& buffer) {
```
kasperl, line 79: I'd probably use `uint8` instead of the _t variants in here.

```suggestion
static void replace_line_endings(const uint8* bytes, size_t length, std::vector<uint8>& buffer) {
```

kasperl (approved):
Looks good to me.

src/primitive_core.cc
floitsch: done.

src/primitive_core.cc

src/primitive_core.cc
floitsch: done.

src/primitive_core.cc
floitsch: done.

src/primitive_core.cc
floitsch: done.

PR #2611, opened 2024-11-07
Force comparisons to be boolean.

floitsch:
A `x == y` or `foo < bar` is now guaranteed to return a boolean type.


kasperl (approved):
Looks good to me.

PR #2616, opened 2024-11-08
Fix external tests.

floitsch:
Also be more consistent how we do conditions in bash.


floitsch:
@coderabbitai review

kasperl (approved):
Looks good to me.

PR #2630, opened 2024-11-12
Remove the gpio library.

floitsch:
The pin.toit library is moved to its location in a follow-up PR.


kasperl (approved):
Looks good to me.

PR #2635, opened 2024-11-13
Drop self-interface.

floitsch:
Without this fix toitdocs state that interfaces implement themselves.


kasperl (approved):
Looks good to me.

PR #2637, opened 2024-11-14
Double the timeouts if the optimization override is set.

floitsch:
Currently, the nightly buildbot run fails because of a timeout with -O0 (on macos).


kasperl (approved):
Looks good to me.

PR #2640, opened 2024-11-21
Make sure to flush when using `print`.

floitsch:


src/primitive_core.cc
```diff
@@ -122,6 +122,9 @@ static Object* write_on_std(const uint8* bytes, size_t length, bool is_stdout, b
     }
   }
 
+  FILE* stream = is_stdout ? stdout : stderr;
```
kasperl, line 125: Are you sure we need this here? I don't like the mixing of win32 and std APIs here. Not sure if FlushFileBuffers would work.

kasperl (approved):
Looks good to me.

PR #2642, opened 2024-11-24
Support active BLE scans.

floitsch:
Active scans send a request to the device when they have received an advertisement.


lib/ble/remote.toit
```diff
@@ -50,14 +50,19 @@ class Central extends Resource_:
 
   Only one scan can run at a time.
 
+  If $active is true, then we request a scan response from discovered devices.
+    Users might need to merge the advertisement data from the scan response with the
```
kasperl, line 54: Unless this is quite unlikely, I think 'may' would be better.

```suggestion
    Users may need to merge the advertisement data from the scan response with the
```

kasperl (approved):
Looks good to me.

lib/ble/remote.toit
floitsch: done.

PR #2648, opened 2024-11-26
Fix bad 'i2c.write-reg' call.

### LINKED ISSUE #2646: i2c.toit function:  write-reg register bytes wrong calling
/lib/i2c.toit in class Device   
write-reg register/int bytes/ByteArray: in line 198 calling wrong function.
Instead of calling:   write bytes: throw it 
It should call:   write-reg register  bytes: throw it

floitsch:
Fixes #2646.

### DIFF
--- a/lib/i2c.toit
+++ b/lib/i2c.toit
@@ -196,7 +196,7 @@ class Device implements serial.Device:
     and then calling $(write bytes).
   */
   write-reg register/int bytes/ByteArray:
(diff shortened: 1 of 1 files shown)


floitsch:
Unfortunately no tests. We are not yet set up to test the i2c functionality. (Hopefully soon).

kasperl (approved):
Looks good to me.

PR #2652, opened 2024-11-29
Set the MBEDTLS_ALLOW_PRIVATE_ACCESS flag all the time.

floitsch:
This removes the warnings we get with link-time optimizations.
It's unfortunately also exposing all of the privates of mbedtls.

Ideally we would like to use `private_foo` members, so it's clear where we use non-public functionality. However, we need to import some private headers and these must have `MBEDTLS_ALLOW_PRIVATE_ACCESS` defined, which makes it impossible (or very hard) for us not to set it.

The difference between setting `MBEDTLS_ALLOW_PRIVATE_ACCESS` and not:
Without:
```
typedef struct mbedtls_entropy_context {
    mbedtls_md_context_t private_accumulator;
    int private_accumulator_started;
    int private_source_count;
    mbedtls_entropy_source_state private_source[20];
}
mbedtls_entropy_context;
```
With:
```
typedef struct mbedtls_entropy_context {
    mbedtls_md_context_t accumulator;
    int accumulator_started;
    int source_count;
    mbedtls_entropy_source_state source[20];
}
mbedtls_entropy_context;
```

### DIFF
--- a/src/aes.h
+++ b/src/aes.h
@@ -17,6 +17,7 @@
 
 #include "top.h"
 
+#define MBEDTLS_ALLOW_PRIVATE_ACCESS
 #include <mbedtls/aes.h>
 
 #include "resource.h"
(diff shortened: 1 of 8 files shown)


kasperl (approved):
Looks good to me.

PR #2664, opened 2024-12-09
Roll esp-idf.

floitsch:
This patch-head contains the fixes for esp32c6 (where the malloc lock
    wasn't set).


kasperl (approved):
Looks good to me.

PR #2678, opened 2024-12-13
Fix PWM for ESP variants.

floitsch:
The ESP32 worked fine, but other variants use different clock speeds.

### DIFF
--- a/src/resources/pwm_esp32.cc
+++ b/src/resources/pwm_esp32.cc
@@ -134,6 +134,8 @@ PRIMITIVE(init) {
                                                &src_clk_frequency);
   if (err != ESP_OK) return Primitive::os_error(err, process);
 
+  // The max frequency is half the source clock frequency. At that frequency there are
+  // only three duty-factors left: 0%, 50% and 100%.
   if (frequency <= 0 || frequency > max_frequency || max_frequency > (src_clk_frequency >> 1)) {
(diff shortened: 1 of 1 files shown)


src/resources/pwm_esp32.cc
```diff
@@ -119,19 +123,23 @@ class PwmResourceGroup : public ResourceGroup {
...
   ARGS(int, frequency, int, max_frequency)
 
-  if (frequency <= 0 || frequency > max_frequency || max_frequency > kMaxFrequency) FAIL(OUT_OF_BOUNDS);
+  uint32 src_clk_frequency = 0;
+  esp_err_t err = esp_clk_tree_src_get_freq_hz(static_cast<soc_module_clk_t>(kDefaultClk),
+                                               ESP_CLK_TREE_SRC_FREQ_PRECISION_EXACT,
+                                               &src_clk_frequency);
+  if (err != ESP_OK) return Primitive::os_error(err, process);
+
+  if (frequency <= 0 || frequency > max_frequency || max_frequency > (src_clk_frequency >> 1)) {
```
kasperl, line 137: Why do we shift `src_clk_frequency`? Add a comment?

kasperl (approved):
Looks good to me!

src/resources/pwm_esp32.cc
floitsch: done.

PR #2684, opened 2025-01-02
Add dhtxx to external repos.

floitsch:
Also don't complain if an external directory doesn't exist. During the first `make enable-external` it's normal that they don't exist.


kasperl (approved):
Looks good to me.

PR #2686, opened 2025-01-02
Print stacktrace when a test fails.

floitsch:


tests/hw/esp-tester/tester.toit
```diff
@@ -156,6 +171,7 @@ class TestDevice:
     if port:
       port.close
       port = null
+    directory.rmdir --recursive tmp-dir
```
kasperl, line 174: Will this work if done a second time? I guess so.

kasperl (approved):
Looks good to me.

tests/hw/esp-tester/tester.toit
floitsch: not yet...
I added a check.
done.

PR #2690, opened 2025-01-08
Add ESP32S3 tests

floitsch:


tests/hw/esp32/variants.toit
```diff
@@ -0,0 +1,443 @@
...
+  The $wait-for-close-pin should be connected to GND with an 1MOhm resistor.
+  */
+  wait-for-close-pin -> int
+
+VARIANTS ::= {
+  system.ARCHITECTURE-ESP32: Esp32,
+  system.ARCHITECTURE-ESP32S3: Esp32s3,
+}
+
+abstract class VariantBase:
```
kasperl, line 230: Does this implement `Variant`? Is it really necessary with both an abstract class and the separate interface?

kasperl (approved):
Looks good to me!

tests/hw/esp32/variants.toit
floitsch: simplified.
done.

PR #2695, opened 2025-01-16
Add Makefile rules for esp32c3,c6,s2 and s3.

floitsch:


kasperl (approved):
Looks good to me.

Convenient!

PR #2698, opened 2025-01-29
Refactor the ADC to use the new API.

floitsch:


lib/gpio/adc.toit
```diff
@@ -10,22 +10,47 @@ Analog-to-Digital Conversion.
...
   print adc.get
   adc.close
+  pin.close
 ```
+
+# Pins
+
+## ESP32
+ADC1: Pins 32-39
+ADC2: Pins 0, 2, 4, 12-15, 25-27.
```
kasperl, line 35:
```suggestion
ADC2: Pins 0, 2, 4, 12-15, 25-27
```

lib/gpio/adc.toit
```diff
@@ -44,17 +69,14 @@ class Adc:
...
-    strapping pins). By default, ADC2 is disabled, and users need to pass in the
+  # ESP32 and variants.
+  On most ESP32 variants, there are two ADCs. Typically, ADC1 should be
+    preferred as ADC2 has lots of restrictions. It can't be
+    used when WiFi is active, and some of the pins are strapping pins.
+  By default, ADC2 is disabled, and users need to pass in the
     $allow-restricted flag to allow its use.
   */
-  constructor .pin --max-voltage/float?=null --allow-restricted/bool=false:
+  constructor .pin --max-voltage/float?=null --allow-restricted/bool?=false:
```
kasperl, line 79: Why is `allow-restricted` suddenly allowed to be `null`?

src/resources/adc_esp32.cc
```diff
@@ -34,68 +35,74 @@ namespace toit {
...
   }
 }
 
 static int get_adc2_channel(int pin) {
-  switch (pin) {
-    case 5: return ADC2_CHANNEL_0;
-    default: return adc2_channel_t(-1);
-  }
+  // On ESP32C3, ADC2 is no longer supported, due to its HW limitation.
+  // There was an errata on the espressif website.
```
kasperl, line 87:
```suggestion
  // There was an errata on the Espressif website.
```

src/resources/adc_esp32.cc
```diff
@@ -106,203 +113,340 @@ static int get_adc2_channel(int pin) {
...
+      calibration_deinit(calibration);
+    }
+  }
 
-  adc_unit_t unit;
-  int chan;
-#ifndef CONFIG_IDF_TARGET_ESP32C6
-  esp_adc_cal_characteristics_t calibration;
-#endif
+  adc_oneshot_unit_handle_t* unit;
```
kasperl, line 336: Maybe make these private?

src/resources/i2s_esp32.cc
```diff
@@ -15,7 +15,7 @@
 
 #include "../top.h"
 
-#ifdef TOIT_ESP32
+#if 0 // def TOIT_ESP32
```
kasperl, line 18: Intentional?

kasperl (approved):
Looks good to me.

src/resources/adc_esp32.cc
floitsch: done.

src/resources/i2s_esp32.cc
floitsch: yes.
The I2S uses the ADC underneath. It's not allowed to use the old and new version of the ADC driver. Until the I2S code is updated it won't compile.

There is already a follow-up PR that fixes the I2S. (Otherwise, I would have added more comments).

PR #2701, opened 2025-01-30
Only run one CI per branch.

floitsch:
On the master, the current workflow is allowed to finish, but intermediate runs might be skipped.
On branches, the latest push cancels existing runs.


kasperl (approved):
Looks good to me.

PR #2708, opened 2025-02-05
Fix uart-io-data test.

floitsch:


tests/hw/esp32/uart-io-data-shared.toit
```diff
@@ -50,6 +50,7 @@ main-board1:
 
 test-board1:
   port := uart.Port --rx=(gpio.Pin RX) --tx=null --baud-rate=BAUD-RATE
```
kasperl, line 52:
```suggestion
  port := uart.Port 
      --rx=(gpio.Pin RX) 
      --tx=null 
      --baud-rate=BAUD-RATE
```

kasperl (approved):
Looks good to me.

PR #2710, opened 2025-02-05
Prepare test devices in parallel.

floitsch:
Also prefix each line with the name of the test-device.


floitsch:
Looks like the current raspberry can't handle both compilations at the same time. Moved it back and added a TODO.

PR #2716, opened 2025-02-25
Add support for stringifying/printing unsigned int64s.

floitsch:


lib/core/numbers.toit
```diff
@@ -954,6 +954,17 @@ abstract class int extends num:
   stringify radix/int -> string:
     #primitive.core.int64-to-string
 
+  /**
+  Variant of $(stringify).
+
+  Treats the number as an unsigned 64-bit integer.
+  */
+  stringify --unsigned-64/True -> string:
```
kasperl, line 962: Don't we use the int64 abbreviation in other places? I think it would be logical if we called this `stringify --uint64`.

lib/core/numbers.toit
```diff
@@ -954,6 +954,17 @@ abstract class int extends num:
...
+    return stringify-unsigned-64_ this
+
+  static stringify-unsigned-64_ number/int -> string:
```
kasperl, line 965: stringify-uint64_

kasperl (approved):
Looks good to me.

lib/core/numbers.toit
floitsch: I initially only had `stringify --unsigned`, and then decided to mention the 64.
Going with your suggestion now.
done.

lib/core/numbers.toit
floitsch: done.

PR #2724, opened 2025-03-06
Scan infinitely without a scan-duration on darwin.

floitsch:


src/resources/ble_darwin.mm
```diff
@@ -840,7 +840,11 @@ - (id)initWithResource:(toit::BleResource*)resource {
 
   AsyncThread::run_async([=]() -> void {
     LightLocker locker(central_manager->scan_mutex());
-    OS::wait_us(central_manager->stop_scan_condition(), duration_us);
+    if (duration_us > 0) {
```
kasperl, line 843:
```suggestion
    if (duration_us >= 0) {
```

kasperl (approved):
Looks good to me.

src/resources/ble_darwin.mm
floitsch: done.

PR #2727, opened 2025-03-07
Allow to connect to TLS without certificate verification.

floitsch:


lib/tls/session.toit
```diff
@@ -121,6 +121,7 @@ class Session:
   certificate/Certificate?
   root-certificates/List
   handshake-timeout/Duration
+  skip-certificate-verification/bool
```
kasperl, line 124: I think this should be validation, not verification.

See https://datatracker.ietf.org/doc/html/rfc4949 under "validate vs. verify" on page 330.

src/resources/tls.cc
```diff
@@ -197,6 +197,10 @@ void BaseMbedTlsSocket::apply_certs(Process* process) {
   }
 }
 
+void BaseMbedTlsSocket::disable_certificate_verification() {
```
kasperl, line 200: validation - but I guess this is where the verification comes from :)

tests/tls-server-cert-test.toit
```diff
@@ -0,0 +1,118 @@
+// Copyright (C) 2019 Toitware ApS.
```
kasperl, line 1: Update

kasperl (approved):
Looks good to me.

lib/tls/session.toit
floitsch: done.

src/resources/tls.cc
floitsch: done.

tests/tls-server-cert-test.toit
floitsch: done.

PR #2733, opened 2025-03-07
Add git-tag message.

floitsch:
Otherwise git interactively asks for the tag message.


kasperl (approved):
Looks good to me.

floitsch:
Abandoning the PR.
What happened: I have a `tag.forceSignAnnotated` enabled on my git. However that requires tags with messages (so they are commits).
The test thus failed. However, we actually prefer lightweight tags. I will fix this differently.

PR #2735, opened 2025-03-07
Make the esp32 tester more efficient

floitsch:
Install the tests in parallel on the two boards (if the test uses both).


tests/hw/esp-tester/tester.toit
```diff
@@ -19,7 +19,13 @@ import .shared
 ALL-TESTS-DONE ::= "All tests done"
 JAG-DECODE ::= "jag decode"
 
+start-time/Time := ?
+
+log message/string:
+  print "--- $(%06d (Duration.since start-time).in-ms): $message"
+
 main args:
+  start-time = Time.now
```
kasperl, line 28: It might be better to use a monotonic clock here. You really don't want this to change if the real time clock changes.

kasperl (approved):
Looks good to me.

tests/hw/esp-tester/tester.toit
floitsch: done.

PR #2738, opened 2025-03-11
Rewrite pulse-counter driver.

floitsch:


lib/pulse_counter.toit
```diff
@@ -122,57 +139,39 @@ class Unit:
...
       if glitch-filter-ns <= 0: throw "INVALID_ARGUMENT"
       // The glitch filter runs on the APB clock (80MHz, 12.5ns), and allows at most 1023 ticks.
       // 12.5 * 1023 == 12787.5.
       if glitch-filter-ns > 12_787: throw "OUT_OF_RANGE"
     else:
-      glitch-filter-ns = -1
+      glitch-filter-ns = 0
+    if low and low >= 0: throw "INVALID_ARGUMENT"
+    if high and high <= 0: throw "INVALID_ARGUMENT"
+    if not low: low = 0
```
kasperl, line 157: So passing 0 is the same as passing null? That feels a bit odd.

kasperl (approved):
Looks good to me.

lib/pulse_counter.toit
kasperl: Where do you check that the user cannot pass in 0?

PR #2747, opened 2025-03-24
Add a getter for the level of a logger.

addshore:
Usecases:
 - Making a new logger or logging related service
   that is the same level as the current default
   logger.
 - Using the level of the default logger to decide
   at runtime how to construct logged strings.
   eg. If log level is debug or higher, I might
   include clickable URLs to further tools or
   resources.

### DIFF
--- a/lib/log/log.toit
+++ b/lib/log/log.toit
@@ -152,7 +152,7 @@ class Logger:
       keys_ = keys
       values_ = values
 
-  /** Returns the level of the logger. */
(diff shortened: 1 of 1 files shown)


addshore:
Happy to sign the CLA if needed

floitsch:
Thanks.
@kasperl  can you help @addshore sign the agreement?
It's not necessary for this PR, but might as well get it done :)

PR #2755, opened 2025-03-27
Press reset for longer.

floitsch:
We have the total-runtime test that fails on the esp32s3.
Maybe pressing the reset button for longer helps.


kasperl (approved):
Looks good to me.

PR #2764, opened 2025-04-14
Fix warnings in spi_flash_esp32.cc

floitsch:
There are still two warnings left, but they are from macros we import.
They might already be fixed on an updated version of the esp-idf.


kasperl (approved):
Looks good to me.

PR #2772, opened 2025-04-23
Add 'set-pull' to GPIO.

floitsch:


lib/gpio/gpio.toit
```diff
@@ -381,6 +381,15 @@ interface Pin:
   */
   set-open-drain value/bool
 
+  /**
```
kasperl, line 384: Is this the most natural interface?

Maybe consider changing the name to something like `set-pull-direction`? At a first glance, `set-pull` smells like a yes or no thing like `set-open-drain`. Not sure if it is just me :)

kasperl (approved):
Looks good to me.

PR #2781, opened 2025-04-29
Drop ifs with two dead blocks

kasperl:


tests/type_propagation/gold/default-arguments-test.gold-O2
```diff
@@ -84,6 +84,4 @@ test-non-default tests/type_propagation/default-arguments-test.toit
 test-non-default-non-literal tests/type_propagation/default-arguments-test.toit
  - argument 0: {True|False}
   0[016] - load local 2
```
kasperl, line 86: Not entirely sure why this local doesn't go away. Checking.

tests/type_propagation/gold/default-test.gold-O2
```diff
@@ -8,18 +8,12 @@ main tests/type_propagation/default-test.toit
...
-  1[084] - branch if not null T4
-  4[016] - load local 2
-  5[083] - branch if false T8
-  8[092] - return null S0 1
+  0[092] - return null S0 1
 
 foo2 tests/type_propagation/default-test.toit
  - argument 0: {False}
   0[016] - load local 2
   1[084] - branch if not null T8
```
kasperl, line 16: We still need to capture the propagated types for identical and branch-if-null, so we can eliminate this branch (we know that false is not null).

tests/type_propagation/gold/default-arguments-test.gold-O2
kasperl: Oh, wait. We're returning the local. It's the load local 2 from position 4 that sticks around. All is well :)

PR #2783, opened 2025-04-30
Improve the RMT Signals class.

floitsch:
Adds support for ns values.

### DIFF
--- a/lib/rmt.toit
+++ b/lib/rmt.toit
@@ -64,7 +64,7 @@ class Signals:
   /**
   The resolution of the signals in Hz.
 
-  Null, if this instance was created without a resolution.
+  If this instance was created without a resolution, $resolution is null.
   */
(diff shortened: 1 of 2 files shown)


lib/rmt.toit
```diff
@@ -58,16 +61,22 @@ class Signals:
   /** The number of signals in the collection. */
   size/int
 
-  bytes_/ByteArray
+  /**
+  The resolution of the signals in Hz.
 
-  /** The empty signal collection. */
-  static ZERO ::= Signals 0
+  Null, if this instance was created without a resolution.
```
kasperl, line 67: I think I'd reword this to avoid capitalizing `null`. 

```suggestion
  If this instance was created without a resolution, $resolution is null.
```

kasperl (approved):
Looks good to me.

PR #2792, opened 2025-05-06
Deprecate Zero Signals constant.

floitsch:


lib/rmt.toit
```diff
@@ -50,7 +50,11 @@ At the lower level, a signal consists of 16 bits: 15 bits for the period and 1
   to be divisible by 4.
 */
 class Signals:
-  /** An empty signals instance. */
+  /**
+  An empty signals instance.
+
+  Deprecated. Signals should not be empty.
```
kasperl, line 56: Didn't you just add this? I guess we didn't release anything with `Signals.ZERO` support in it.

kasperl (approved):
Looks good to me.

lib/rmt.toit
floitsch: It was already there, but in a different place (a bit lower).

https://libs.toit.io/rmt/class-Signals#ZERO(0%2C0%2C0%2C)

PR #2804, opened 2025-05-08
Add esp-idf license.

floitsch:
There isn't much code from the esp-idf, but we do use the one in the mbedtls directory.


kasperl (approved):
Looks good to me.

PR #2816, opened 2025-05-14
Fix Windows tests.

floitsch:


tests/strip/utils.toit
```diff
@@ -25,17 +27,24 @@ backticks-failing args/List -> string:
   process.stdin.close
 
   // We are merging stdout and stderr into one stream.
+  stdout-done := monitor.Latch
```
kasperl, line 30: An alternative would be a single semaphore.

kasperl (approved):
Looks good to me.

tests/strip/utils.toit
floitsch: Agreed. I thought of the same thing.
But I had already started with the latches, and they read nicer.

PR #2818, opened 2025-05-15
Use 'toit' as argument for external tests.

floitsch:
They used to get 'toit.run', but nothing outside the Toit repository should use that.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -290,7 +290,7 @@ function(add_repo_tests REPO)
 
     add_test(
       NAME "${TEST_EXPECTATION_NAME}"
-      COMMAND $<TARGET_FILE:toit.run> "${FILE}" $<TARGET_FILE:toit.run>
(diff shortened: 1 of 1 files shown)


external/CMakeLists.txt
```diff
@@ -290,7 +290,7 @@ function(add_repo_tests REPO)
 
     add_test(
       NAME "${TEST_EXPECTATION_NAME}"
-      COMMAND $<TARGET_FILE:toit.run> "${FILE}" $<TARGET_FILE:toit.run>
+      COMMAND $<TARGET_FILE:toit.run> "${FILE}" $<TARGET_FILE:toit>
```
kasperl, line 293: So we still use `toit.run` for running the test, but pass `toit` to it?

kasperl (approved):
Looks good to me.

external/CMakeLists.txt
floitsch: Correct.
We could use `toit` for running the tests as well, but inside the Toit repository I'm fine with using the individual tools.

PR #2825, opened 2025-05-21
Fix memory leak in list_toit_directory_entries.

floitsch:


src/compiler/lsp/completion.cc
```diff
@@ -369,14 +369,14 @@ void CompletionHandler::import_path(const char* path,
                                     Filesystem* fs) {
   if (is_first_segment) {
     current_package.list_prefixes([&](const std::string& candidate) {
-      complete(candidate.c_str(), CompletionKind::MODULE);
+      complete(candidate, CompletionKind::MODULE);
```
floitsch, line 372: Changes in this file don't change memory behavior. They are just cleaner this way.

kasperl (approved):
Looks good to me!

PR #2827, opened 2025-05-21
Give warning if import segments don't match the case of the path.

floitsch:


src/compiler/compiler.cc
```diff
@@ -966,24 +967,81 @@ enum class AddSegmentResult {
...
+                                    bool should_check_is_toit_file,
+                                    const Source::Range& error_range,
+                                    Diagnostics* diagnostics) {
+  char* dir_path = path_builder->strdup();
+  Defer free_dir_path { [&] { free(dir_path); } };
+
+  // Checks that the casing of the found path is actually correct.
+  // Assumes that the file was found.
+  // On Linux this is usually not an issue, but might happen with mounted filesystems that are
+  // not case-sensitive?
```
kasperl, line 979:
```suggestion
  // On Linux this is usually not an issue, but it might happen with mounted filesystems that are
  // not case-sensitive.
```

src/compiler/compiler.cc
```diff
@@ -1220,7 +1278,9 @@ Source* Pipeline::_load_import(ast::Unit* unit,
     auto result = add_segment(&import_path_builder,
                               next_segment,
                               filesystem(),
-                              true);  // Must be a toit file.
+                              true,  // Must be a toit file.
+                              segments[segments.length() -1]->selection_range(),
```
kasperl, line 1282:
```suggestion
                              segments[segments.length() - 1]->selection_range(),
```

tests/negative/fail.cmake
```diff
@@ -20,3 +20,10 @@ list(APPEND TOIT_OPTIMIZATION_SKIP_TESTS
   # The following tests only work with standard optimizations.
   tests/negative/field-type5-test.toit
 )
+
+# The following test is only relevent on Windows and macOS.
```
kasperl, line 24:
```suggestion
# The following test is only relevant on Windows and macOS.
```

kasperl (approved):
Looks good to me.

PR #2837, opened 2025-05-23
Pessimize local types in the presence of non-local branches.

kasperl:


src/compiler/propagation/type_scope.h
```diff
@@ -37,6 +37,7 @@ class TypeScope {
   }
 
   int level() const { return level_; }
+  int level_linked() const { return level_linked_; }
```
floitsch, line 40: keep?

floitsch (approved):
LGTM.

src/compiler/propagation/type_scope.h
kasperl: Keep? I made this public :)

PR #2844, opened 2025-05-26
Add custom envelope dispatch.

floitsch:


floitsch:
I haven't checked whether the dispatch token is valid for that repository. If it isn't we will get a red build during the next release. However, it should only affect this workflow.

kasperl (approved):
Looks good to me.

PR #2847, opened 2025-05-26
Fix toitdoc generation for the SDK.

floitsch:


tools/toitdoc/toitdoc.toit
```diff
@@ -193,15 +193,13 @@ compute-sdk-path --sdk-path/string? --toitc/string? --ui/Ui -> string:
...
-
-  // Try first with the directory of the toitc.
-  sdk-path = fs.dirname toitc
-  lib-dir := fs.join sdk-path "lib"
-  if file.is-directory lib-dir: return sdk-path
-  // Try "..".
-  sdk-path = fs.join sdk-path ".."
-  lib-dir = fs.join sdk-path "lib"
-  if file.is-directory lib-dir: return sdk-path
+  lib-bin-dir := fs.dirname toitc
```
kasperl, line 196: So this is the path to `toit.compile` which is in `lib/toit/bin/`, so you go back to `lib/toit/lib/`.

kasperl (approved):
Looks good to me.

tools/toitdoc/toitdoc.toit
floitsch: Correct.

PR #2857, opened 2025-05-28
Remove symlink from repository.

floitsch:
The tarball was *sometimes* not extractable on Windows, as the symlink target appeared before the symlink.


kasperl (commented):
Looks good.

PR #2862, opened 2025-05-29
Make toitdoc use 'toit' as compiler.

floitsch:


tools/toit.toit
```diff
@@ -532,6 +536,15 @@ tool-path sdk-dir/string? tool/string -> string:
 
   return fs.join tool-bin-dir tool
 
+toit-path sdk-dir/string? -> string:
+  // Use ourself as the toitc command.
```
kasperl:
```suggestion
  // Use ourself as the toit command.
```

kasperl (approved):
Looks good to me.

PR #2865, opened 2025-06-02
Fix gnueabi cross compilation.

floitsch:
The cmake CACHE variables are only set the first time, unless there is a `FORCE`.


toolchains/arm.cmake
```diff
@@ -67,7 +67,7 @@ set(CMAKE_CXX_COMPILER_TARGET "${ARM_TARGET}" CACHE STRING "c++ compiler target
 set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -x assembler-with-cpp" CACHE STRING "asm flags")
 
 set(CMAKE_EXE_LINKER_FLAGS_INIT "${CMAKE_EXE_LINKER_FLAGS_INIT} -fuse-ld=lld"
```
kasperl, line 70: Maybe explain why `FORCE` is necessary in a comment?

kasperl (approved):
Looks good to me.

toolchains/arm.cmake
floitsch: done.

PR #2871, opened 2025-06-02
Use the cli-pkg's test classes.

floitsch:


tests/pkg/gold-tester.toit
```diff
@@ -134,34 +134,34 @@ class GoldTester:
...
+        --create-stderr
         toit-exec_
         full-args
-    stdout := fork-data[1]
-    stderr := fork-data[2]
-    child-process := fork-data[3]
+    stdout := process.stdout
+    stderr := process.stderr
 
     stdout-data := #[]
```
kasperl, line 146: This looks like something that could be rewritten with `Task.group`.

kasperl (approved):
Looks good to me.

tests/pkg/gold-tester.toit
floitsch: Leaving for now.
It might be possible, but the exact conditions we have here are not easily matched to `Task.group`. I believe we don't need them, but it would take a bit of time to test.

PR #2876, opened 2025-06-03
Fix rs485 test.

floitsch:
The 1st board would finish the program and shut down the UART. However,
    at that time there could have still been data in the buffers.


kasperl (approved):
Looks good to me.

PR #2877, opened 2025-06-03
Don't run wireless tests at the same time.

floitsch:
Also run wireless tests up to three times.


tests/hw/esp32/CMakeLists.txt
```diff
@@ -96,11 +96,36 @@ function(parse_args_from_file file_path result_var)
   set(${result_var} ${result_list} PARENT_SCOPE)
 endfunction()
 
+function(parse_wireless_from_file file_path result_var)
```
kasperl, line 99: It is a bit weird that you map from WIRELESS to flaky. Maybe make the category hw-wireless? You do more than just retries in the sense that you also separate them from the other tests.

PR #2878, opened 2025-06-03
Run the compilation and wait-for-connect of esp tests in parallel.

floitsch:


tests/hw/esp-tester/tester.toit
```diff
@@ -124,19 +124,31 @@ run-test invocation/cli.Invocation:
...
       board1-ready.set true
 
     if port-board2:
       board2 = TestDevice --name="board2" --port-path=port-board2 --ui=ui --toit-exe=toit-exe
-      log "Connecting to board2"
-      board2.connect-network
+      image2/ByteArray? := null
+      Task.group [
+        ::
+          log "Connecting to board2"
```
kasperl: Maybe move the logging to `connect-network`? The `TestDevice` already knows its name and the `Task.group` call would look even better.

kasperl (approved):
Looks good to me.

tests/hw/esp-tester/tester.toit
floitsch: done.

PR #2879, opened 2025-06-03
Linux SPI fixes.

floitsch:


src/event_sources/async_posix.cc
```diff
@@ -56,11 +56,14 @@ void AsyncEventThread::entry() {
     auto func = element->func;
     delete element;
     state_ = RUNNING;
+    word result;
```
kasperl, line 59: Maybe add a comment to explain why you want to be in the IDLE state before calling `on_event`?

src/resources/spi_linux.cc
```diff
@@ -120,7 +113,14 @@ Object* SpiResource::transfer_start(Blob data, int from, int length,
   // TODO(florian): allow to neuter incoming external byte arrays.
   uint8* buffer = unvoid_cast<uint8*>(malloc(length));
   if (buffer == null) FAIL(MALLOC_FAILED);
-  Defer free_buffer{ [&] { if (!successfully_dispatched) free(buffer); } };
+  buffer_ = buffer;
+  buffer_size_ = length;
+  Defer free_buffer{ [&] {
+    if (!successfully_dispatched) {
+      buffer_ = null;
```
kasperl: This makes absolutely no sense.

src/event_sources/async_posix.cc
floitsch: done.

src/resources/spi_linux.cc
floitsch: done.

PR #2880, opened 2025-06-04
Fix GPIO pin, when using input, output, and initial value.

floitsch:


src/resources/gpio_esp32.cc
```diff
@@ -294,14 +294,7 @@ PRIMITIVE(config) {
     .intr_type = GPIO_INTR_DISABLE,
   };
 
-  if (input) {
-    cfg.intr_type = GPIO_INTR_ANYEDGE;
```
kasperl, line 298: You dropped the `intr_type` change?

kasperl (approved):
Looks good to me.

src/resources/gpio_esp32.cc
floitsch: it's still there. in line 313.
The main change is, that the `output` is now run before and sets the output value.
Before we had the problem that input+output would only enter the `input` branch and then not set the output.

I separated the concerns a bit better, with a section for input, one for output, and then one for setting the cfg.mode.

PR #2884, opened 2025-06-08
Fix windows unicode issues.

floitsch:


src/main_utf_8_helper.cc
```diff
@@ -0,0 +1,64 @@
...
+  // Since the callee may change the argv, we keep a copy of the UTF-8 arguments.
+  char** utf_8_args = toit::unvoid_cast<char**>(malloc(sizeof(char*) * (wargc + 1)));
+  char** utf_8_copy = toit::unvoid_cast<char**>(malloc(sizeof(char*) * (wargc + 1)));
+  utf_8_args[wargc] = null;
+  for (int i = 0; i < wargc; ++i) {
+    uword total_len = toit::Utils::utf_16_to_8(wargv[i], wcslen(wargv[i]));
+    char* utf_8_arg = toit::unvoid_cast<char*>(malloc(total_len + 1));  // +1 for null terminator.
+    toit::Utils::utf_16_to_8(wargv[i], wcslen(wargv[i]), reinterpret_cast<uint8*>(utf_8_arg), total_len);
+    utf_8_arg[total_len] = '\0';  // Null-terminate the string.
+    utf_8_args[i] = utf_8_arg;  // Keep a copy.
```
kasperl: Isn't the copy in the next line? Maybe move the comment?

Maybe explain why you need the copy? I guess you're worried that argv is changed by the callee?

kasperl (approved):
Looks good to me.

src/main_utf_8_helper.cc
floitsch: Correct. We change the argv in our executables.
done.

PR #2885, opened 2025-06-08
Accept she-bang Toit files that don't have 'toit' extension.

floitsch:


tools/toit.toit
```diff
@@ -31,6 +31,22 @@ import .toitdoc as toitdoc
 import .lsp.server.server as lsp
 import .snapshot as snapshot-lib
 
+/**
+Whether the given $path is a Toit source that can be run with `toit run`.
+Also accepts snapshots.
+
+Might return true for files that might not be valid Toit source files.
```
kasperl:
```suggestion
Might return true for files that aren't valid Toit source files.
```

kasperl (approved):
Looks good to me.

PR #2887, opened 2025-06-17
Only escape strings if they are quoted.

floitsch:


lib/encoding/yaml/encoder.toit
```diff
@@ -51,7 +51,7 @@ class YamlEncoder extends EncoderBase_:
                     str.contains "\r" or
                     not (parse --on-error=(: null) str) is string
 
-    escaped := escape-string str
+    escaped := should-quote ? escape-string str : str
```
kasperl: Maybe quicker to parse for us humans?

```suggestion
    escaped := should-quote ? (escape-string str) : str
```

PR #2904, opened 2025-08-21
Add TRACE log level

addshore:
TRACE is a commonly accepted log level these days.
So allow users to use it directly.
These logs will not be logged by defualt, as the
default level is DEBUG.


floitsch:
Sorry for the late reply.
The reason we are hesitating is that this would be a breaking change for some applications.
For example, the cellular modem drivers use integers to configure the log-level. (The log-level is encoded in JSON and therefore can't use the actual constants).

addshore:
Gotcha,
`-1` could also be an option?

PR #2905, opened 2025-08-25
Add support for updating mtime and atime of files.

floitsch:


src/primitive_file_non_win.cc
```diff
@@ -404,6 +404,44 @@ PRIMITIVE(stat) {
...
+    if (mtime_sec != 0) FAIL(INVALID_ARGUMENT);
+    mtime_nsec = UTIME_OMIT;
+  }
+
+  struct timespec times[2];
+  times[0].tv_sec = atime_sec;
+  times[0].tv_nsec = atime_nsec;
+  times[1].tv_sec = mtime_sec;
+  times[1].tv_nsec = mtime_nsec;
+  if (utimensat(AT_FDCWD, path, times, 0) == -1) {
```
kasperl: Nit: I'd probably do a `!= 0` compare.

src/primitive_file_win.cc
```diff
@@ -354,8 +354,8 @@ int64 low_high_to_int64(DWORD high, DWORD low) {
 #define WINDOWS_TICKS_PER_SECOND 10000000
 #define SEC_TO_UNIX_EPOCH 11644473600LL
 
-Object* time_stamp(Process* process, FILETIME* time) {
-  int64 windows_ticks = low_high_to_int64(time->dwHighDateTime, time->dwLowDateTime);
+Object* time_stamp(Process* process, LARGE_INTEGER* time) {
```
kasperl: Why isn't this a `const LARGE_INTEGER&`? I think this is what you use below `file_time`.

kasperl (approved):
Looks good to me.

src/primitive_file_non_win.cc
floitsch: done.

src/primitive_file_win.cc
floitsch: I just changed the type, and didn't think about the `const` reference.
done.

PR #2906, opened 2025-08-27
Make the master key optional for ESP-NOW services.

floitsch:
The key could already be null, but it was required to provide it.


kasperl (approved):
Looks good to me.

PR #2909, opened 2025-08-27
Improve espnow address class.

floitsch:


lib/esp32/espnow.toit
```diff
@@ -203,9 +203,42 @@ class Address:
...
 
+  /** Variant of $(parse str [--on-error]) that throws on errors. */
+  static parse str/string -> Address:
+    return Address.parse str --on-error=: throw it
+
+  /**
+  Parses the given $str as MAC address.
+
+  The $on-error block is called when the $str is not a valid MAC address. The
+    result of the block is then returned.
```
kasperl: Maybe you should add that it must be an Address or null?

floitsch: done.

PR #2910, opened 2025-08-27
Use 'WiFi' instead of 'Wi-Fi'.

floitsch:
In our code base and documentation we tend to use 'WiFi' and not 'Wi-Fi'.


kasperl (approved):
Looks good to me.

PR #2911, opened 2025-08-29
Use the VM version when calling 'toit version'.

floitsch:
Normally, the app and vm version of the 'toit' executable should be the same, but during development, the 'toit' snapshot is sometimes not rebuilt. In that case we want to report the VM version.


kasperl (approved):
Looks good to me.

PR #2914, opened 2025-08-31
Support large espnow messages.

floitsch:


lib/esp32/espnow.toit
```diff
@@ -363,14 +363,40 @@ class Service:
     keys is unencrypted even if a master $key is provided.
 
   The $channel parameter must be a valid WiFi channel number.
+
+  # Advanced
+  The $buffer-byte-size parameter sets the size of the internal buffer used
```
kasperl, line 369: Maybe mention the `% 128` size requirements?

src/resources/espnow_esp32.cc
```diff
@@ -53,70 +44,139 @@ class SpinLocker {
...
-    return null;
+
+    int index = (head_ + used_) % queue_size_;
+    Datagram* datagram = &datagrams_[index];
+    int offset = used_ > 0
+        ? (newest->offset + newest->len) % buffer_size_
+        : 0;
+    datagram->offset = offset;
+    datagram->len = data_size;
+    // Copy in two steps to handle wrap-around.
```
kasperl, line 120: I wonder if we could use the ring buffer we have in the UART implementation somehow?

src/resources/espnow_esp32.cc
```diff
@@ -131,49 +191,60 @@ enum class EspNowEvent {
...
+// the queue, we will never add a receive event, as there is already one in the queue.
+const int kEventQueueSize = 3;
 static QueueHandle_t event_queue;
 
 // This function is registered as callback and will then be called on the high-priority WiFi task.
 static void espnow_send_cb(const uint8* mac_addr, esp_now_send_status_t status) {
   tx_status = status;
   auto event = EspNowEvent::SEND_DONE;
-  auto ret = xQueueSend(event_queue, &event, 0);
+  auto ret = xQueueSendFromISR(event_queue, &event, 0);
```
kasperl: Are we sure this is called from an ISR or just on the Wifi task?

src/resources/espnow_esp32.cc
```diff
@@ -131,49 +191,60 @@ enum class EspNowEvent {
...
+
+  // Always keep at least one slot for a "send-done" event in the queue.
+  // We just need *one* event in the queue for receive events. Since the
+  // queue is bigger than 2 elements, and there is never more than one
+  // send-done event, we just need to check that there is more than one
+  // slot left.
+  static_assert(kEventQueueSize >= 3, "Unexpected event queue size");
+  if (uxQueueSpacesAvailable(event_queue) > 1) {
+    auto event = EspNowEvent::NEW_DATA_AVAILABLE;
+    portBASE_TYPE ret = xQueueSendFromISR(event_queue, &event, 0);
```
kasperl: Are we sure this is called from an ISR or just on the Wifi task?

src/resources/espnow_esp32.cc
```diff
@@ -489,43 +555,36 @@ PRIMITIVE(send_succeeded) {
...
-    ESP_LOGE("ESPNow", "Didn't get peeked queue entry");
-    return process->null_object();
+    bool success = datagram_pool->consume(peeked, ByteArray::Bytes(data).address());
+    if (!success) {
+      // The oldest datagram was discarded to make space for a new one.
+      peeked = datagram_pool->peek();
+      if (!peeked.is_valid()) FATAL("Expected valid datagram");
+      continue;
+    }
+    break;
```
kasperl: It almost feels like you should move the return path up here.

lib/esp32/espnow.toit
floitsch: done.

src/resources/espnow_esp32.cc
floitsch: It looks pretty similar. I will add it to my TODO list to investigate whether we can easily share them.

src/resources/espnow_esp32.cc
floitsch: you are right. Only Wifi task.

src/resources/espnow_esp32.cc
floitsch: ditto.
done.

src/resources/espnow_esp32.cc
floitsch: done.

PR #2916, opened 2025-09-07
Allow multiple services/characteristics with the same uuid.

floitsch:
In theory it could be possible to also allow duplicate descriptors under each characteristic, but that's just asking for troubles.

Also make it nicer to add a read-only descriptor (which is the most common one).

### DIFF
--- a/lib/ble/local.toit
+++ b/lib/ble/local.toit
@@ -167,7 +167,6 @@ class LocalService extends Resource_ implements Attribute:
 
   constructor .peripheral-manager .uuid:
     if peripheral-manager.deployed_: throw "Peripheral is already deployed"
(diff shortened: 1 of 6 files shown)


lib/ble/local.toit
```diff
@@ -557,6 +555,18 @@ class LocalCharacteristic extends LocalReadWriteElement_ implements Attribute:
     descriptors_.add descriptor
     return descriptor
 
+  /**
+  Adds a read-only descriptor to this characteristic.
+  $uuid is the uuid of the descriptor
```
kasperl: That second line is a slightly weird sentence with no punctuation.

```suggestion
  Adds a read-only descriptor to this characteristic with the given $uuid.
```

kasperl (approved):
Looks good to me.

PR #2922, opened 2025-09-14
Back off when BLE notifications run out of memory.

floitsch:
Instead of throwing, allow notifications to try again if they couldn't allocate memory.


kasperl (approved):
Looks good to me.

PR #2923, opened 2025-09-14
Add test that we can read the dhtxx without sleeping.

floitsch:
The sensor is only updating every 1 (DHT11) or 2 (DHT22) seconds. If we query too fast it doesn't even respond. The driver should handle that.


kasperl (approved):
Looks good to me.

PR #2926, opened 2025-09-21
Allow to parse timestamps without seconds.

### LINKED ISSUE #2920: Parsing time without seconds using `Time.parse`
Accessing [OpenMeteo API](https://open-meteo.com), my query come back with a date value like: `2025-09-09T04:30`
- without seconds
- without "Z" or "+00:00" at the end of the string.

The parser will continue without the Z/timezone but not without me adding at least `:00` (seconds component).  Without regexp (yet) to do some trustworthy checking/correcting of the values using code, I'm hoping the parser could be improved slightly to cope when a date doesn't have seconds.

As an aside:  The API's I'm using also work with 'unixtime'.  I wonder if it would be easy to import/export these values also?

Thanks team, for bringing this awesomeness into the world!!

floitsch:
Fixes #2920.

### DIFF
--- a/tests/time-test.toit
+++ b/tests/time-test.toit
@@ -190,7 +190,6 @@ simple-constructor-test:
 examples-test:
   EXAMPLES.do: test-time-string it
 
-
 duration-test:
   d / Duration := Duration --h=1 --m=2 --s=3 --ms=4 --us=5 --ns=6
   expect-equals 3723004005006 d.in-ns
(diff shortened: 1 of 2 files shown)


tests/time-test.toit
```diff
@@ -179,6 +190,7 @@ simple-constructor-test:
 examples-test:
   EXAMPLES.do: test-time-string it
 
+
```
kasperl: Maybe drop this again?

kasperl (approved):
Looks good to me.

tests/time-test.toit
floitsch: done.

PR #2930, opened 2025-10-11
Disable UDP broadcast test on macos.

floitsch:


.github/workflows/ci.yml
```diff
@@ -349,6 +349,11 @@ jobs:
         run: |
           strip -s build/host/sdk/bin/*
 
+      - name: Configure network for UDP broadcast (macOS)
+        # Needed for the UDP broadcast test to work.
```
kasperl, line 353: So this claims the test will work, but we still disable it?

floitsch: This was an attempt to fix the issue, but it didn't actually solve it.
Will remove it.

floitsch: https://github.com/toitlang/toit/pull/2939

PR #2932, opened 2025-10-18
Roll ESP-IDF to 5.4.2.

floitsch:


mbedtls/include/default_config.h
```diff
@@ -1622,6 +1622,46 @@
  */
 //#define MBEDTLS_SSL_ASYNC_PRIVATE
 
+/** \def MBEDTLS_SSL_CLI_ALLOW_WEAK_CERTIFICATE_VERIFICATION_WITHOUT_HOSTNAME
```
floitsch, line 1625: As per the README in the mbedtls/include directory:
The default_config.h file has been updated to match the one in the mbedtls directory.

The toit_config.h was updated to match these changes. I just applied the same diff.
Unless I screwed it up, the patch to default_config.h should be exactly the same as for toit_config.h. And that diff doesn't actually change anything.

floitsch:
TBR.

PR #2933, opened 2025-10-18
Fix the drom-map-end.

floitsch:
It was changed in https://github.com/espressif/esp-idf/commit/7c1d9a5813217b0a27de5114eeff7997fd577149.


kasperl (commented):
Looks good.

When did this change for the s3?

PR #2934, opened 2025-10-19
Add support for the esp32p4

floitsch:


src/os_esp32.cc
```diff
@@ -425,6 +429,9 @@ OS::HeapMemoryRange OS::get_heap_memory_range() {
 #ifdef CONFIG_IDF_TARGET_ESP32S3
   range.address = reinterpret_cast<void*>(0x3fca0000);
   range.size = 384 * KB;
+#elif CONFIG_IDF_TARGET_ESP32P4
```
floitsch, line 432: I need some help here. I have no idea what I'm doing...
Also, I think that the C3, C6, and S2 might need a different value too?

floitsch:
With the default configuration the esp32-p4 crashes *immediately*.
If I disable cmpct-malloc, it manages to run a bit.
However, It has a tendency to run out of memory:
```
Heap report @ out of memory in 'stack grow' 216:324:
  ┌───────────┬──────────┬─────────────────────────────────────────────────────┐
  │   Bytes   │  Count   │  Type                                               │
  ├───────────┼──────────┼─────────────────────────────────────────────────────┤
  └───────────┴──────────┴─────────────────────────────────────────────────────┘
  Total: 0 bytes in 0 allocations (0%), largest free 376k, total free 439k
```
It looks like there is memory, but Toit doesn't recognize it.

src/os_esp32.cc
floitsch: Should be correct now.

floitsch:
TBR.

PR #2943, opened 2025-11-07
Fix Deque.reserve.

floitsch:


lib/core/collections.toit
```diff
@@ -3307,15 +3307,19 @@ A deque is a generalization of a stack and a queue, and can be used for both
...
+  // be equal to the size of the backing list, but that can be changed to
+  // reserve space.
+  // Using a growable list, gives an extra level of indirection, but the same
+  // amortized big-O performance, and much less code, avoiding all the
   // complications around wrapping around the end of the backing.  If we
   // determine that the collection is used very frequently and requires better
   // performance, we can reimplement with the a more traditional
   // implementation.
   first_ := 0
+  last_ := 0
```
kasperl: Nit: It is a little bit weird that first is inclusive and last isn't, just from a naming perspective. Maybe start/end would be nicer?

lib/core/collections.toit
```diff
@@ -3415,8 +3432,15 @@ class Deque extends List implements Collection:
...
-    result := backing_.remove-last
+    last := last_
+    backing := backing_
+    if first_ == last: throw "OUT_OF_BOUNDS"
+    result := ?
+    if last == backing.size:
+      result = backing.remove-last
+    else:
+      result = backing[last - 1]
+    last_ = last - 1
```
kasperl: You should probably null out the removed element in the backing store, like it is done for `remove-first`.

lib/core/collections.toit
```diff
@@ -3514,9 +3552,11 @@ class Deque extends List implements Collection:
   */
   remove --at/int -> any:
     last := size - 1
```
kasperl, line 3557: This is a different last than `last_`.

lib/core/collections.toit
```diff
@@ -3514,9 +3552,11 @@ class Deque extends List implements Collection:
...
-    if at < 0 or at > last: throw "OUT_OF_BOUNDS"
-    if at >= last >> 1:
-      return backing_.remove --at=(first_ + at)
+    if not 0 <= at <= last: throw "OUT_OF_BOUNDS"
+    if at >= (last >> 1):
```
kasperl, line 3564: Add a comment here?

lib/core/collections.toit
```diff
@@ -3544,15 +3584,20 @@ class Deque extends List implements Collection:
...
   shrink-if-needed_ -> none:
     backing := backing_
     first := first_
-    if first * 2 > backing.size:
-      backing.replace 0 backing first first + size
-      backing.resize size
+    last := last_
+    reserved := backing.size - last
+    sz := last - first
+    if first * 2 > last:
```
kasperl: Add a comment here? Why do we multiply here and shift in the other place?

lib/core/collections.toit
```diff
@@ -3544,15 +3584,20 @@ class Deque extends List implements Collection:
...
 
   shrink-if-needed_ -> none:
     backing := backing_
     first := first_
-    if first * 2 > backing.size:
-      backing.replace 0 backing first first + size
-      backing.resize size
+    last := last_
+    reserved := backing.size - last
+    sz := last - first
```
kasperl: Why do we compute this eagerly? Why isn't this called `size`?

tests/deque-test.toit
```diff
@@ -211,3 +213,376 @@ test-copy:
   d2.remove-first
   expect-equals "[1, 42]" d.stringify
   expect-equals "[103, 3]" d2.stringify
+
+
```
kasperl: Drop one newline?

lib/core/collections.toit
floitsch: done.

lib/core/collections.toit
floitsch: good catch.
done.

lib/core/collections.toit
floitsch: done. And refactored it to make it easier to follow.

lib/core/collections.toit
floitsch: somehow this file started to copy fields into locals (probably for efficiency reasons), and I have continued that "tradition".
It can't be called "size", as that would shadow the method, which isn't allowed.

lib/core/collections.toit
floitsch: switched to shift for consistency.
done.

tests/deque-test.toit
floitsch: done.

PR #2945, opened 2025-11-15
Port list-registry test

floitsch:


tools/pkg/commands/list.toit
```diff
@@ -74,46 +77,63 @@ class ListCommand extends PkgCommand:
 
   If $allow-extra-fields, then allows dependencies and environment keys too.
   */
-  static verbose-description description/Description --allow-extra-fields=false -> Map:
+  static verbose-description description/Description --allow-extra-fields/bool=false -> Map:
     filtered := description.content.filter: | k _ |
```
kasperl: I'd prefer key over k.

tools/pkg/commands/list.toit
```diff
@@ -74,46 +77,63 @@ class ListCommand extends PkgCommand:
...
+      a.name.compare-to b.name --if-equal=:
+        a.version.compare-to b.version --if-equal=:
+          a.ref-hash.compare-to b.ref-hash
+
+    if ui.wants-structured:
+      ui.emit-list --result descriptions
+      return
+
+    if ui.wants-human:
+      if descriptions.is-empty:
```
kasperl: Maybe combine with `and`?

tools/pkg/commands/list.toit
floitsch: done.

tools/pkg/commands/list.toit
floitsch: done.

PR #2949, opened 2025-11-15
Fix kebabify on Windows.

floitsch:
When a path had a `\` in it, the kebabify tool would fail.

### DIFF
--- a/tools/kebabify.toit
+++ b/tools/kebabify.toit
@@ -141,6 +141,8 @@ migrate invocation/cli.Invocation toitc/string?:
 
   ui.emit --info "Migrating $point-count locations."
   joined := migration-points.join ","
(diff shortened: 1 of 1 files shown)


tools/kebabify.toit
```diff
@@ -140,7 +140,9 @@ migrate invocation/cli.Invocation toitc/string?:
   migration-points.resize point-count
 
   ui.emit --info "Migrating $point-count locations."
-  parsed-points := json.parse "[$(migration-points.join ",")]"
+  joined := migration-points.join ","
+  joined = joined.replace --all "\\" "\\\\"
```
kasperl, line 146: Maybe explain why you need to do this in a comment?

floitsch: done.

PR #2954, opened 2025-12-07
Fix RMT on esp32c3.

floitsch:


src/resources/rmt_esp32.cc
```diff
@@ -610,7 +610,13 @@ IRAM_ATTR static size_t encoder_callback(const void* data,
     }
     // The current sequence fits.
     // Copy it over and update the active instance.
-    memcpy(symbols, sequence_bytes, sequence_length);
+    // The original 'memcpy' didn't work on esp32c3. We use a loop instead.
```
kasperl: Maybe explain the theory of why the loop works when memcpy doesn't?

We can't use memcpy here, because ...

src/resources/rmt_esp32.cc
```diff
@@ -610,7 +610,13 @@ IRAM_ATTR static size_t encoder_callback(const void* data,
...
+    // memcpy(symbols, sequence_bytes, sequence_length);
```
kasperl: Not sure these two comment lines help much.

src/resources/rmt_esp32.cc
floitsch: done.

src/resources/rmt_esp32.cc
floitsch: done.

PR #2957, opened 2025-12-11
Update firmware extraction format descriptions

addshore:
Clarified usage instructions for firmware extraction formats, specifically binary vs image, mentioning esptool in both


floitsch (approved):
LGTM.
Thanks.

PR #3018, opened 2026-04-06
Fix ESP32 RSA watchdog timeout during key generation

valientegaston:
**Problem**
On ESP32, generating an RSA 2048-bit key takes longer than 10 seconds, causing the watchdog timer to kill the process, mistakenly believing it is hung.

**Root cause**
The mbedtls RSA operations (rsa_generate, rsa_sign, rsa_encrypt, rsa_decrypt) were passing NULL as the context to the rsa_rng callback. The scheduler watchdog monitors the run_timestamp of each process and kills it if it hasn't been updated within the timeout window. Since the RSA key generation is a long-running blocking operation, the timestamp was never updated during its execution.

**Fix**
The rsa_rng callback is called many times by mbedtls during RSA operations — in particular during key generation, it is called repeatedly to generate prime candidates and random bases for Miller-Rabin primality tests. This makes it a natural and low-overhead place to reset the watchdog by updating the process run_timestamp.
The fix passes the Process* as the context to all rsa_rng calls, and inside rsa_rng the scheduler mutex is taken to safely update the timestamp.

**Safety**
We verified that primitives execute inside an Unlocker block in Scheduler::run_process, meaning the scheduler mutex is not held during RSA operations. Therefore, taking it inside rsa_rng is safe and will not cause a deadlock.


valientegaston:
I also noticed that CONFIG_MBEDTLS_HARDWARE_MPI is explicitly set to n across all ESP32 toolchain configurations (esp32, esp32s2, esp32s3, esp32c3, esp32c6). Was this intentional? I understand it wouldn't help with key generation (as the prime generation and Miller-Rabin tests run in software regardless), but it could potentially speed up everyday RSA operations like signing, verification, and encryption/decryption.

valientegaston:
Hi! I have implemented the RSA key generation following your suggestions:

**Non-blocking**: The primitives are now non-blocking. The key generation runs in a separate thread, and the Toit Task is suspended while waiting for the resource event.

**RsaGeneration Resource**: I created a new RsaGeneration resource (contained in StructTag as RsaGenerationResourceTag and RsaGenerationResourceGroupTag) to manage the lifecycle of the asynchronous task.

**Architecture**: I followed the approach in async_posix.cc, using the AsyncEventThread to coordinate the background generation and the completion event.

**Flags & Configuration**: The new resource and primitives are correctly guarded by the CONFIG_TOIT_CRYPTO_EXTRA sdkconfig flag as you proposed. I've also updated the sdkconfig.defaults for ESP32 targets to enable this by default.

**Binary Compatibility**: I made sure to append the new tags at the end of the StructTag enum to avoid any issues with existing bootstrap snapshots.

Please let me know if there's any part of the implementation you'd like me to refine further. Thanks!

floitsch:
Opened a fix PR on your fork: https://github.com/valientegaston/toit/pull/1

It addresses a few resource cleanup issues:
- The resource group and state were not cleaned up on exceptions (e.g. `with-timeout` wrapping the `generate` call). Added a `RsaGenerationResource_` class with the standard finalizer/close pattern.
- Missing `resource_proxy->clear_external_address()` on error paths in `rsa_generate_finish` (dangling pointer).
- Missing `Thread::cancel()` on ESP32 (linker error since `async_posix` is now compiled for ESP32).

src/primitive_crypto.cc
```diff
@@ -870,6 +874,201 @@ static ByteArray* rsa_export_der(mbedtls_pk_context* pk, Process* process, bool
...
+        ret = MBEDTLS_ERR_PK_ALLOC_FAILED;
+      } else {
+        int prv_len = mbedtls_pk_write_key_der(&pk, prv, RSA_PRV_DER_MAX_BYTES);
+        int pub_len = mbedtls_pk_write_pubkey_der(&pk, pub, RSA_PUB_DER_MAX_BYTES);
+        if (prv_len < 0 || pub_len < 0) {
+          ret = prv_len < 0 ? prv_len : pub_len;
+          free(prv); free(pub);
+        } else {
+          // mbedtls writes from the end. Move to start.
+          unsigned char* prv_start = (unsigned char*)malloc(prv_len);
```
floitsch, line 999: If we don't have enough memory here, we would have to throw away everything we did so far and start again.
Also: I'm not even sure things would behave correctly: By marking the error as `MBEDTLS_ERR_PK_ALLOC_FAILED` the `rsa_generate_finish` would return a "malloc failed". This would indicate to the system that the primitive ran out of memory and that we should do a GC, then try again. However, the `rsa_generate_finish` would just fail now as the resource_proxy just has been cleared.

-> We need to allocate the memory in the main-thread (that has called the primitive) and not inside the event-thread. Unfortunately, this means that we need to know the prv_len und pub_len already before we actually call `mbedtls_pk_write_key_der`.

That said: it looks to me like a resize is the better option anyway. The `prv_len`/`pub_len` just seem to say how much of the given buffer was written too.

src/scheduler.h
```diff
@@ -147,6 +147,8 @@ class Scheduler {
   bool is_locked() const { return OS::is_locked(mutex_); }
   bool is_boot_process(Process* process) const { return boot_process_ == process; }
 
+  Mutex* mutex() const { return mutex_; }
```
floitsch, line 150: Necessary?
If yes, we should try to find a better approach than to make the scheduler's mutex public.

toolchains/esp32c3/sdkconfig.defaults
```diff
@@ -51,7 +51,7 @@ CONFIG_MBEDTLS_SSL_IN_CONTENT_LEN=7800
 CONFIG_MBEDTLS_SSL_OUT_CONTENT_LEN=3700
 CONFIG_MBEDTLS_X509_TRUSTED_CERT_CALLBACK=y
 CONFIG_MBEDTLS_ECP_RESTARTABLE=y
-CONFIG_MBEDTLS_HARDWARE_MPI=n
+CONFIG_MBEDTLS_HARDWARE_MPI=y
```
floitsch, line 54: Is there any downside to enabling MPI?

For some esp32 variants we were fighting the partition sizes already. So if this increases the image size, we need to make sure we aren't too big now.

floitsch:
@valientegaston do you think you can address the feedback?

valientegaston:
Hi! Yes, I'll be addressing the feedback in the coming days. We've been quite busy lately but I'll make sure to go through all the comments and make the necessary changes. Thanks for your patience!

src/primitive_crypto.cc
valientegaston: Replaced the two extra `malloc` calls with `memmove` on the same buffer to shift the data to the start, followed by `realloc` to shrink it to the exact size. If `realloc` fails it returns null without altering the original block, so the original pointer is preserved and passed to `set_results` correctly — no memory leak or dangling pointer.

src/scheduler.h
valientegaston: No longer necessary. In the previous approach, the `rsa_rng` callback needed to take the scheduler mutex to update the process `run_timestamp`. With the new async design, the Toit process is suspended during key generation, so there's no need to update the timestamp from the RNG callback at all. The `mutex()` accessor has been removed.

toolchains/esp32c3/sdkconfig.defaults
valientegaston: I measured the binary size impact on ESP32:
- Without `HARDWARE_MPI`: 0x1561f0 bytes (18% free of 0x1a0000 partition)
- With `HARDWARE_MPI`: 0x156390 bytes (18% free of 0x1a0000 partition)

The difference is only ~416 bytes. 

I only tested on ESP32 so far — still checking the other variants.

valientegaston: ESP32S3:

Without HARDWARE_MPI: 0x14d5c0 bytes (20% free)
With HARDWARE_MPI: 0x14d840 bytes (20% free)
Difference: ~640 bytes

Still checking the remaining variants (esp32c3, esp32c6, esp32s2).

valientegaston: ESP32S2:

Without HARDWARE_MPI: 0x116a50 bytes (33% free)
With HARDWARE_MPI: 0x116ce0 bytes (33% free)
Difference: ~656 bytes

valientegaston: ESP32C3:

Without HARDWARE_MPI: 0x163e00 bytes (14% free)
With HARDWARE_MPI: 0x164560 bytes (14% free)
Difference: ~1888 bytes


ESP32C6:

Without HARDWARE_MPI: 0x1818c0 bytes (11% free)
With HARDWARE_MPI: 0x182020 bytes (11% free)
Difference: ~1888 bytes

src/primitive_crypto.cc
```diff
@@ -870,6 +874,199 @@ static ByteArray* rsa_export_der(mbedtls_pk_context* pk, Process* process, bool
...
+    RsaGenerationResource* res = static_cast<RsaGenerationResource*>(r);
+    mbedtls_pk_context pk;
+    mbedtls_pk_init(&pk);
+    int ret = mbedtls_pk_setup(&pk, mbedtls_pk_info_from_type(MBEDTLS_PK_RSA));
+    if (ret == 0) {
+      ret = mbedtls_rsa_gen_key(mbedtls_pk_rsa(pk), rsa_rng, null, res->bits(), 65537);
+    }
+
+    if (ret == 0) {
+      unsigned char* prv = (unsigned char*)malloc(RSA_PRV_DER_MAX_BYTES);
```
floitsch, line 986: When do we free the memory again? Shouldn't that happen in rsa_generate_finish?
But that's more complicated:
- what if nothing ever calls 'rsa_generate_finish'? This means that the memory needs to be attached to the resource.
- However, that immediately reveals another race condition (which was already present): what if the program stops while we are running in parallel. The `res` resource could already be freed at that point.

src/primitive_crypto.cc
floitsch: The same still applies to the 'malloc' in line 986/987. (the ret goes to the rsa_generate_finish, leading to an unrecoverable OOM). The mallocs need to be done in the main thread. Here we just use constants (RSA_PRV_DER_MAX_BYTES), so that's relatively easy to fix.

src/primitive_crypto.cc
```diff
@@ -870,6 +874,199 @@ static ByteArray* rsa_export_der(mbedtls_pk_context* pk, Process* process, bool
...
+        int pub_len = mbedtls_pk_write_pubkey_der(&pk, pub, RSA_PUB_DER_MAX_BYTES);
+        if (prv_len < 0 || pub_len < 0) {
+          ret = prv_len < 0 ? prv_len : pub_len;
+          free(prv); free(pub);
+        } else {
+          // mbedtls writes from the end. Move to start.
+          memmove(prv, prv + RSA_PRV_DER_MAX_BYTES - prv_len, prv_len);
+          memmove(pub, pub + RSA_PUB_DER_MAX_BYTES - pub_len, pub_len);
+
+          unsigned char* prv_resized = (unsigned char*)realloc(prv, prv_len);
```
floitsch, line 1002: Maybe add the following information:
RSA_PRV_DER_MAX_BYTES is typically ~550 bytes.
Typically, keys only need ~300 bytes.
As such the realloc is useful.
However, if it fails, it's not the end of the world either.

src/primitive_crypto.cc
floitsch, line 986: Use `unvoid_cast` instead of C-style casts. Here and in all the other places.

src/primitive_crypto.cc
```diff
@@ -870,6 +874,199 @@ static ByteArray* rsa_export_der(mbedtls_pk_context* pk, Process* process, bool
...
+  ARGS(RsaGenerationResource, resource);
+  if (resource->error() != 0) {
+    int err = resource->error();
+    resource->resource_group()->unregister_resource(resource);
+    resource_proxy->clear_external_address();
+    return tls_error(null, process, err);
+  }
+
+  ByteArray* prv_der = process->allocate_byte_array(resource->prv_len());
+  ByteArray* pub_der = process->allocate_byte_array(resource->pub_len());
```
floitsch, line 1037: Another option is to use an external ByteArray and use the existing buffers. That would decrease the peak amount of memory.
I think for 250+ bytes it's ok to have an external byte-array.
However, at that point I would do another realloc, just in case the one in the other thread didn't succeed. If it still doesn't succeed, you can do a MALLOC_FAILED.

src/primitive_crypto.cc
```diff
@@ -870,6 +874,199 @@ static ByteArray* rsa_export_der(mbedtls_pk_context* pk, Process* process, bool
...
+  if (!prv_der || !pub_der) {
+    resource->resource_group()->unregister_resource(resource);
+    resource_proxy->clear_external_address();
```
floitsch, line 1040: That's a bit too eager.
For OOMs the primitive will do a GC and then try again. If you destroy the resource the second attempt wouldn't work anymore.

floitsch (commented):
Still some subtle issues.
Not completely sure what the best way for the resource handling is.
Maybe one way is to keep track of whether the parallel thread is running and store that in the resource. When the resource gets the signal to destroy itself, it could then delay that action until the thread returns. Alternatively (if possible) killing the thread first, would also work.
We do something slightly similar in the BLE code.

toolchains/esp32c3/sdkconfig.defaults
floitsch: thanks.

src/primitive_crypto.cc
valientegaston: Replaced all C-style casts in the new and modified code with unvoid_cast. Resource type conversions use static_cast as is the convention in the rest of the codebase.

valientegaston: Good catch on both points. Here's how they've been addressed:

**Buffer ownership:** prv_buf_ and pub_buf_ are now members of RsaGenerationResource, allocated in the main thread before launching the background thread. The resource destructor always frees them, so memory is never leaked even if rsa_generate_finish is never called.

**Race condition:** Addressed with a thread_running_ / destroy_when_done_ pattern protected by the EventSource mutex. If the resource is destroyed while the thread is still running, delete_or_mark_for_deletion sets destroy_when_done_ = true instead of freeing immediately. The actual delete this happens in on_thread_done once the thread finishes — and crucially, after releasing the mutex, to avoid use-after-free on the Locker destructor itself.

src/primitive_crypto.cc
valientegaston: Done. Added the following comment above the realloc calls:
```
// RSA_PRV_DER_MAX_BYTES is typically ~550 bytes, but generated keys
// usually only need ~300 bytes. The realloc shrinks the buffer to the
// actual size to recover the unused memory. If realloc fails, it returns
// null without freeing the original block, so the original pointer is
// preserved and no memory is leaked — it just means we keep the larger
// buffer, which is acceptable.

```

A second best-effort realloc is also attempted in rsa_generate_finish in case the one in the generation thread failed due to heap fragmentation at that point in time.

src/primitive_crypto.cc
valientegaston: This approach was explored but ran into a safety issue with GC retries: if prv_der allocation succeeds but pub_der fails, the runtime does GC and retries the primitive — but prv_buf_ ownership has already been transferred to the external ByteArray, leaving the resource in an inconsistent state where the private key buffer is gone and cannot be recovered on retry.
Given this complexity and the fact that RSA key generation is a one-time operation, the allocate_byte_array + memcpy pattern was kept instead. A best-effort realloc is attempted in rsa_generate_finish before copying to reduce the buffer sizes as much as possible before the peak occurs. The original buffers are freed immediately after copying on the success path.

src/primitive_crypto.cc
valientegaston: Fixed. unregister_resource and clear_external_address are now only called on the success path and on non-recoverable mbedtls errors. All ALLOCATION_FAILED paths return early without touching the resource, keeping it intact for GC retry.

valientegaston:
> Still some subtle issues. Not completely sure what the best way for the resource handling is. Maybe one way is to keep track of whether the parallel thread is running and store that in the resource. When the resource gets the signal to destroy itself, it could then delay that action until the thread returns. Alternatively (if possible) killing the thread first, would also work. We do something slightly similar in the BLE code.

Implemented the first approach — tracking whether the background thread is running via thread_running_ and destroy_when_done_ flags, protected by the EventSource mutex. If the resource receives the signal to destroy itself while the thread is still active, destruction is deferred until on_thread_done is called. The BLE code was used as a reference for the general pattern.

src/primitive_crypto.cc
```diff
@@ -905,46 +912,55 @@ class RsaGenerationResource : public Resource {
...
-    return thread_;
+    if (should_delete) delete this;
   }
 
- private:
-  int bits_;
-  int error_ = 0;
+  int bits() const { return bits_; }
+
+  bool thread_running_ = false;
```
floitsch, line 986: these fields should not be public.

src/primitive_crypto.cc
```diff
@@ -983,40 +1007,42 @@ PRIMITIVE(rsa_generate_start) {
...
+        res->prv_buf_ = prv;
+        res->prv_len_ = prv_len;
+        res->pub_buf_ = pub;
+        res->pub_len_ = pub_len;
       }
     }
     mbedtls_pk_free(&pk);
-    res->set_error(ret);
+    res->error_ = ret;
     return (word)1; // Indicate done.
```
floitsch, line 1075: Do we really need that cast? If yes, use 'static_cast'.

src/primitive_crypto.cc
```diff
@@ -1028,29 +1054,49 @@ PRIMITIVE(rsa_generate_start) {
 
 PRIMITIVE(rsa_generate_finish) {
   ARGS(RsaGenerationResource, resource);
-  if (resource->error() != 0) {
-    int err = resource->error();
+  if (resource->error_ != 0) {
```
floitsch: don't expose the private fields.

src/primitive_crypto.cc
```diff
@@ -1028,29 +1054,49 @@ PRIMITIVE(rsa_generate_start) {
...
+    int err = resource->error_;
     resource->resource_group()->unregister_resource(resource);
     resource_proxy->clear_external_address();
     return tls_error(null, process, err);
```
floitsch, line 1100: we should probably check whether this is an oom error. Not sure that can actually happen, but if it can, then just returning the oom error here would not do what we want.
Instead, we should replace the oom error with something recognizable, and then do a retry-loop from the Toit code. (with a GC in between).
That said: getting an OOM from mbedtls is maybe extremely rare (since we pass in the buffers), so maybe OOMing at that point is ok. But that would still require a few changes. Remember: if a primitive returns OOM the system runs the GC and tries again (up to 3?) times. With the current code that would yield an error as we cleared the resource-proxy. -> we can't repeat the primitive call.

toolchains/esp32/sdkconfig
```diff
@@ -2121,7 +2122,7 @@ CONFIG_TOIT_REPORT_I2S_DATA_LOSS=y
 CONFIG_TOIT_INTERPRETER_IN_IRAM=y
 # CONFIG_TOIT_INTERPRETER_HELPERS_IN_IRAM is not set
 CONFIG_TOIT_CRYPTO=y
-# CONFIG_TOIT_CRYPTO_EXTRA is not set
+CONFIG_TOIT_CRYPTO_EXTRA=y
```
floitsch, line 2160: I would expect a change in sdkconfig.defaults as well.

floitsch:
@coderabbit review

floitsch:
@valientegaston Do you think you can address the remaining comments, or should I take over?

valientegaston:
Hi! I've addressed all the remaining comments:

Private fields: All RsaGenerationResource fields are now private, accessed via getters and setters.
Thread guard: Added is_thread_running() check at the start of rsa_generate_finish to prevent accessing unfinished buffers (INVALID_STATE).
Resource tags: Moved RsaGenerationResource into NON_BLE_RESOURCE_CLASSES_DO and RsaGenerationResourceGroup into RESOURCE_GROUP_CLASSES_DO — the async path now works correctly (previously it was silently falling back to sync).
sdkconfig.defaults: Added CONFIG_TOIT_CRYPTO_EXTRA=y to all five ESP32 variants.

All tests passing with rsa_gen_test.toit.
Could you take a look and let me know if this looks good to you? Thanks!

floitsch (approved):
LGTM.

floitsch:
I will trigger the CI and land it, once it's green.

PR #3024, opened 2026-04-08
Fix assertion failure in LSP rename for flattened IR interfaces/mixins

close2:
## Summary

- Fixes debug-build assertion failure in `emit_class_hierarchy_references` (`rename.cc:885`) where IR interface/mixin lists were assumed to correspond 1:1 with AST lists
- After resolution, the IR lists are replaced with the full transitive closure (flattened and deduplicated), breaking the positional mapping
- Replaced the broken positional-index approach with name-based matching: check IR list membership first, then scan AST entries using `range_matches_name()`

## Failing tests fixed
- `tests/lsp/cross-file-class-rename-test.toit`
- `tests/lsp/return-type-rename-test.toit`

## Details

The `check_interface_implementations_and_flatten` function in `resolver.cc` replaces each class's IR interfaces with the full transitive closure (via `replace_interfaces()`), and similarly for mixins. This means the IR lists can have a different length and ordering than the AST lists (which always contain only the directly-declared interfaces/mixins from source). The assertions `ASSERT(ir_interfaces.length() == ast_interfaces.length())` and `ASSERT(ir_mixins.length() == ast_mixins.length())` were therefore incorrect.

The fix:
1. Checks whether `target_class` appears in the flattened IR interfaces/mixins list
2. If so, scans the AST interfaces/mixins by name using the existing `range_matches_name()` helper to find the actual source reference to emit

### DIFF
--- a/src/compiler/compiler.cc
+++ b/src/compiler/compiler.cc
@@ -300,6 +300,12 @@ class FindReferencesPipeline : public LocationLanguageServerPipeline {
   // Each entry maps a resolved target definition to the source range
   // of the show/export identifier that references it.
   std::vector<Resolver::ShowExportReference> show_export_references_;
+
+  // Class hierarchy references (implements/with) collected from the resolver.
+  // Each entry maps a holder class and a resolved interface/mixin to the
+  // AST expression in the source clause.  Collected before the resolver's
+  // flattening passes destroy the 1:1 AST/IR correspondence.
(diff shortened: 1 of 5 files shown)


floitsch (approved):
LGTM.
Thanks.

floitsch (approved):
Still LGTM.

PR #3025, opened 2026-04-10
Make it easier to build and flash with jag

addshore:
I once again found myself wanting to poke the toit code with an app runnable via
jaguar, and encountered some issue while following the README, but to me it's all
very nice and eacy to follow and or copy and paste now!

Changes:

README: Hints about other chip builds
README: Hints about other port formats
README: Section on "Flashing with Jaguar (live reloading)"
README: remove external esptool download/ESPTOOL_PATH workaround?

Remove underscore→hyphen conversion in firmware.toit so bundled esptool v4.x works
Rebuild SDK and verify flashing with bundled esptool

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,18 +314,18 @@ in `build/esp32/firmware.envelope`:
 make esp32
 ```
 
-If you want to flash the generated firmware on your device, you can use the `firmware`
-too. Internally, the `firmware` tool calls out to
-[esptool](https://github.com/espressif/esptool)
-so you need to install that one first.
-You can also set the environment variable `ESPTOOL_PATH` to point
-to a valid esptool (for example the one in the shipped esp-idf:
-`export ESPTOOL_PATH=$PWD/third_party/esp-idf/components/esptool_py/esptool/esptool.py`).
-Assuming your device is connected through `/dev/ttyUSB0`
-you can then flash a device as follows:
+For other chip variants use the corresponding make target. For example, for the ESP32-C6 which would generate in `build/esp32c6/firmware.envelope`:
 
 ``` sh
-build/host/sdk/lib/toit/tool firmware -e build/esp32/firmware.envelope \
+make esp32c6
+```
+
+If you want to flash the generated firmware on your device, you can use the `firmware`
+tool. Internally it calls the bundled esptool from the esp-idf submodule.
+You can override this with the `ESPTOOL_PATH` environment variable if you want to use a different version of esptool.
+
+``` sh
+build/host/sdk/bin/toit tool firmware -e build/esp32/firmware.envelope \
     flash --port /dev/ttyUSB0 --baud 921600
 ```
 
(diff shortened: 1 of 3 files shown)


addshore:
I'm stilly trying to figure out the "most ideal" way to be able to quickly iterate on the toit repo while also wanting to run custom code via jag the normal way.
I think this PR is part way there, but not all the way?

tools/firmware.toit
```diff
@@ -1116,29 +1116,17 @@ flash invocation/cli.Invocation -> none:
 
           esptool := find-esptool_
 
-          // The new esptool has deprecated underscores in some arguments.
```
floitsch, line 1119: I'm not a fan of unconditionally removing the replacements.
Arch users have the latest esptool installed and that would yield errors. The 5.2 esptool version which is shipped with Toit handles the '-' just fine.

Compromise:
until we update the esp-idf we can make these replacements on whether `JAG_TOIT_REPO_PATH` is not set.

README.md
```diff
@@ -314,18 +314,18 @@ in `build/esp32/firmware.envelope`:
...
 If you want to flash the generated firmware on your device, you can use the `firmware`
-too. Internally, the `firmware` tool calls out to
-[esptool](https://github.com/espressif/esptool)
-so you need to install that one first.
-You can also set the environment variable `ESPTOOL_PATH` to point
-to a valid esptool (for example the one in the shipped esp-idf:
-`export ESPTOOL_PATH=$PWD/third_party/esp-idf/components/esptool_py/esptool/esptool.py`).
-Assuming your device is connected through `/dev/ttyUSB0`
-you can then flash a device as follows:
+tool. Internally it calls the bundled esptool from the esp-idf submodule.
```
floitsch, line 324: That only works if one has sourced the `export.sh`. Otherwise it doesn't. (I didn't know that until today...)

README.md
```diff
@@ -314,18 +314,18 @@ in `build/esp32/firmware.envelope`:
...
+You can override this with the `ESPTOOL_PATH` environment variable if you want to use a different version of esptool.
 
 ``` sh
-build/host/sdk/lib/toit/tool firmware -e build/esp32/firmware.envelope \
+build/host/sdk/bin/toit tool firmware -e build/esp32/firmware.envelope \
```
floitsch, line 328: good change. thanks.

README.md
```diff
@@ -340,6 +340,39 @@ build/host/sdk/bin/toit tool firmware -e build/esp32/firmware.envelope \
     flash --port /dev/ttyUSB0 --baud 921600
 ```
 
+### Flashing with Jaguar (live reloading)
+
+To run Jaguar on your locally built firmware, compile the Jaguar service snapshot
```
floitsch, line 345: Normally I use `JAG_TOIT_REPO_PATH` from the Jaguar directory. When would you use the approach here instead of the env-variable one?

PR #3028, opened 2026-04-14
Use newer pkg-cli with typed options and default command support

floitsch:
Upgrade pkg-cli to 2.14.0 and pkg-http to 2.12.0. Migrate all CLI option declarations from `cli.Option --type="file"/"dir"` to the typed alternatives (OptionPath, OptionInFile, OptionOutFile). Add a default command to the root `toit` CLI so that `toit foo.toit` runs the file directly. Add shell completion callbacks for --os/--arch in compile and for the filter argument in toitp.


floitsch:
TBR.

PR #3032, opened 2026-04-29
Fix UART on Windows

floitsch:
- Re-issue ReadFile on 0-byte completions so the port can't wedge.
- Close the comm-events event handle in do_close.
- Wire EV_TXEMPTY to kWriteState and implement wait_tx via
  ClearCommError/cbOutQue so flush() works.
- Initialize DCBlength before GetCommState in get/set_baud_rate.
- Use UartResource (not IntResource) in the errors primitive.
- Assert write_ready_ in send to guard the write_buffer_ free.


floitsch:
TBR.

Issue #2950, opened 2025-11-16
LSP crashes


floitsch:
[repro_33547304427-8.tar.gz](https://github.com/user-attachments/files/23568155/repro_33547304427-8.tar.gz)
[repro_42660807680-1.tar.gz](https://github.com/user-attachments/files/23568154/repro_42660807680-1.tar.gz)

floitsch:
[repros.zip](https://github.com/user-attachments/files/26169553/repros.zip)

floitsch:
All crashes should be fixed.

PR #3033, opened 2026-04-29
Fix type-checker crash on global initializer with self-assignment

floitsch:
When a global initializer contains an assignment back to the same global (e.g. inside a block argument), the type checker visits the assignment before the global's return type has been computed. The ASSERT in `check` then fired because `global->return_type()` is invalid.

`AssignmentGlobal` creation in the resolver already documents that the receiver type may be unset at this point and that the type check is inserted "later" — handle the invalid type gracefully here instead of crashing.

Regression test in tests/negative covers the block and lambda variants. Closes the assertion crash from #2950.


floitsch:
TBR.

PR #3034, opened 2026-04-29
Add repro-debug skill

floitsch:
Add SKILL.md for debugging LSP repro crashes
Captures the workflow for triaging compiler crashes from LSP repro tar
archives (issue attachments): tar layout, how to start the file server +
drive toit.compile, the gdb wrapper, how to reduce to a tests/negative
case, and the make update-gold flow.


floitsch:
TBR

PR #3035, opened 2026-04-29
Clamp out-of-range positions in compute_source_offset

floitsch:
The LSP server is asynchronous: a `did-change` that arrives between the moment a completion (or definition / hover / rename) request is queued and the moment the forked compiler actually reads the buffer over the TCP file-server can leave the compiler with a (line, column) position that is past the end of the (now shorter) line. The compiler then walks past the end-of-line/end-of-file sentinel and aborts on the `UNREACHABLE()` in `compute_source_offset`.

The same shape also occurs without any race when the cursor is at the end of the last line of a file with no trailing newline -- LSP allows column = N for an N-character line, which already lands on the trailing '\0' in the C string view.

Make `compute_source_offset` clamp to the closest valid position instead of aborting in both spots, and document the new contract.

LSP regression test exercises positions past EOL and past EOF. Closes the SIGABRT crashes from #2950 (the "Aborted" repros).

### DIFF
--- a/src/compiler/compiler.cc
+++ b/src/compiler/compiler.cc
@@ -968,17 +968,22 @@ SnapshotBundle Compiler::compile(const char* source_path,
 /// The line number should be 1-based.
 /// The column number should be 1-based.
 ///
-/// Aborts the program if the file is not big enough.
+/// If the requested position is past the end of the file (or past the end of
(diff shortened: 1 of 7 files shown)


floitsch:
TBR

PR #3044, opened 2026-05-16
Ignore mDNS AD/CD/Z bits on reception

close2:
## Summary
- mask the mDNS-only ignored status bits before `DecodedPacket` validation
- keep regular DNS decoding strict while allowing Android-style multicast queries through the lenient path
- add a regression test for an mDNS query with the AD bit set

## Testing
- `ctest --test-dir build/host -R 'tests/dns-tools-test\\.toit$' -C slow --output-on-failure`
- `build/host/sdk/bin/toit tests/dns_packet_test.toit foo bar gee`


floitsch (approved):
LGTM.
thanks!

PR #3045, opened 2026-05-16
Fix memory leak in float_to_string primitive

girtsf:
`safe_double_print` returns a malloc'd `char*` that `allocate_string` copies but never frees. Every `"$(%.Nf v)"` interpolation or `v.to-string --precision=N` call leaked ~16 B plus heap overhead, which OOMs apps that format floats in a steady loop.


floitsch (approved):
LGTM!
thanks.

PR #3049, opened 2026-05-16
pkg: defer auto-sync and cache git registries by ref-hash

floitsch:
## Summary
- Make registry auto-sync lazy: sync only fires when a command actually reads
  registry data (search, retrieve-versions, list-packages, …) instead of on
  every project command. The `auto-sync` flag moves from `ProjectConfiguration`
  to `Registries`, and the eager `registries.sync` in `PkgProjectCommand` is
  gone.
- Switch the git-registry cache from a single-file pack to a `DirectoryStore`
  containing `registry.pack` + `ref-hash.txt`. On sync we now consult the
  remote HEAD and reuse the cached pack when the hash is unchanged, so a sync
  no longer always re-downloads.
- Add a `--clear-cache` flag to both `toit pkg sync` and
  `toit pkg registry sync` to force a fresh download.
- Make `toit pkg registry add` idempotent: re-adding a registry with the same
  name and URL/path is a no-op; only a mismatching URL/path aborts. The
  command normalizes local paths to absolute before handing them to the
  `Registries` API, which now requires the path to already be absolute.
- Drop the redundant `registries.sync` from the `search` command — the lazy
  sync covers it.
- Tests: new gold tests `registry-sync-gold-test` and `registry-git-gold-test`
  covering auto-sync behavior, the `--clear-cache` flag, and that `install`
  doesn't trigger a sync; the gold tester learns to inspect/delete the
  registry cache and resolve git-registry paths. Existing
  `registry-test`/`solver-test`/`package-test` updated for the new
  `--auto-sync` constructor argument and the absolute-path requirement.


floitsch:
TBR

PR #3050, opened 2026-05-16
pkg: gold test for invalid registry sync and absolute-path validation

floitsch:
## Summary
- Add a gold test asserting `toit pkg registry sync <unknown>` reports a clear error and aborts.
- In `Registries.add --local`, treat a non-absolute path as a programmer error (`throw`) and check it before the directory-exists check; the abort-with-message path is gone.
- Update the corresponding `registry-test` case to pass an absolute path so it exercises the "already exists" check rather than tripping the new assertion.


floitsch:
TBR.

PR #3051, opened 2026-05-16
pkg: update spec constraints to resolved versions; pretty-print ^/~ ranges

floitsch:
## Summary

- After install/update/install-local, rewrite the lower bound of each remote dependency's version constraint in the project spec to the version actually picked by the solver. For example, installing a package whose latest matching version is `1.2.3` now records `^1.2.3` instead of `^1.0.0`, and `>=1.0.0,<2.0.0` becomes `^1.0.1` once the solver picks `1.0.1`.
- Pretty-print constraints in the spec file: `>=X.Y.Z,<(X+1).0.0` is emitted as `^X.Y.Z` and `>=X.Y.Z,<X.(Y+1).0` as `~X.Y.Z`.
- Add gold tests for the new "preferred" behavior (installing a new package does not bump versions of existing packages) and for `pkg update`. Existing gold files are updated to reflect the new constraint formatting.
- Minor cleanup in `specification.toit` to use `Map.get --init`.


floitsch:
TBR

PR #3052, opened 2026-05-16
pkg: port init test from Go to Toit

floitsch:
## Summary
- Ports the Go `pkg init` test to a Toit gold test ([init-gold-test.toit](tests/pkg/init-gold-test.toit)), covering basic init, double-init, install on the generated lock file, and init with an explicit `--project-root`.
- Adjusts [gold-tester.toit](tests/pkg/gold-tester.toit) so `pkg` commands only get an injected `--project-root` when the caller hasn't already supplied one, enabling the new test case for `pkg init --project-root <other>`.
- Adds the matching gold files and test assets under [tests/pkg/assets/init/](tests/pkg/assets/init/).


floitsch:
TBR

PR #3053, opened 2026-05-16
pkg: port install-for-package test from Go to Toit

floitsch:
## Summary
- Adds a gold-based Toit port of the `install-for-package` test, covering `pkg init`, installing two packages by name, and re-running `pkg install` after deleting `package.lock` to verify the lock file is recreated from `package.yaml`.
- Includes the two expected gold files (`00-test.gold`, `10-test.gold`) capturing the CLI output and resulting `package.lock` / `package.yaml` contents.


floitsch:
TBR.

PR #3054, opened 2026-05-16
pkg: abort with clear error when TOIT_PACKAGE_INSTALL_PATH is set

floitsch:
## Summary
- The `TOIT_PACKAGE_INSTALL_PATH` environment variable is no longer supported.
- Detect it in `pkg install` and abort with a clear error message instead of silently ignoring it.


floitsch:
TBR.

PR #3055, opened 2026-05-16
pkg: detect package.yaml / package.lock prefix mismatch

floitsch:
- Before resolving dependencies, compare the prefixes in `package.yaml` against those recorded in `package.lock`.
- If either file has prefixes the other doesn't, emit a warning listing the offending prefixes and abort with a clear error instead of silently continuing.
- Add a gold test covering the two mismatch directions (extra in lock, extra in yaml).


floitsch:
TBR.

PR #3056, opened 2026-05-16
pkg: add gold test for default registry

floitsch:
Adds a gold test that verifies behavior when no registry cache file is
present and the default `toit` registry from GitHub is used.

Extends `with-gold-tester` with a `--with-default-registry` flag that
suppresses writing the registry cache file so the package manager falls
back to the default registry. The new test exercises `pkg registry
list`, `pkg init`, and `pkg install github.com/toitware/toit-morse@1.0.6`
against that default registry.


floitsch:
TBR.

PR #3057, opened 2026-05-16
pkg: add gold test for registry remove

floitsch:
Adds a gold test that exercises `pkg registry remove`, covering removal of the default `toit` registry, removal of git and local registries, and the error case for removing a non-existent registry.


floitsch:
TBR.

PR #3058, opened 2026-05-16
pkg: use global --verbose flag for list and search

floitsch:
Removes the local `--verbose` flag from the `list` and `search` package commands and instead relies on the global `--verbose` flag via `ui.level`.

The gold tester is updated so that `--verbose` arguments in test fixtures set the UI level to verbose rather than being passed through as a command-line option.


floitsch:
TBR.

PR #3060, opened 2026-05-17
Fix UART write+flush hang on partial writes.

floitsch:
`UartWriter.write` passed `--flush=flush` through to `try-write` inside the loop. When `try-write_` returned a partial count, `try-write` flushed those bytes to completion, leaving the ring buffer empty. The outer loop then called `wait-for-more-room_`, which waited indefinitely for an event that hardware can no longer generate.

Match the pattern in `io.Writer.write`: flush once after all data has been written.


floitsch:
TBR.

PR #3061, opened 2026-05-17
Use SOC_DRAM_LOW/HIGH for heap-range fallback on all targets.

floitsch:
The hard-coded fallback in OS::get_heap_memory_range() is only used when cmpctmalloc doesn't populate lowest_address/highest_address (which shouldn't happen in normal Toit ESP32 builds, but acts as a safety net).

Previous values were stale or wrong:
- ESP32-C3 and -C6 fell through to the ESP32 #else branch with addresses (0x3FFC0000) that don't exist on those chips.
- ESP32-S2 had the wrong start address, missing the low 64 KB of DRAM.
- ESP32-S3 covered 384 KB starting at 0x3FCA0000, missing the lower ~96 KB of valid DRAM.
- ESP32 missed the low pools (BT/ROM-shared region).

Use SOC_DRAM_LOW/SOC_DRAM_HIGH from <soc/soc.h> uniformly; the GC's ONLY_FOR_MALLOC adjustment trims the low part anyway.


floitsch:
TBR.

PR #3062, opened 2026-05-17
Fix JSON encoding of floats; reject NaN and infinity

girtsf:
Two related fixes to `lib/encoding/json.toit`:

1. **Fix data loss when encoding floats.** The encoder used
   `float.stringify 2`, forcing 2-decimal-place format and
   silently truncating small magnitudes (e.g. `-0.00001` was
   encoded as `-0.00`). Switched to `float.to-string`, which
   produces the shortest representation that round-trips back
   to the original value. Also drops a deprecated API call.

2. **Reject NaN and ±Infinity.** These are not valid JSON
   numbers but were written into the output as bare `nan` /
   `inf` / `-inf` tokens that JSON parsers generally won't
   accept. Now throws `INVALID_JSON_NUMBER` so the encoder
   either produces valid JSON or fails loudly. Matches the
   behaviour of Go's `encoding/json` and Rust's `serde_json`.

### DIFF
--- a/lib/encoding/json.toit
+++ b/lib/encoding/json.toit
@@ -172,6 +172,8 @@ class Encoder extends EncoderBase_:
     writer_.write-byte '"'
 
   encode-number_ number:
+    if number is float and not number.is-finite:
+      throw "INVALID_JSON_NUMBER"
     str := number.to-string
     writer_.write str
 
--- a/lib/encoding/json.toit
+++ b/lib/encoding/json.toit
@@ -172,7 +172,7 @@ class Encoder extends EncoderBase_:
     writer_.write-byte '"'
 
   encode-number_ number:
-    str := number is float ? number.stringify 2 : number.stringify
+    str := number.to-string
     writer_.write str
 
   encode-true_:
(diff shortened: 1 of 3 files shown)


floitsch (approved):
LGTM!

floitsch:
CI is failing because the health tests don't succeed anymore.
```
79/179 Test  #538: health-tests/json-test.toit ..........................................***Failed    0.15 sec
--- /home/runner/work/toit/toit/build/host/tmp/GOLD_YoC8HwGk0NCt	2026-05-18 21:16:05.419367591 +0000
+++ /home/runner/work/toit/toit/build/host/tmp/OUTPUT_YoC8HwGk0NCt	2026-05-18 21:16:05.419367591 +0000
@@ -1,3 +1,3 @@
-tests/json-test.toit:386:17: warning: Class 'BufferedReader' is deprecated. Use 'io.Reader' instead
+tests/json-test.toit:393:17: warning: Class 'BufferedReader' is deprecated. Use 'io.Reader' instead
     buffered := BufferedReader (TestReader it)
                 ^~~~~~~~~~~~~~
CMake Error at tests/tools/health_run.cmake:86 (message):
  Not equal
```

The easiest is to just run `make update-health-gold`

girtsf:
> CI is failing because the health tests don't succeed anymore.

Done!

floitsch (commented):
Still LGTM.

PR #3072, opened 2026-05-24
Switch to a single file for the registry cache.

floitsch:
Also use 'cache.update' which is why we now need a more recent cli package.


floitsch:
TBR.

PR #3076, opened 2026-06-03
Fix 32-bit host: accept int64 epoch seconds; enlarge Toit stack

floitsch:
## Summary
Two related fixes for intermittent failures in the 32-bit Debug
nightly CI (pkg gold tests).

### `file.update-time`: accept int64 seconds
The `update_times` primitive declared `atime_sec` / `mtime_sec` as
`int`, which in the ARGS macro only accepts Smi. Smi max on 32-bit
is 2^30 - 1 ≈ 1.07e9, but the current Unix epoch is ≈ 1.78e9, so
`Time.now.s-since-epoch` is a `LargeInteger` and the primitive
rejected it with `OUT_OF_RANGE`.

### 32-bit host page size matches 64-bit
A Toit stack's max length is bounded by `TOIT_PAGE_SIZE`, which was
4kB on every 32-bit build vs 32kB on 64-bit — an 8x smaller stack
cap. On constrained devices (FreeRTOS) we want to keep 4kB, but a
32-bit host build has plenty of memory. The smaller cap was tipping
`yaml.decode` over the edge when it ran deep inside the pkg solver,
showing up as the intermittent "Failed to decode <path> as YAML"
`STACK_OVERFLOW` in nightly 32-bit Debug.

Now `TOIT_PAGE_SIZE_LOG2` only falls back to the 32-bit value when
`TOIT_FREERTOS` is defined.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,10 +101,8 @@ on:
 env:
   TOIT_VERSION: ${{ github.event.inputs.do-release || github.event.release.tag_name || '' }}
   DO_RELEASE: ${{ github.event_name == 'release' || startsWith(github.event.inputs.do-release, 'v') }}
-  # TEMP: floitsch/debug-ci-failure runs the full scheduled test set so we can
-  # reproduce the intermittent pkg gold-test flake seen in nightly CI.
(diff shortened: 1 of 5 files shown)


floitsch:
TBR.

PR #3077, opened 2026-06-13
Add support for specifying types for 'any'.

floitsch:
This is an internal mechanism that won't stay the way it currently is.

### DIFF
--- a/src/compiler/ast.h
+++ b/src/compiler/ast.h
@@ -876,12 +876,12 @@ class Call : public Expression {
 /// through a `// __TYPE-MIGRATION__ <name>: <type>` comment in front of the declaration.
 ///
 /// Temporary mechanism to migrate `any` parameters to concrete types.
-class TypeAnnotation {
+class MigrationType {
  public:
-  TypeAnnotation(Expression* type,
-                 bool is_deprecated,
-                 Symbol deprecation_message,
-                 Source::Range range)
+  MigrationType(Expression* type,
+                bool is_deprecated,
+                Symbol deprecation_message,
+                Source::Range range)
       : type_(type)
       , is_deprecated_(is_deprecated)
       , deprecation_message_(deprecation_message)
@@ -924,9 +924,9 @@ class Parameter : public Expression {
   bool is_field_storing() const { return is_field_storing_; }
   bool is_block() const { return is_block_; }
 
-  List<TypeAnnotation*> type_annotations() const { return type_annotations_; }
-  void set_type_annotations(List<TypeAnnotation*> annotations) {
-    type_annotations_ = annotations;
+  List<MigrationType*> migration_types() const { return migration_types_; }
+  void set_migration_types(List<MigrationType*> types) {
+    migration_types_ = types;
   }
 
   Source::Range full_range() const override;
@@ -938,7 +938,7 @@ class Parameter : public Expression {
    bool is_named_;
    bool is_field_storing_;
    bool is_block_;
-   List<TypeAnnotation*> type_annotations_;
+   List<MigrationType*> migration_types_;
 };
 
 class LiteralNull : public Expression {
--- a/src/compiler/ast.h
+++ b/src/compiler/ast.h
@@ -872,6 +872,35 @@ class Call : public Expression {
   bool is_call_primitive_;
 };
 
+/// A type alternative for a parameter that is declared as `any`, given
+/// through a `// __TYPE-MIGRATION__ <name>: <type>` comment in front of the declaration.
+///
+/// Temporary mechanism to migrate `any` parameters to concrete types.
+class TypeAnnotation {
+ public:
+  TypeAnnotation(Expression* type,
+                 bool is_deprecated,
+                 Symbol deprecation_message,
+                 Source::Range range)
+      : type_(type)
+      , is_deprecated_(is_deprecated)
+      , deprecation_message_(deprecation_message)
+      , range_(range) {}
+
+  Expression* type() const { return type_; }
+  bool is_deprecated() const { return is_deprecated_; }
+  /// Either empty, or starting with ". ", so it can be appended
+  /// directly to a warning. Only valid if [is_deprecated].
+  Symbol deprecation_message() const { return deprecation_message_; }
+  Source::Range range() const { return range_; }
+
+ private:
+  Expression* type_;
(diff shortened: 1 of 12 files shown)


src/compiler/ast.h
```diff
@@ -872,6 +872,35 @@ class Call : public Expression {
   bool is_call_primitive_;
 };
 
+/// A type alternative for a parameter that is declared as `any`, given
+/// through a `// __TYPE-MIGRATION__ <name>: <type>` comment in front of the declaration.
+///
+/// Temporary mechanism to migrate `any` parameters to concrete types.
+class TypeAnnotation {
```
floitsch: The name is a bit too generic and makes me think of the normal types (like `foo/Type`).
maybe `CommentType`, `TypeHack`, `MigrationType` or something else? Come up with suggestions.

src/compiler/type_annotations.cc
```diff
@@ -0,0 +1,310 @@
...
+#include "symbol_canonicalizer.h"
+
+namespace toit {
+namespace compiler {
+
+namespace {  // anonymous
+
+static const char* const TYPE_MARKER = "__TYPE-MIGRATION__";
+
+static bool is_identifier_start(int c) {
```
floitsch: are these not accessible from the scanner?

src/compiler/type_annotations.cc
```diff
@@ -0,0 +1,310 @@
...
+}
+
+static bool matches(const uint8* text, int pos, int to, const char* str) {
+  int len = strlen(str);
+  if (pos + len > to) return false;
+  return memcmp(text + pos, str, len) == 0;
+}
+
+static std::string trim_whitespace(const std::string& str) {
+  auto start = str.find_first_not_of(" \t");
```
floitsch: don't overcomplicate. We write the annotations. We don't allow `\t`.

src/compiler/migration_type.cc
```diff
@@ -0,0 +1,310 @@
...
+      auto comment = comments_[i];
+      if (comment.is_multiline() || comment.is_toitdoc()) continue;
+      auto text = source_->text();
+      int from = source_->offset_in_source(comment.range().from());
+      int to = source_->offset_in_source(comment.range().to());
+      ASSERT(matches(text, from, to, "//"));
+      int pos = skip_spaces(text, from + 2, to);
+      if (!matches(text, pos, to, TYPE_MARKER)) continue;
+      pos += strlen(TYPE_MARKER);
+      // Require a space after the marker, so we don't pick up unrelated
```
floitsch, line 84: We can leave this, but it wouldn't have been necessary.

src/compiler/migration_type.cc
```diff
@@ -0,0 +1,310 @@
...
+        parsed.is_used = true;
+        annotations.add(parsed.annotation);
+      }
+      if (!annotations.is_empty()) {
+        parameter->set_type_annotations(annotations.build());
+      }
+    }
+
+    // Report annotations of this block that didn't match any parameter.
+    for (int i = first; i <= closest; i++) {
```
floitsch, line 107: I feel like this is making things complicated and duplicated.
Just run through the comments (i = first -> closest).
For each run through the parameters and attach or report that you couldn't find a target.
Should make the code easier.

src/compiler/type_check.cc
```diff
@@ -827,6 +817,96 @@ class TypeChecker : public ReturningVisitor<Type> {
...
+    if (value_type == null_type_ && parameter->has_default_value()) return;
+    for (auto annotation : annotations) {
+      if (annotation.is_deprecated()) continue;
+      if (is_assignable(annotation.type(), value_type)) return;
+    }
+    for (auto annotation : annotations) {
+      if (!annotation.is_deprecated()) continue;
+      if (is_assignable(annotation.type(), value_type)) {
+        report_warning(range,
+                       "Deprecated type '%s' for parameter '%s'%s",
```
floitsch: Merge:
```
for (auto annotation : annotations) {
  if (!is_assignable(annotation.type(), value_type()) continue;
  if (!annotation.is_deprecated()) return;
  report_warning( ... );
  return;
```

tests/negative/type-migration-test.toit
```diff
@@ -0,0 +1,98 @@
...
+
+interface Reader:
+  read -> none
+
+class StringReader implements Reader:
+  read -> none:
+
+// __TYPE-MIGRATION__ rx: Pin. Deprecated. Provide an integer instead.
+// __TYPE-MIGRATION__ rx: int
+// __TYPE-MIGRATION__ tx: Pin. Deprecated. Provide an integer instead.
```
floitsch, line 18: Add an example where the type is prefixed. Something like `gpio.Pin`.

src/compiler/ast.h
floitsch: done — renamed to `MigrationType` (ast + ir), incl. accessors `migration_types()`/`set_migration_types()`.

src/compiler/type_annotations.cc
floitsch: done — reuse `Scanner::is_identifier_start` and `is_decimal_digit` from scanner.h; only kept a thin `is_identifier_part` wrapper for the `-` case.

src/compiler/type_annotations.cc
floitsch: done — now trims spaces only (renamed to `trim_spaces`).

src/compiler/migration_type.cc
floitsch: done — left as is.

src/compiler/migration_type.cc
floitsch: done — rewrote as a single pass over the comments; for each, find the parameter and attach (accumulating per parameter) or report. Dropped the separate reporting loop.

src/compiler/type_check.cc
floitsch: done — merged into the single loop you sketched.

tests/negative/type-migration-test.toit
floitsch: done — added `consume-deprecated-port` with a prefixed `lib.Port` deprecated alternative, plus a `lib.SubPort` subclass to exercise subtype matching through a prefix.

floitsch:
TBR.

PR #3078, opened 2026-06-13
Don't print diagnostics when TISON encoding/decoding fails.

### LINKED ISSUE #692: Error decoding assets while listing them
I have a decoding error while listing my assets, which persists even if I start a new asset file from scratch, and repeats once per asset.

It is easily reproducible through normal operation:

```
$ jag toit tool assets create -e deleteme

$ jag toit tool assets -e deleteme add debug debug.bin

$ jag toit tool assets -e deleteme list
[message decoder: wrong tison marker 0x10001000 - expected 0xa68901f7]
Assets:
  debug:
    size: 516
    kind: binary
```

floitsch:
The message encoder/decoder is shared between the inter-process messaging system and the user-facing `encoding.tison` library. For the messaging system a malformed message or an unencodable object is a genuine bug, so the diagnostic prints are kept. For TISON, however, failures are recoverable and surfaced to the caller as a thrown exception: callers routinely probe whether some bytes are TISON by decoding inside a `catch` (for example the `assets` tool when listing assets). Printing diagnostics in that case leaks spurious output to stdout, e.g.:

    [message decoder: wrong tison marker 0x10001000 - expected 0xa68901f7]

Silence the diagnostics on the TISON path while keeping them for IPC.

Fixes toitlang/jaguar#692.

### DIFF
--- a/src/messaging.cc
+++ b/src/messaging.cc
@@ -237,7 +237,9 @@ bool MessageEncoder::encode_any(Object* object) {
     write_uint64(bit_cast<uint64>(LargeInteger::cast(object)->value()));
     return true;
   } else if (is_heap_object(object)) {
-    printf("[message encoder: cannot encode object with class tag = %d]\n", HeapObject::cast(object)->class_tag());
+    if (!encoding_tison()) {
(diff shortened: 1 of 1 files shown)


floitsch:
TBR.

PR #3081, opened 2026-06-28
feat(debug): host bytecode debugger (dbg: protocol, breakpoints, stepping, inspect)

davidg238:
## What

Adds an in-image **bytecode debugger** to the host (POSIX) Toit VM: a breakpoint /
single-step mechanism in the interpreter plus a controller thread that speaks the
line-based `dbg:` protocol over stdin/stdout. An operator can set breakpoints,
step (into/over/out), and inspect live frame registers of a program running under
`toit.run --debug <snapshot>`.

See **[`docs/debugger.md`](docs/debugger.md)** for the full design and the `dbg:`
protocol reference.

## Highlights

- **Zero overhead when off** — the per-bytecode hook is a single predicted-not-taken
  branch on a flag that is false unless `--debug` / `OEVM_DEBUG` is set.
- **Park, don't block** — a hit returns `Result::DEBUG_PAUSED`; the scheduler parks
  the target while a dedicated controller thread (not a scheduler worker, so it may
  block on stdin) drives the `dbg:` channel. Capped to one worker thread under debug.
- **Snapshot-relative breakpoints** keyed on `(program, method-id, offset)`.
- Verbs: `dbg:methods` / `break` / `clear` / `continue` / `step` / `over` / `out` /
  `inspect`. The VM emits **numeric ids only**; names, source positions, and class
  names are resolved offline from the snapshot.
- **Typed register values** (`inspect`): `null`/`true`/`false`, integers, doubles,
  quoted/escaped strings, and `<obj:<class_id>>` for heap objects.
- **Offline snapshot dumps** in `tools/toitp.toit`: `positions`
  (`absolute_bci → file:line`) and `class-names` (`class_id → name`), beside the
  existing `bytecodes` dump.

## Operator tooling

The operator-facing CLI + browser UI is the companion PR toitlang/jaguar#695
(`jag debug` / `jag debug --web`). This PR is the VM-side prerequisite. A throwaway
Python proof-of-concept driver lives in `tools/debug/`.

## Testing

13 commits; built on the host (`cd build/host && ninja sdk/bin/toit`) with
integration coverage in `tests/debugger/`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/debugger.md
+++ b/docs/debugger.md
@@ -1,0 +1,164 @@
+# Host bytecode debugger
+
+The host (POSIX) Toit VM has an in‑image bytecode debugger: a breakpoint /
+single‑step mechanism in the interpreter plus a controller that speaks the
+line‑based `dbg:` protocol over stdin/stdout. It lets an operator set
+breakpoints, step, and inspect live frames of a program running under the host
+VM.
+
+This document is the in‑repo reference for the VM side. The operator‑facing
+tooling — the `jag debug` CLI and its `--web` browser UI, which resolve method
+**names**, source **lines**, and class **names** offline from the snapshot — lives
+in the `jaguar` repo (`docs/debug_spec.md` / `docs/debug_design.md` there). The
+`dbg:` protocol mirrors the OEVM reference debugger so a Toit node can be debugged
+like the OEVM node.
+
+## Enabling it
+
+The debugger is active only when debugging is requested: the `--debug` flag on the
+inner runner (`toit.run --debug <snapshot>`) or the `OEVM_DEBUG` / `TOIT_DEBUG`
+environment variable. When off, the only cost on the bytecode path is a single
+predicted‑not‑taken branch on a `false` flag — **zero measurable overhead** when
+not debugging (a hard design constraint).
+
+Debug a **pre‑compiled snapshot**, not raw source. Bytecode indices (`bci`) are
+program‑relative and the runtime interleaves several programs (privileged system,
+services, and under `toit run` the compiler); debugging a snapshot means only the
+target app + system run, so breakpoints can key on `(program, method‑id, offset)`.
+
+## The `dbg:` protocol
+
+Line‑based, newline‑delimited, over the VM's stdin (requests) and stdout
+(responses). Response lines are interleaved with the program's own stdout; the
+operator splits on `\n` and distinguishes protocol lines by the `dbg:` prefix.
+
+**Requests** (operator → VM):
+
+| Request | Meaning |
+|---|---|
+| `dbg:methods` | list the target's methods |
+| `dbg:break <id> <off>` | set a breakpoint at method `id`, offset `off` |
+| `dbg:clear <id> <off>` | clear that breakpoint |
+| `dbg:continue` | resume the parked process |
+| `dbg:step` / `dbg:over` / `dbg:out` | single‑step (into / over / out) |
+| `dbg:inspect [frame]` | dump a parked frame's registers (default frame 0) |
+
+**Responses** (VM → operator):
+
+| Response | Meaning |
+|---|---|
+| `dbg:ready` | the controller is up; emitted once before the program runs |
+| `dbg:ok <verb>` | acknowledges a request |
+| `dbg:paused break <id> <off>` | hit a breakpoint at method `id`, offset `off` |
+| `dbg:paused step <id> <off>` | stopped after a step |
+| `dbg:stack off=<bci> r0=<v> r1=<v> …` | a frame's registers (answer to `inspect`) |
+| `dbg:error <msg>` | error (e.g. unknown method id) |
+
+The VM emits **numeric method ids only**; ids are 1‑based, assigned by
+`dbg:methods` in dispatch‑table order, and stable for the session. Names, source
+positions, and class names are all resolved **offline** from the snapshot by the
+operator tooling — the wire protocol stays "VM numeric, names resolved offline".
+
+After `dbg:ready`, the VM forces a pause at the first non‑privileged program's
+entry (`dbg:paused break -1 0`) so the operator can install breakpoints before the
+program makes progress.
+
+## Architecture
+
+Three layers, with a deliberate thread split (`src/debugger.{h,cc}`,
+`src/interpreter*.{h,cc}`, `src/scheduler.{h,cc}`):
+
+1. **Interpreter mechanism.** For every bytecode of a non‑privileged process,
+   `should_break(program, bci)` is checked (gated behind the debug‑active flag). On
+   a hit — a matching breakpoint, or any bytecode while stepping — the interpreter
+   stores the stack and returns a new `Result::DEBUG_PAUSED`.
+2. **Scheduler park/resume.** On `DEBUG_PAUSED` the scheduler **parks** the process
+   (does *not* re‑ready it) and calls `register_paused`. `resume_debug_process(pid,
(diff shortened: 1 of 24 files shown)


docs/debugger.md
```diff
@@ -0,0 +1,171 @@
...
+not debugging (a hard design constraint).
+
+Debug a **pre‑compiled snapshot**, not raw source. Bytecode indices (`bci`) are
+program‑relative and the runtime interleaves several programs (privileged system,
+services, and under `toit run` the compiler); debugging a snapshot means only the
+target app + system run, so breakpoints can key on `(program, method‑id, offset)`.
+
+## The `dbg:` protocol
+
+Line‑based, newline‑delimited, over the VM's stdin (requests) and stdout
```
floitsch, line 31: If I don't forget, I will repeat some of my comments from here in the actual code.

I don't like interleaving with stdin/stdout.
The protocol is fine, but it should be through a different executable or so. Basically, something that attaches to the running process and gives the stdin/stdout functionality this way.

Not sure what the typical way to do this is.

docs/debugger.md
```diff
@@ -0,0 +1,171 @@
...
+| Response | Meaning |
+|---|---|
+| `dbg:ready` | the controller is up; emitted once before the program runs |
+| `dbg:ok <verb>` | acknowledges a request |
+| `dbg:paused break <id> <off>` | hit a breakpoint at method `id`, offset `off` |
+| `dbg:paused step <id> <off>` | stopped after a step |
+| `dbg:stack off=<bci> r0=<v> r1=<v> …` | a frame's registers (answer to `inspect`) |
+| `dbg:error <msg>` | error (e.g. unknown method id) |
+
+The VM emits **numeric method ids only**; ids are 1‑based, assigned by
```
floitsch, line 57: Methods already have their numeric IDs. We should use those and make sure that they are identical.

docs/debugger.md
```diff
@@ -0,0 +1,171 @@
...
+The host (POSIX) Toit VM has an in‑image bytecode debugger: a breakpoint /
+single‑step mechanism in the interpreter plus a controller that speaks the
+line‑based `dbg:` protocol over stdin/stdout. It lets an operator set
+breakpoints, step, and inspect live frames of a program running under the host
+VM.
+
+This document is the in‑repo reference for the VM side. The operator‑facing
+tooling — the `jag debug` CLI and its `--web` browser UI, which resolve method
+**names**, source **lines**, and class **names** offline from the snapshot — lives
+in the `jaguar` repo (`docs/debug_spec.md` / `docs/debug_design.md` there). The
```
floitsch, line 12: we should have `toit debug` and not `jag debug`. (`jag debug` could forward, though).
1. this is a core functionality.
2. more code in Toit than in Golang.
3. people typically don't use Jaguar when writing host code (which is the only place where debugging works).

docs/debugger.md
```diff
@@ -0,0 +1,171 @@
...
+
+1. **Interpreter mechanism.** For every bytecode of a non‑privileged process,
+   `should_break(program, bci)` is checked (gated behind the debug‑active flag). On
+   a hit — a matching breakpoint, or any bytecode while stepping — the interpreter
+   stores the stack and returns a new `Result::DEBUG_PAUSED`.
+2. **Scheduler park/resume.** On `DEBUG_PAUSED` the scheduler **parks** the process
+   (does *not* re‑ready it) and calls `register_paused`. `resume_debug_process(pid,
+   step_mode)` re‑readies it later. While debugging, the scheduler is capped to a
+   single worker thread.
+3. **Controller thread.** A dedicated OS thread (not a scheduler worker, so it may
```
floitsch, line 79: I think I would always use a (local) port for that. A helper executable can then translate the socket to stdout/stdin.
Maybe in reverse order:
spawn the helper which reserves a port, then run the Toit executable telling it to open a debug-socket on that port. (This way we wouldn't need to worry about multiple programs using the same port).

docs/debugger.md
```diff
@@ -0,0 +1,171 @@
...
+   block on stdin) reads `dbg:` commands, mutates the breakpoint table, walks the
+   parked stack for `inspect`, and resumes the target. The debugger's state is
+   guarded by a mutex; a condition variable hands pause/target events between the
+   scheduler thread and the controller.
+
+**Why park instead of run‑to‑completion or block in place:** a breakpoint must
+suspend just the target process without blocking a scheduler worker (the spike
+proved an in‑worker block deadlocks the VM). `DEBUG_PAUSED` + a separate controller
```
floitsch, line 87: I'm guessing this is the "deadlock detected" by the scheduler.
We could let the scheduler check whether a process is parked for debugging and avoid the "deadlock detected" message then.
Would that make the implementation simpler? (Would require fewer changes to the scheduler).

docs/debugger.md
```diff
@@ -0,0 +1,171 @@
...
+out of system/service frames.
+
+## Register values
+
+`dbg:inspect` answers with `dbg:stack off=<bci> r<slot>=<value> …` for the parked
+frame. `emit_stack` formats each register by type so the operator sees real values:
+
+- `null`, `true`, `false`; integers (smi) and doubles as numbers.
+- Strings via `emit_string`: a double‑quoted, escaped token (`\" \\ \n \r \t`,
+  `\xNN` for control chars), capped at 128 chars — so a value with spaces stays
```
floitsch, line 102: capping at 128 chars feels annoying. What is the reason for that?

docs/debugger.md
```diff
@@ -0,0 +1,171 @@
...
+- **Device/FreeRTOS.** The debugger is host‑only today (a POSIX controller thread on
+  stdin). Device debugging needs the in‑image supervisor and new VM primitives.
+- **Named locals**, conditional breakpoints, watchpoints, expression evaluation
+  beyond `inspect`, and DAP/editor integration are out of scope.
+
+## Build & run
+
+```bash
+# Build the host VM/SDK after changing src/debugger.cc (or .h):
+cd build/host && ninja sdk/bin/toit
```
floitsch, line 160: `make` or `make sdk` is the canonical way to rebuild the SDK.

docs/debugger.md
```diff
@@ -0,0 +1,171 @@
...
+# After changing tools/toitp.toit (a Toit tool compiled into the snapshot):
+cd build/host && ninja generated/toit.snapshot sdk/bin/toit   # ninja, not make
+```
+
+Run a snapshot under the debugger directly with
+`build/host/sdk/lib/toit/bin/toit.run --debug <snapshot>` and drive it over stdin,
```
floitsch, line 166: I don't like exposing `toit.run`. It should be accessible through `toit`. So `toit run --debug --`.

src/CMakeLists.txt
```diff
@@ -46,6 +46,8 @@ file(GLOB toit_core_SRC
   "*.c"
   "*.cc"
   )
+# Note: debugger.cc (the in-image bytecode debugger) is picked up by the glob
```
floitsch, line 49: No need to add this comment. (would be funny to use a glob and then mention each picked up file in a comment...)

src/debugger.cc
```diff
@@ -0,0 +1,394 @@
+// Copyright (C) 2026 Toitware ApS.
```
floitsch, line 1: Use "Toit contributors" instead. (here and in any other new file).

floitsch:
Sorry it took me so long to review.
I haven't fully reviewed yet, but added some comments on the general architecture approach first.
Read through those first, and discuss with me. Once we converge on something I will review the actual code.

PR #3086, opened 2026-06-30
Use the partition table to determine the ESP32 flash size.

floitsch:
The bootloader stores the flash size in its image header and the ROM rejects the whole partition table -- causing a silent boot loop -- if a partition extends beyond it. That previously forced users onto an envelope built for the exact flash size (e.g. a dedicated 16MB envelope) when using a big partition table.

build-esp32-image now rounds the partition table's extent up to a supported flash size (never below the size the envelope was built for) and stamps it into the bootloader image header and the flashing settings, so a stock envelope can be flashed with a bigger partition table. The flash command additionally detects the device's flash size (best-effort, via 'esptool flash-id' captured with host.fork and bounded by a timeout) and aborts if the chip is too small; detection failures only warn and never block flashing.

Add a 'check-esp-idf-assumption' make target that fails if CONFIG_ESPTOOLPY_FLASHSIZE is read in runtime sources, since this relies on the flash size coming solely from the bootloader image header (g_rom_flashchip.chip_size).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -634,6 +634,11 @@ jobs:
           esp32: true
           cache-key-prefix: esp32
 
+      # Verify that the firmware tool's flash-size assumption still holds: the
+      # flash size must not be read from CONFIG_ESPTOOLPY_FLASHSIZE at runtime.
+      - name: Check esp-idf assumptions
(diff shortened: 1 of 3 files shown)


floitsch:
TBR.

PR #3087, opened 2026-07-01
Let peripherals take GPIO pin numbers instead of gpio.Pin.

floitsch:
Peripherals (ADC, DAC, PWM, touch, I2C, I2S, SPI, UART, RMT and
pulse-counter) now accept an integer GPIO number and reserve/release the
pin from a shared pool themselves, instead of taking a gpio.Pin owned on
the Toit side. This lets a pin be shared across containers and avoids
close-ordering bugs.

The gpio.Pin form is deprecated: the affected parameters became `any`
with `__TYPE-MIGRATION__` annotations, and the toitdocs point users at the
integer form. Nullable pin parameters migrate to `int?`.

Pins are passed as a single signed integer (see `to-pin-num_` in
lib/gpio/gpio.toit): -1 means "no pin"; a value >= 0 tells the peripheral
to reserve the pin; a value <= -2 is a deprecated gpio.Pin that owns its
own reservation and must not be reserved by the peripheral. The shared
pool and the transactional reserver live in src/resources/gpio_esp32.h.

Add a TouchResource so touch pads participate in the reservation scheme.
Add a negative test for the migration diagnostics and a hardware test for
pin reservation.


floitsch:
TBR.

PR #3089, opened 2026-07-15
Drop the Go and xxd build dependencies

floitsch:
## Summary

Drops the last two non-CMake build dependencies from the tree: **Go** and **xxd**. Both
`golang-go` and `xxd` disappear from `debian/control`'s `Build-Depends`.

## Package manager as a library

Package management is now pure Toit. Instead of shelling out to the Go-built `toit.pkg`
executable, `toit pkg` runs the native `tools/pkg` implementation in-process:

- `tools/toit.toit` imports the pkg library and adds `pkg.build-command` directly,
  replacing the duplicated command tree and `run-pkg-command`.
- `pkg.toit` exposes `build-command`; the trusted-root installation moves into the
  `PkgCommand` base constructor, so it happens for the commands that need the network.
- The pkg sources join the `tools` package: its dependencies move into
  `tools/package.yaml` (adding `lockfile`, bumping `cli`) and the nested
  `tools/pkg/package.{yaml,lock}` are removed.
- The Go build (`build_go_tools`) and the `tools/tpkg` sources are removed, along with
  the `golang-go` build dependency and the workflow / `dh_dwz` bits that referenced the
  Go executable.

The single-file git registry cache also gains an `.ar` extension, so it no longer
collides with the old directory-based cache — which otherwise made existing users hit a
`cache entry is a directory` error.

## Pure-CMake binary embedder

The embedded snapshot/image binaries were turned into C arrays with `xxd -i`, which
required xxd as a build dependency. A small CMake script (`tools/embed-binary.cmake`)
generates the arrays instead.

The generator converts the whole file in a single `string(REGEX REPLACE)` pass over the
`file(READ ... HEX)` output (~1.3s for a 1 MB snapshot). It emits the same symbols
`xxd -i` did, so the `extern` declarations are unchanged. Being pure CMake, it needs no
external tool and works with any compiler and C++ standard — unlike `#embed`, which
would require GCC 15 / Clang 19 and so break older toolchains, including Debian's.


tools/pkg/commands/base_.toit
```diff
@@ -29,6 +30,9 @@ class PkgCommand:
   auto-sync_/bool
 
   constructor invocation/cli.Invocation:
+    // Install the trusted roots so that the package manager can talk to
+    // registries and download packages over HTTPS.
+    certificate-roots.install-all-trusted-roots
```
floitsch: Not a huge fan of installing the certificate roots in the constructor here.
I think they should be installed in the main functions that use the pkg manager. Could be convinced otherwise. But even if they are installed by the 'pkg' code, it shouldn't be in the `PkgCommand` constructor.

src/CMakeLists.txt
```diff
@@ -440,22 +437,27 @@ if (NOT "${TOIT_SYSTEM_NAME}" MATCHES "esp32")
...
+    # Embed the snapshot/image binaries into C++ source files instead of using
+    # `xxd -i` (see `tools/embed-binary.cmake`). This is a pure-CMake generator,
+    # so it needs no external tool and works with any compiler and C++ standard.
     add_custom_command(
       OUTPUT "${embedded_snapshot_cc}"
-      # Note: we can't use an absolute path for the '-i' argument..
-      COMMAND xxd -i "${snapshot_filename}" "${embedded_snapshot_cc}"
-      DEPENDS "${embedded_snapshot}"
-      WORKING_DIRECTORY "${GEN_DIR}"
+      COMMAND "${CMAKE_COMMAND}"
```
floitsch, line 445: This launches a fresh cmake.
I believe we only use the embed-binary functionality in this CMakeLists.txt.
Feels simpler to just move the function here.

tools/pkg/commands/base_.toit
floitsch: Agreed. Moved it out of the constructor to the point of network use: it's now in `GitProtocol_`'s constructor, right before `net.open`. Since `protocol_` is a lazily initialized module-level global, the roots are installed exactly once, and only when the package manager actually touches the network — so `clean`/`list`/`version` no longer mutate the process's TLS trust. Done in 8319ad66.

Verified end-to-end against the real registry over HTTPS (the pkg tests only use `http://localhost`, so they don't exercise TLS): `pkg registry sync` succeeds with the install and fails with `NOT_TRUSTED` when I temporarily remove it.

src/CMakeLists.txt
floitsch: You're right that it's only used here and that it spawns a fresh cmake. But it can't become a plain function called from this `CMakeLists.txt`: the script runs as an `add_custom_command`, i.e. at build time, when `${embedded_snapshot}`/`${embedded_image}` actually exist (they're produced earlier in the build by `toit.compile` and `snapshot-to-image`). A function called from `CMakeLists.txt` runs at configure time, when those files don't exist yet, so `file(READ ...)` would have nothing to read. `cmake -P <script>` is the standard way to run cmake code as a build step, and `-P` needs a real file — so the separate file is required rather than stylistic. (The fresh process is also not a regression: `xxd -i` spawned one per file too.)

Leaving it as-is for now — as you say, we may want to reuse it elsewhere anyway.

floitsch:
TBR.

PR #3090, opened 2026-07-16
Add shell completion suggestions to the pkg commands.

floitsch:
The pkg commands now provide completion candidates through the cli
package's completion support:

- `pkg install` completes package names and URLs (with descriptions)
  from the locally cached registries. A '@' in the word completes the
  version part (`pkg install morse@<tab>`). Ambiguous names are only
  offered as URLs. With `--local` the shell falls back to file
  completion.
- `pkg uninstall` completes the prefixes of the installed packages,
  honoring `--project-root`.
- `pkg describe` completes package URLs, and completes the version
  argument with the versions of the package given as first argument.
- `pkg registry remove`, `pkg registry sync` and `pkg list` complete
  registry names.
- `--project-root` and describe's `--out-dir` complete directories.

Completion callbacks run while the user is waiting for the shell
prompt, so they never go to the network: registries whose content is
not cached locally are skipped (new `Registry.is-cached`), the
registries are never auto-synced, and any error or timeout results in
an empty candidate list.

The gold tester gained a `complete` command that runs
`pkg.toit __complete` in a subprocess, and a new completion gold test
covers the behavior, including that an unsynced registry produces no
candidates.

### DIFF
--- a/src/resources/pipe_win.cc
+++ b/src/resources/pipe_win.cc
@@ -510,71 +510,6 @@ bool is_inherited(Object* object) {
 }
 const int MAX_COMMAND_LINE_LENGTH = 32768;
 
-// Appends a single wide character to `command_line`, returning false if that
-// would overflow the `capacity` (excluding the terminating null).
-static bool append_wchar(wchar_t* command_line, int* pos, int capacity, wchar_t c) {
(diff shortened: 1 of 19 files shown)


floitsch:
TBR.

PR #3091, opened 2026-07-17
Add 'uart.Port.console' to read input from the console UART.

floitsch:
The console UART (the one used for logging and `print`) can now be opened with `uart.Port.console`. It keeps the configuration it was given during boot; only the UART driver is installed, so RX becomes interrupt driven while system output continues through the polling VFS path (the same coexistence ESP-IDF's own console REPL uses). A dedicated one-slot pool guards single-open and returns the port on close, and in-flight TX is drained before the driver install since that resets the TX FIFO.

This is the base for driving hardware tests over the serial connection instead of WiFi (following the EC618 tester's serial mini-jag model): the esp-tester now answers `UART-INPUT-REQUEST: <payload>` marker lines by writing the payload back over the serial port, and the new `uart-console-test` exercises the full loop (open, double-open rejection, host echo received byte-exact, driver-side write, close/reopen).

Verified on the hardware rig: `uart-console-test.toit-esp32` passes, and the remaining 15 uart hardware tests pass on the new firmware (the teardown change touches every UART's close path). Skipped on esp32s3: that rig's board1 is connected through USB-Serial-JTAG, so host writes don't reach UART0 RX — S3 input needs a usb_serial_jtag equivalent as a follow-up.

### DIFF
--- a/src/resources/uart_esp32.cc
+++ b/src/resources/uart_esp32.cc
@@ -24,6 +24,7 @@
 #include "soc/uart_periph.h"
 #include "esp_log.h"
 #include "esp_rom_gpio.h"
+#include "esp_rom_uart.h"
 #include "esp_timer.h"
 #include "freertos/FreeRTOS.h"
 
@@ -88,6 +89,14 @@ static ResourcePool<uart_port_t, kInvalidUartPort> uart_ports(
 #endif
 );
 
+#ifdef CONFIG_ESP_CONSOLE_UART
+// The console UART is excluded from `uart_ports` above. It can only be
+// opened through the dedicated console primitive.
+static ResourcePool<uart_port_t, kInvalidUartPort> console_uart_port(
+    static_cast<uart_port_t>(CONFIG_ESP_CONSOLE_UART_NUM)
+);
+#endif
+
 class UartResourceGroup;
 
 class UartResource : public EventQueueResource {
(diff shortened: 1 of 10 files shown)


src/resources/uart_esp32.cc
```diff
@@ -498,6 +513,73 @@ PRIMITIVE(create_path) {
...
+
+  ByteArray* proxy = process->object_heap()->allocate_proxy();
+  if (proxy == null) FAIL(ALLOCATION_FAILED);
+
+  // The console UART keeps the configuration (baud rate, pins, ...) it was
+  // given during boot; only the driver is installed, which makes RX
+  // interrupt driven. System output (logging, `print`) keeps going through
+  // the polling VFS path and thus doesn't need the driver.
+  // Installing the driver resets the TX FIFO, so drain in-flight output first.
+  esp_rom_output_tx_wait_idle(port);
```
floitsch: I guess there could be a race condition?
If we wait for idle, and another container prints something we might still drop a few bytes.
If there isn't a good way around that, then we should document that.

tests/hw/esp32/fail.cmake
```diff
@@ -15,6 +15,10 @@
 
 
 set(TOIT_SKIP_TESTS
+  # On the test rig the S3's board1 is connected through the USB-Serial-JTAG
+  # console, so data written by the tester doesn't reach the UART0 RX pin.
+  uart-console-test.toit-esp32s3
```
floitsch: I don't think that's true.

tests/hw/esp32/uart-console-test.toit
```diff
@@ -0,0 +1,42 @@
...
+import ..esp-tester.shared show UART-INPUT-REQUEST
+import .test
+
+MESSAGE ::= "hello-console"
+
+main:
+  run-test: test
+
+test:
+  port := uart.Port.console
```
floitsch, line 57: Also add a test (or modify this one) that we can change the uart baudrate of the console. (obviously needs to work together with the tester).

PR #3093, opened 2026-07-18
Cache the parsed registry descriptions.

floitsch:
Loading a git registry parsed every description file out of the git pack on every invocation (roughly 1.3 seconds for the toit registry). This parses them when the registry is synced instead, and stores the parsed form in a separate cache entry (ubjson, keyed by the pack hash).

Loads fall back to parsing the pack when the entry is missing, stale or corrupt, and repair the entry in passing, so caches written by older versions keep working and get upgraded on first use. Older versions of the pkg tool ignore the new entry entirely.

With this, `pkg install <tab>` on the real registry goes from ~1.3s to ~0.2s.

Also silences the default logger during completions: the cache's file-lock logging would otherwise end up in the completion candidates.

Builds on the shell-completion work from #3090.

### DIFF
--- a/tools/pkg/registry/git.toit
+++ b/tools/pkg/registry/git.toit
@@ -94,27 +94,36 @@ class GitRegistry extends Registry:
     registry-content := content  // Also sets $loaded-hash_.
     key := descriptions-cache-key_
     hash := loaded-hash_
-    if cache.contains key:
-      catch:  // On any error fall back to parsing the description files.
-        encoded := cache.get key: unreachable
-        decoded := ubjson.decode encoded
-        if decoded[DESCRIPTIONS-VERSION-KEY_] == DESCRIPTIONS-FORMAT-VERSION_ and
-            decoded[DESCRIPTIONS-HASH-KEY_] == hash:
-          descriptions := decoded[DESCRIPTIONS-KEY_].map: | description-content/Map |
-            Description description-content --path="<cached>" --ui=ui_
-          return DescriptionUrlCache.filled descriptions
+    cached/DescriptionUrlCache? := null
+    cache-error := catch:
+      encoded := cache.get key: | store/FileStore |
+        descriptions := DescriptionUrlCache registry-content --ui=ui_
+        store.save (encode-description-cache_ descriptions hash)
+      decoded := ubjson.decode encoded
+      if decoded[DESCRIPTIONS-VERSION-KEY_] == DESCRIPTIONS-FORMAT-VERSION_ and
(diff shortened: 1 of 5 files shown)


tools/pkg/registry/git.toit
```diff
@@ -57,10 +67,55 @@ class GitRegistry extends Registry:
...
+    content. Otherwise parses all description files from the $content
+    and updates the cached parsed form, so that later invocations (in
+    particular shell completions) don't need to parse them again.
+  */
+  load-description-cache_ -> DescriptionUrlCache:
+    registry-content := content  // Also sets $loaded-hash_.
+    key := descriptions-cache-key_
+    hash := loaded-hash_
+    if cache.contains key:
+      catch:  // On any error fall back to parsing the description files.
```
floitsch: I don't like `catch` that are silently eaten.
Maybe let the `throw` through and then catch it somewhere where we have the ui to report it? (at least in verbose mode).

tools/pkg/registry/git.toit
```diff
@@ -57,10 +67,55 @@ class GitRegistry extends Registry:
...
+        encoded := cache.get key: unreachable
```
floitsch: if it's unreachable, then just use `cache[key]`.

tools/pkg/registry/git.toit
```diff
@@ -57,10 +67,55 @@ class GitRegistry extends Registry:
...
+  Uses the cached parsed form if it matches the current registry
+    content. Otherwise parses all description files from the $content
+    and updates the cached parsed form, so that later invocations (in
+    particular shell completions) don't need to parse them again.
+  */
+  load-description-cache_ -> DescriptionUrlCache:
+    registry-content := content  // Also sets $loaded-hash_.
+    key := descriptions-cache-key_
+    hash := loaded-hash_
+    if cache.contains key:
```
floitsch: Why ask for the existence first, if we fill it up later anyway.
The typical cache pattern is to get it with a block that fills it if it isn't there yet.

If it's to `catch` the failed `update` then we should catch that outside and report it with the `ui` (if verbose or so).

tools/pkg/registry/git.toit
floitsch: Done.

tools/pkg/registry/git.toit
floitsch: Done.

tools/pkg/registry/git.toit
floitsch: Done.

floitsch:
TBR.

PR #3094, opened 2026-07-19
ESP32: use semaphores for condition-variable wakeups

floitsch:
FreeRTOS task notifications are a shared per-task slot. A timed wait or another notification user can leave bits that an unrelated condition-variable wait consumes, and the SIGNAL_ALL relay can then target the wrong waiter list.

Give each task a dedicated binary wake semaphore instead. signal wakes the first waiter, signal_all wakes every waiter directly, and spawned tasks release their semaphore on exit.


floitsch:
@coderabbit review

floitsch:
TBR.

PR #3095, opened 2026-07-19
uart: keep filling after partial writes

floitsch:
PR #3060 fixed write --flush so it flushes once after the complete write instead of flushing each partial chunk. One part of the EC618 gap-free UART fix was still missing: UartWriter.write slept after every partial write, preventing drivers with a second TX staging buffer from receiving the next chunk before the current chunk completed.

Retry immediately while try-write makes progress, and wait for the TX event only after a zero-progress write.

Validation:
- SDK build passed.
- tests/uart-test.toit compiles successfully.
- The runtime PTY test hits its existing 40-second timeout both with this change and on unmodified master in this environment.
- The originating EC618 gap-free TX hardware test passes through 921600 baud.


floitsch:
TBR.

PR #3096, opened 2026-07-19
pkg: use UI logger for package cache lock

floitsch:
Pass the project UI logger to the filesystem lock so lock messages respect the pkg command verbosity level. Adds a focused regression test for debug-level lock acquisition and release messages. Tested with Toit analysis, the focused regression test, and tests/pkg/package-test.toit.


floitsch:
TBR.

PR #3104, opened 2026-07-21
Prepare source builds for Homebrew

floitsch:
## Summary
- prevent release packaging jobs from overwriting existing tarball and Debian assets
- use the standard CMake BUILD_TESTING switch
- keep tests enabled by default, while allowing package managers to configure with -DBUILD_TESTING=OFF

## Verification
- configured, built, and installed the full SDK locally with BUILD_TESTING=OFF
- ran the installed toit binary and verified `toit info sdk` resolves the installed prefix

This supports the companion formula in toitlang/homebrew-toit#15.

### DIFF
--- a/.github/workflows/tar-ball.yml
+++ b/.github/workflows/tar-ball.yml
@@ -98,7 +98,7 @@ jobs:
           file: ${{ steps.create-tarball.outputs.archive_path }}
           asset_name: ${{ steps.create-tarball.outputs.archive_name }}
           tag: ${{ github.event.release.tag_name }}
-          overwrite: true
+          overwrite: false
 
     outputs:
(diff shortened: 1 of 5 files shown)


floitsch:
CI note: all compiler/platform builds, ESP32/cross builds, source archive/deb installs, and installed-binary tests passed. The `external-and-flaky` aggregate is red because every external shard hits the pre-existing `toit-datacake/examples` lock-file-without-specification failure; two Windows flaky shards also failed. The matching master commit (`ed360569`, run 29855493197) likewise failed only `external-and-flaky`. The companion Homebrew tap PR is merged and passed Intel macOS, Apple Silicon, and Linux.

PR #3112, opened 2026-07-31
pkg: unwind aborts before exiting

floitsch:
## Summary

- update the tool dependency from pkg-cli 2.16.0 to 2.17.0
- rely on the CLI package default abort behavior to unwind finalizers before Command.run exits with status 1
- add a real child-process regression test proving an abort releases the package-cache file lock

The temporary package-manager-specific Ui subclass and outer catch are no longer needed now that the behavior is provided by pkg-cli itself.

## Testing

- full SDK build with pkg-cli 2.17.0
- info-test and abort-cleanup-test
- complete tests/pkg suite: 50/50 passed


floitsch:
We should really fix the cli package.
Buildbot is red.
otherwise lgtm.

PR #3114, opened 2026-07-31
pkg: refresh dependencies of local packages

floitsch:
## Summary

- recompute dependency solutions when a project contains local packages
- retain the existing lockfile fast paths for projects containing only remote packages
- cover a changed local package that introduces a dependency on another major version

## Testing

- install-gold-test.toit
- complete tests/pkg suite

### DIFF
--- a/tests/pkg/install-gold-test.toit
+++ b/tests/pkg/install-gold-test.toit
@@ -20,6 +20,21 @@ test tester/GoldTester:
     ["exec", "main.toit"],
   ]
 
+  file.write-contents --path="$tester.working-dir/target/package.yaml" """
+      name: target
+      description: desc
+      dependencies:
+        sub:
+          url: localhost:$tester.port/pkg/sub
+          version: ^2.0.0
+      """
+
+  tester.gold "15-local-change" [
+    ["// Target should now use sub-2, while bar should continue to use sub-3."],
+    ["pkg", "install"],
+    ["package.lock"],
+  ]
+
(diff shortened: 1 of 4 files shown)


tests/pkg/install-gold-test.toit
```diff
@@ -20,6 +20,20 @@ test tester/GoldTester:
...
+      description: desc
+      dependencies:
+        sub:
+          url: localhost:$tester.port/pkg/sub
+          version: ^2.0.0
+      """
+
+  tester.gold "15-local-change" [
+    ["pkg", "install"],
+    ["package.lock"],
```
floitsch, line 35: Add a comment to explain what should be changed (just to make it easier to check the gold file).

tools/pkg/project/project.toit
```diff
@@ -220,11 +220,13 @@ class Project:
       save
 
   install --recompute/bool --registries/Registries -> none:
-    if not recompute and lock-file and lock-file.is-downloaded:
+    if not recompute and lock-file and
```
floitsch: Maybe make a helper function for 'recompute, lock-file, and not has-local-packages'. If it has a good name, it might be easier to read. (and then use it below as well).

floitsch: done

tests/pkg/install-gold-test.toit
floitsch: done

PR #3115, opened 2026-07-31
pkg: document lockfile package IDs

floitsch:
## Summary

- document that remote package IDs carry a stable major-version suffix because multiple majors can coexist
- document numeric disambiguators for normalized URL and local basename collisions
- assert the distinct major-version IDs in the install regression test

## Testing

- install-gold-test.toit
- complete tests/pkg suite


floitsch:
LGTM.

PR #3154, opened 2026-08-06
Add asynchronous I2C target support

floitsch:
## Summary

- Add an event-driven `i2c.Target` API with 7-bit and 10-bit addressing, broadcast support, bounded send/receive buffers, overflow reporting, and explicit read-request handling.
- Keep ESP-IDF ISR callbacks allocation-free; callbacks only update native state and enqueue preallocated event words.
- Make every primitive return immediately. Toit tasks wait through `ResourceState_`, outside the primitive call.
- Use "target" throughout the public Toit API while retaining ESP-IDF's existing names internally.

## ESP-IDF dependency

Builds on #3203. The required ESP-IDF patch head is already on master.

## Testing

- Two-device hardware tests on ESP32 and ESP32-S3.
- Covers 7-bit and 10-bit addresses, broadcast, repeated starts, request/receive ordering, buffer boundaries and overflow, task cancellation, closure, and bus reuse.


floitsch:
@coderabbit review

lib/i2c.toit
```diff
@@ -34,6 +35,191 @@ In case of the Bosch [BME280 sensor](https://cdn.sparkfun.com/assets/e/7/3/b/1/B
...
 DEFAULT-FREQUENCY ::= 400_000
 
+/** Default native buffer size for an I2C target. */
+DEFAULT-TARGET-BUFFER-SIZE ::= 256
+
+/**
+An addressable I2C target.
+
+A target receives complete controller write transactions with $read and queues
+  bytes for controller read transactions with $write. Native primitives never
```
floitsch: No need to state this explicitly. That's just how all of Toit works.

lib/i2c.toit
```diff
@@ -34,6 +35,191 @@ In case of the Bosch [BME280 sensor](https://cdn.sparkfun.com/assets/e/7/3/b/1/B
...
+    native space used to queue unread transactions.
+
+  If $pull-up is true, the weak internal pull-ups are enabled. External
+    pull-ups are recommended for normal and fast bus speeds.
+
+  $broadcast makes the target acknowledge the general-call address. It is not
+    supported by every ESP32 variant and cannot be combined with a 10-bit
+    address.
+
+  $allow-power-down allows ESP-IDF to power down the peripheral during light
```
floitsch: Toit doesn't yet really deal with light sleep. (I believe).
I would remove the comment from the toitdocs and remove the argument.
The primitive can still have the flag.

lib/i2c.toit
```diff
@@ -34,6 +35,191 @@ In case of the Bosch [BME280 sensor](https://cdn.sparkfun.com/assets/e/7/3/b/1/B
...
+    sleep on chips that support peripheral retention.
+  */
+  constructor
+      --sda/any
```
floitsch: Don't support gpio.Pin. It's deprecated for the other i2c code already.

lib/i2c.toit
```diff
@@ -34,6 +35,191 @@ In case of the Bosch [BME280 sensor](https://cdn.sparkfun.com/assets/e/7/3/b/1/B
...
+      state_.clear-state REQUEST-STATE_
+
+      written := i2c-target-write_ resource_ bytes offset
+      if written < 0:
+        yield
+        continue
+      offset += written
+      if offset == bytes.size: return
+
+      // Retry before waiting so a controller read that raced with the first
```
floitsch: I don't understand this comment/special case.

lib/i2c.toit
```diff
@@ -34,6 +35,191 @@ In case of the Bosch [BME280 sensor](https://cdn.sparkfun.com/assets/e/7/3/b/1/B
...
+    transaction. The original ESP32 cannot stretch for this event; responses
+    must be queued before the controller starts reading.
+  */
+  wait-for-read-request -> int:
+    while true:
+      if not resource_: throw "CLOSED"
+      state_.clear-state REQUEST-STATE_
+      count := i2c-target-take-request-count_ resource_
+      if count != 0: return count
+      count = i2c-target-take-request-count_ resource_
```
floitsch: why the duplication?

src/event_sources/ev_queue_esp32.cc
```diff
@@ -27,7 +27,12 @@
 
 // The max queue set size is the maximum number of events in the queue. This is used for the gpio queue,
 // up to two UART queues and the stop semaphore.
-#define MAX_QUEUE_SET_SIZE (GPIO_QUEUE_SIZE + 2 * UART_QUEUE_SIZE + STDIN_QUEUE_SIZE + 1)
+// Leave room for the length-one queues owned by peripherals such as I2S and
+// I2C targets. FreeRTOS requires a queue set to be sized for the sum of all
+// member queue lengths.
+#define PERIPHERAL_EVENT_QUEUE_COUNT 16
+#define MAX_QUEUE_SET_SIZE \
+  (GPIO_QUEUE_SIZE + 2 * UART_QUEUE_SIZE + STDIN_QUEUE_SIZE + PERIPHERAL_EVENT_QUEUE_COUNT + 1)
```
floitsch: Would be interesting to see if we forgot this for other events.

src/resources/i2c_esp32.cc
```diff
@@ -43,6 +49,124 @@ class I2cResourceGroup : public ResourceGroup {
     : ResourceGroup(process) {}
 };
 
+const word kTargetReceiveState = 1 << 0;
```
floitsch, line 55: should we create a separate file for the target?
or is there a lot of code that is easier to share with the controller code?

toolchains/esp32/sdkconfig
```diff
@@ -879,7 +879,7 @@ CONFIG_GPTIMER_OBJ_CACHE_SAFE=y
 #
 # CONFIG_I2C_ISR_IRAM_SAFE is not set
 # CONFIG_I2C_ENABLE_DEBUG_LOG is not set
-# CONFIG_I2C_ENABLE_SLAVE_DRIVER_VERSION_2 is not set
+CONFIG_I2C_ENABLE_SLAVE_DRIVER_VERSION_2=y
```
floitsch, line 882: look at what 'make menuconfig' does, since it affects the sdkconfig.default which need to be updated as well.

src/resources/i2c_esp32.cc
```diff
@@ -139,6 +263,186 @@ PRIMITIVE(init) {
...
+    .on_receive = target_receive_handler,
+  };
+  err = i2c_slave_register_event_callbacks(handle, &callbacks, resource);
+  if (err != ESP_OK) return Primitive::os_error(err, process);
+
+  resource->owned_pins().adopt(reserver);
+  reserver.keep();
+  group->register_resource(resource);
+  proxy->set_external_address(resource);
+  registered = true;
```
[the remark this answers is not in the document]
floitsch: All three queue-set conditions are already handled by this patch and the existing event source. on_register_resource drains the queue and retries xQueueAddToSet until it succeeds; this also safely handles an event arriving after callbacks are armed but before registration. on_unregister_resource removes the resource from the event-source list, drains its queue, and retries xQueueRemoveFromSet before ResourceGroup deletes the resource. Both return values are checked by those loops. Finally, PERIPHERAL_EVENT_QUEUE_COUNT adds 16 entries to MAX_QUEUE_SET_SIZE specifically for these length-one peripheral queues. Registering the resource before callbacks is therefore unnecessary and would complicate rollback if callback registration failed.

lib/i2c.toit
```diff
@@ -34,6 +35,191 @@ In case of the Bosch [BME280 sensor](https://cdn.sparkfun.com/assets/e/7/3/b/1/B
...
+      // Retry before waiting so a controller read that raced with the first
+      // attempt cannot leave this task asleep with available buffer space.
+      written = i2c-target-write_ resource_ bytes offset
+      if written < 0:
+        yield
+        continue
+      offset += written
+      if offset == bytes.size: return
+
+      state_.wait-for-state REQUEST-STATE_
```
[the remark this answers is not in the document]
floitsch: The interpretation of -1 is incorrect for this ESP-IDF implementation. i2c_slave_write returns ESP_ERR_TIMEOUT here only when its operation_mux cannot be acquired with the zero timeout. A full TX ring buffer still returns ESP_OK with write_len equal to zero. REQUEST-STATE_ is generated by a controller read and does not indicate that operation_mux became available, so waiting for REQUEST-STATE_ on -1 could sleep forever. The yield is a transient retry for lock contention; the normal zero-progress path reaches the request wait below.

lib/i2c.toit
```diff
@@ -34,6 +35,191 @@ In case of the Bosch [BME280 sensor](https://cdn.sparkfun.com/assets/e/7/3/b/1/B
...
+      result := i2c-target-receive_ resource_
+      if result: return result
+
+      // Close the race between clearing the state and reading the native
+      // message queue before going to sleep.
+      check-receive-overflow_
+      result = i2c-target-receive_ resource_
+      if result: return result
+
+      state_.wait-for-state RECEIVE-STATE_ | OVERFLOW-STATE_
```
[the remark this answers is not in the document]
floitsch: This is already handled by ResourceState_. ResourceState_ is a monitor, and wait-for-state_ awaits the condition "not resource_ or result != 0". dispose sets its resource_ to null inside that monitor, so waiters re-evaluate the condition and resume. They then loop, see Target.resource_ has been cleared by close, and throw CLOSED. The native notifier is for resource events; it is not the wake mechanism for a mutation of the monitor itself.

src/resources/i2c_esp32.cc
floitsch: There is little target/controller state worth sharing. The practical reason to keep them together is MODULE_IMPLEMENTATION: it declares the module’s primitive functions static and builds one table in this translation unit. A clean split would need primitive-registration/header plumbing, or an included implementation fragment. For roughly 300 target-specific lines I do not think that refactor belongs in this PR.

lib/i2c.toit
floitsch: There is a public light-sleep API in esp32.pm-configure, so Toit has some support. However, the shipped ESP32 sdkconfigs have CONFIG_PM_ENABLE disabled, and this target-retention path is not tested. I agree that exposing a new permanent Target option is premature: remove it from the public constructor/docs and keep the native parameter available with false passed for now.

lib/i2c.toit
floitsch: I agree that the second write is unnecessary. REQUEST-STATE_ is level-triggered in ResourceState_: after clear-state, a request racing with the write either sets the bit before wait-for-state or wakes it afterwards. The extra native call does not close a remaining lost-wakeup race. The same reasoning applies to the duplicated receive and request-count checks above.

lib/i2c.toit
floitsch: The special value is a nonblocking retry indication: the primitive maps ESP_ERR_TIMEOUT from i2c_slave_write to -1 because it calls ESP-IDF with timeout zero. In this driver, that timeout only means operation_mux was momentarily busy; it does not mean the TX buffer is full. A full buffer returns ESP_OK with written == 0. The yield retries mutex contention without blocking a primitive. Under Toit’s serialized primitive execution this should be very rare, but the handling is defensive.

src/event_sources/ev_queue_esp32.cc
floitsch: Yes. At this PR’s base, the old formula also omitted ESP-NOW’s length-3 event queue and the length-1 queues created per I2S and RMT resource. Unused UART capacity probably masked that in common configurations. The new 16-slot peripheral budget covers those existing users as well as I2C targets; registration already fails loudly if the shared budget is exhausted.

PR #3156, opened 2026-08-06
Migrate I2C controller operations to async

floitsch:
## Summary

- Migrate controller probe, read, write, and write-read operations to the asynchronous ESP-IDF driver while preserving the synchronous-looking Toit API.
- Serialize complete controller transactions per bus while keeping native primitive calls nonblocking.
- Copy moving-heap data into per-operation native buffers and release it after completion or synchronous abort.
- Finalize dispatched operations in a non-cancelable cleanup path; abort retires the transaction and then performs best-effort bus recovery.
- Expose per-device clock-stretch timeout and ACK-check policy, preserve 7-bit/10-bit addressing, and return explicit `I2C_NACK` / `I2C_TIMEOUT` errors.
- Route synchronous `ESP_ERR_NO_MEM` through the existing retryable OOM path after cleaning all partial state; callbacks allocate nothing and cannot report OOM.
- Keep Toit I2C callbacks in IRAM only when `CONFIG_I2C_ISR_IRAM_SAFE` is enabled.

## Dependencies

- Builds on #3155.
- The required ESP-IDF patch head is already on master.

## Testing

- Host, ESP32, and ESP32-S3 builds.
- `toit analyze -Werror` for changed libraries and two-device hardware tests.
- Paired-board coverage includes present/missing probes and scan, NACK and post-NACK recovery, ACK-check disable, 7-bit/10-bit device identity, 1–1024-byte transfers, FIFO/command-boundary sizes, concurrent task serialization, cancellation during clock stretching, and bus reuse.
- Complete I2C/SPI hardware selections pass on ESP32 and ESP32-S3 both after setup and immediately without fixture setup.
- `git diff --check` in both repositories.


floitsch:
@coderabbit review

lib/i2c.toit
```diff
@@ -301,10 +301,16 @@ class RegisterTarget:
 /**
 Bus for communicating using I2C.
 
-The communication is synchronous.
+Controller operations suspend only the calling Toit task. They are dispatched
```
floitsch: This is default behavior in Toit. We shouldn't mention it.

lib/i2c.toit
```diff
@@ -399,45 +426,71 @@ class Bus:
...
   Variant of $(device i2c-address --frequency --address-size) that uses the
     default frequency given to the bus at construction.
   */
-  device i2c-address/int --address-size/int=7 -> Device:
-    return device i2c-address --frequency=frequency_ --address-size=address-size
+  device i2c-address/int
+      --address-size/int=7
+      --timeout-us/int=100_000
+      --disable-ack-check/bool=false
+      -> Device:
```
floitsch: The return type should be on the first line (`device i2c-address/int -> Device`)

src/resources/i2c_esp32.cc
```diff
@@ -779,16 +885,100 @@ PRIMITIVE(bus_close) {
...
+    default: return 3;
+  }
+}
+
+PRIMITIVE(bus_probe_finish) {
   ARGS(I2cBusResource, resource);
+  if (!resource->operation_in_flight()) FAIL(INVALID_STATE);
+  int result = controller_result(resource->completion_event());
+  resource->finish_operation();
+  if (result == 3) FAIL(INVALID_STATE);
```
floitsch: magic constant?

src/resources/i2c_esp32.cc
```diff
@@ -850,44 +1045,131 @@ PRIMITIVE(device_close) {
...
 
+static esp_err_t prepare_controller_operation(I2cBusResource* bus,
+                                              const uint8_t* tx,
+                                              size_t tx_length,
+                                              size_t rx_length) {
+  if (bus->operation_in_flight()) return ESP_ERR_INVALID_STATE;
+
+  uint8_t* tx_buffer = null;
+  if (tx_length > 0) {
+    tx_buffer = static_cast<uint8_t*>(heap_caps_malloc(
```
floitsch: unvoid_cast? (here and in other places?)

floitsch:
CodeRabbit follow-up after the latest review commits:

- Completion wait: fixed by making it deadline-respecting and running the new bounded native abort on cancellation; ESP-IDF PR #130 fixes the NACK and BUS_BUSY paths that could previously lose completion. A derived whole-transaction timeout would be incorrect because `scl_wait_us` is per stretch.
- Close cancellation safety: fixed for both bus and device teardown.
- Callback unregistration: no code change. In this ESP-IDF revision it can return `ESP_ERR_INVALID_STATE` only when the bus has no async queue; every resource here is constructed with one. Cleanup is serialized after transaction retirement, and refusing cleanup on an impossible invariant failure would leak.
- Probe speed: fixed by documenting ESP-IDF's intentional 100 kHz probe rate.
- Finish closed guards: fixed defensively in all three finish primitives.
- Test `catch --unwind`: fixed.
- Dead/redundant test control flow: fixed, including removal of the last redundant architecture guard.

PR #3158, opened 2026-08-06
Add asynchronous SPI target support

floitsch:
Stacked on #3156. The required ESP-IDF patch head is already on master.

This introduces an ESP32 `spi.Target` API for transaction-sized controller/target exchanges. Native primitives only allocate, queue with a zero timeout, copy completed data, or tear down resources; waiting happens through `ResourceState_` and suspends only the calling Toit task.

Highlights:

- independent MOSI/MISO, modes 0 through 3, independent RX/TX bit order, DMA selection, fill byte, early controller termination, and configurable maximum transaction size
- preallocated native buffers with retryable OOM cleanup before dispatch and allocation-free ISR callbacks
- explicit `PendingExchange` so applications can assert a ready GPIO after the peripheral is armed
- classic ESP32 DMA restrictions validated at both the Toit and native API boundaries
- controller CS setup/hold options needed for target timing tests
- non-cancelable target state transitions and cleanup
- two-board hardware coverage for modes, direction, bit order, DMA/non-DMA, 50 kHz through 2 MHz, sizes through 4092 bytes, fill behavior, and early termination

The ESP-IDF support and subsequent I2C/SPI hardening are included in the patch head now on master. The target API intentionally exposes complete transactions only; it does not add half-buffer streaming.


floitsch:
@coderabbit review

lib/spi.toit
```diff
@@ -6,6 +6,13 @@ import gpio
 import io
 import serial
 import monitor
+import system
```
floitsch, line 9: off-topic:
some interrupt function must live in iram. However, we would also like to put performance-sensitive functions there.
Is there a way to move functions into iram while we don't use the peripherals (like i2c, spi, uart, rmt, ...). Not sure if it's possible to make the iram that flexible, but in theory we would like to move the interrupt handlers to iram only when needed.

spi.toit
69: Don't mention "native primitives". Primitives in toit must never block. So no need to mention this here.

118: all gpio arguments should be typed as `int?`. The backwards-compatibility of accepting Gpio instances isn't necessary here.
This affects a few lines here and there (no `gpio.to-pin-num_` and the primitive doesn't need to care for <2 values)

127: prefer if not 0 <= mode <= 3: throw "INVALID_ARGUMENT".

128: once we accept only int?, use `if not mosi and not miso: ...`.
Same for 133. and 135

154: how does $exchange know whether the controller ends a transaction earlier? SPI doesn't really have "transactions" afaik.

160: not being able to cancel is really bad. What if I have a `with-timeout`? Does a shutdown of the container reset everything?
...
This is something we need to look into For the i2c we ended up changing the esp-idf to support cancelation.

167: the `-> ByteArray` should be on line 163.

180: why do we expose `start-exchange`?
184: `-> PendingExchange` should be in line 180.

188: `if not 0 < transfer-size <= max-transfer-size: ...`

205: again: not being able to abort something feels very wrong.

234: also: we really want to abort things (and release the pins...)

241: why is there an interface PendingExchange? Why not just the class?
I wouldn't expose the 'start-exchange'. Do we still need the 'PendingExchange'?

spi_esp32.cc:
95: if we free the rx_buffer_ but the operation is active, don't we risk writing into freed memory?

spi_esp32.cc: might not apply, but we need a certain amount of space in the event-queue. That was previously handled in one header file far away from the actual peripherals. Now we have added static_asserts in every peripheral to ensure that the corresponding constant is correct (also reminding us that we need to update the queue size).


244: is this just a `spi_host_devices.take`?

419: are there constants in the esp-idf for this? (setup_cycles/hold_cycles)

spi_esp32.h. Do we need to have the classes in the header file?

floitsch:
CodeRabbit follow-up after the latest review commits:

- Finalizer/resource retention: the claimed GC race cannot occur because an executing `exchange` keeps its receiver strongly reachable. Forced process teardown is now separately hardened: the native destructor aborts and waits for retirement before releasing buffers, host, and pins.
- Classic ESP32 DMA truncation: fixed by aligning the transferred byte count before clamping to the requested receive size.
- ISR wake latency: fixed with `portYIELD_FROM_ISR` for target callbacks.
- Generated IRAM settings: fixed for C3, C6, P4, S2, and S3; each slave ISR setting now has its matching HAL setting. Classic ESP32 intentionally uses the non-IRAM target path because of its DMA restrictions.

PR #3172, opened 2026-08-19
Add audio ESP32 hardware test

floitsch:
add an ESP32 audio hardware test


floitsch:
Moving this opt-in hardware test to toit-pkg/toit-audio, since the SDK's default ESP32 envelopes do not enable the audio primitives.

PR #3184, opened 2026-08-26
Add SDK support for device memory inspection

floitsch:
This extracts the SDK/runtime portion of the device-memory inspector work so the required SDK surface can be reviewed independently of the tool itself.

## Included

- source-level parameter names and local stack-slot lifetimes in snapshot debug metadata
- an experimental privileged ESP32 terminal UART memory-dump primitive
- optional stable VM-state transition checkpoints (`TOIT_VM_STATE_CHECKPOINTS`)
- GC-metadata and Bluetooth allocator attribution
- snapshot/memory-report decoding support for the new metadata and tags

## Deliberately excluded

- the inspector model, query API, server, and HTML viewer
- UART/QEMU acquisition clients and capture fixtures
- the GDB RSP package
- the special capture boot image
- build-time GDB/ELF extraction and inspector envelope entries

The last item should return as generic ELF/DWARF and envelope metadata support rather than make the SDK build import the inspector package.

This is 28 files, 1,357 insertions, and 44 deletions. The low-level ESP32 dump is isolated in `device_memory_dump_esp32.cc`; `os_esp32.cc` gains only three allocation-accounting lines. The changes are split into snapshot-metadata and ESP32-runtime commits.

RISC-V allocation tagging depends on #3183, which is intentionally a separate PR and can be merged independently.

## Validation

- host VM/CLI compilation reached the package-sync stage successfully (package sync itself is unavailable in the sandbox cache)
- frame-debug snapshot decode test passed
- frame-debug `toitp` presentation test passed
- the split ESP32 dump, primitive, and `os_esp32.cc` translation units cross-compiled successfully
- the renamed VM-state checkpoint implementation and all changed consumers compiled successfully
- `git diff --check` passed


src/compiler/source_mapper.cc
```diff
@@ -327,6 +369,96 @@ SourceMapper::MethodEntry SourceMapper::build_method_entry(ir::Node* node,
...
+      auto location = manager_->compute_location(range.from());
+      position = {
+        .line = location.line_number,
+        .column = location.offset_in_line + 1,
+      };
+    }
+    result.parameters.push_back({
+      .index = index,
+      .name = parameter->name().c_str(),
+      .kind = index == 0 && implicit_parameter_kind != 0
```
floitsch: Why check for implicit_parameter_kind != 0. If it is, then we still return 0.

src/compiler/source_mapper.cc
```diff
@@ -327,6 +369,96 @@ SourceMapper::MethodEntry SourceMapper::build_method_entry(ir::Node* node,
...
+          ? implicit_parameter_kind
+          : 0,
+      .position = position,
+    });
+    seen[index] = true;
+  }
+  for (int index = 0; index < arity; index++) {
+    if (seen[index]) continue;
+    int kind = index == 0 ? implicit_parameter_kind : 3;
```
floitsch: Can we avoid the magic constanst?
Also: a switch seems nicer (but could be convinced to keep the nested conditionals.

src/resources/ble_esp32.cc
```diff
@@ -3504,6 +3549,7 @@ PRIMITIVE(get_bonded_peers) {
 }
 
 PRIMITIVE(set_gap_device_name) {
+  HeapTagScope heap_tag_scope(ITERATE_CUSTOM_TAGS + BLE_MALLOC_TAG);
```
floitsch: Don't we need to do the same for i2c, spi, rmt, ...? (or do we already?).

src/device_memory_dump_esp32.cc
```diff
@@ -0,0 +1,706 @@
...
+
+#include "device_memory_dump_esp32.h"
+#include "utils.h"
+
+namespace toit {
+
+void panic_put_char(char c);
+
+// These are used by the IDF panic path, but are not exposed by a public
+// header. They access the watchdog hardware directly and do not allocate or
```
floitsch: do they access the watchdog directly or the uart?
Can we avoid using the watchdog function and replicate what it does instead?

src/vm_state_checkpoint.h
```diff
@@ -0,0 +1,49 @@
...
+  VM_STATE_SCAVENGE_AFTER_ROOTS = 4,
+  VM_STATE_SCAVENGE_COMPLETE = 5,
+  VM_STATE_MARK_AFTER_ROOTS = 6,
+  VM_STATE_MARK_COMPLETE = 7,
+  VM_STATE_SWEEP_STARTED = 8,
+  VM_STATE_COMPACTION_STARTED = 9,
+  VM_STATE_GC_COMPLETE = 10,
+};
+
+#ifdef TOIT_VM_STATE_CHECKPOINTS
```
floitsch: We should also have an sdkconfig flag to even bundle the whole dumping code. (Not really useful for a device that is in the wild without serial port).

PR #3185, opened 2026-08-26
Add terminal control primitives

floitsch:
## Summary

- add host terminal primitives for TTY detection, raw-mode entry/restoration, and terminal dimensions
- add resize notifications using SIGWINCH and a self-pipe on POSIX, with console polling on Windows
- add a pseudo-terminal integration test covering raw input, full mode restoration, dimensions, and resize events

## Testing

- `ctest --test-dir build/host -R "^tests/terminal/terminal-test$" --output-on-failure`
- terminal and stdio tests repeated 10 times
- `toit analyze -Werror tests/terminal/terminal-test.toit`
- Win64 cross-compilation of the terminal resources, event source, VM registration, and propagated types


src/event_sources/terminal.cc
```diff
@@ -0,0 +1,226 @@
+// Copyright (C) 2026 Toit contributors.
+// Use of this source code is governed by an MIT-style license that can be
+// found in the lib/LICENSE file.
+
+#include "../top.h"
+
+#if !defined(TOIT_FREERTOS)
```
floitsch: I would prefer a different `#if`. What if we port to a zephyr target?

src/event_sources/terminal.cc
```diff
@@ -0,0 +1,226 @@
...
+  while (!stopping_) {
+    while (!stopping_ && resources().is_empty()) {
+      OS::wait(changed_);
+    }
+    if (stopping_) return;
+    // Windows does not deliver console window-size changes as a process
+    // signal. Keep the compatibility implementation inside the native event
+    // source so Toit applications still receive event-driven notifications.
+    OS::wait_us(changed_, 50 * 1000);
+    if (!stopping_) dispatch_changes_(locker);
```
floitsch: that looks like we are doing busy-polling. Is there really no better way to do this on Windows?

src/event_sources/terminal.cc
```diff
@@ -0,0 +1,226 @@
...
+  }
+#else
+  while (true) {
+    uint8 markers[64];
+    ssize_t result;
+    do {
+      result = read(wake_pipe_[0], markers, sizeof(markers));
```
floitsch: Instead of having a separate process/thread isn't there a way to use the existing event-sources (epoll)?

src/event_sources/terminal.h
```diff
@@ -0,0 +1,82 @@
...
+
+  // Called with the TerminalResizeEventSource lock held.
+  bool refresh();
+
+ private:
+  int fd_;
+  TerminalDimensions dimensions_;
+};
+
+class TerminalResizeEventSource : public LazyEventSource, public Thread {
```
floitsch: does this class (and others in this file) need to be in the header file?

src/resources/terminal.cc
```diff
@@ -0,0 +1,267 @@
...
+
+  proxy->set_external_address(group);
+  return proxy;
+}
+
+PRIMITIVE(is_terminal) {
+  ARGS(int, fd);
+
+#ifdef TOIT_WINDOWS
+  return _isatty(fd) ? process->true_object() : process->false_object();
```
floitsch: return BOOL(...)

src/resources/terminal.cc
```diff
@@ -0,0 +1,267 @@
...
+#else
+  return isatty(fd) ? process->true_object() : process->false_object();
+#endif
+}
+
+PRIMITIVE(enter_raw) {
+  ARGS(SimpleResourceGroup, group, int, fd);
+
+  ByteArray* proxy = process->object_heap()->allocate_proxy();
```
floitsch: How do we ensure that there aren't two processes (in parallel) that claim the console? Should we have a resource-pool? Or do we protect the creation/taking of the fd?

src/resources/terminal.cc
```diff
@@ -0,0 +1,267 @@
...
+  if (resource->resource_group() != group) FAIL(WRONG_OBJECT_TYPE);
+
+  group->unregister_resource(resource);
+  resource_proxy->clear_external_address();
+  return process->null_object();
+}
+
+}  // namespace toit
+
+#else
```
floitsch: add comments for which '#ifdef' this '#else' applies (same for the #ifdef).

src/resources/terminal.cc
```diff
@@ -0,0 +1,267 @@
...
+
+#include "../primitive.h"
+
+namespace toit {
+
+MODULE_IMPLEMENTATION(terminal, MODULE_TERMINAL)
+
+PRIMITIVE(init) { FAIL(UNSUPPORTED); }
```
floitsch: Is this necessary? I think to remember that we just see that no corresponding primitive is linked in and automatically fail with unsupported.

tests/terminal/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+# Copyright (C) 2026 Toit contributors.
+# Use of this source code is governed by an MIT-style license that can be
+# found in the lib/LICENSE file.
+
+if(UNIX)
```
floitsch: Do we have a good test for Windows?

We work primarily on Linux, so having good tests for Windows is even more important.

PR #3186, opened 2026-08-26
compiler: add pretty-printer safety foundations

floitsch:
Part 1 of 10 in the AST-driven Toit pretty-printer stack.

## Summary

- specify the formatter contract and implementation roadmap
- add structural AST-equivalence and comment-preservation verification
- add atomic source-file replacement
- model source comments, frozen trailing-comment lines, and multiline comments

This layer establishes safety and source facts without exposing the formatter command.

## Verification

- AST-equivalence, atomic-writer, and source-fact CTests pass
- the complete stack formats and verifies all 269 Toit files under lib, tools, and examples

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a></sub>

### DIFF
--- a/src/compiler/ast_equivalence.cc
+++ b/src/compiler/ast_equivalence.cc
@@ -1,0 +1,453 @@
+// Copyright (C) 2026 Toit contributors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; version
+// 2.1 only.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// The license can be found in the file `LICENSE` in the top level
+// directory of this repository.
+
+#include <cstring>
+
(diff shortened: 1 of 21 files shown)


src/compiler/ast_equivalence.cc
```diff
@@ -0,0 +1,453 @@
+// Copyright (C) 2026 Toit contributors.
```
floitsch: I think ast_equivalence is ok for debugging and bring-up, but long-term we should remove it.

src/compiler/format_source.cc
```diff
@@ -0,0 +1,176 @@
...
+namespace compiler {
+
+FormatSource::FormatSource(Source* source, List<Scanner::Comment> comments)
+    : source_(source) {
+  ASSERT(source_ != null);
+  build_lines();
+  build_comments(comments);
+}
+
+void FormatSource::build_lines() {
```
floitsch: Do we need lines and line numbers? What for?

src/compiler/format_source.h
```diff
@@ -0,0 +1,84 @@
...
+#include <string>
+#include <vector>
+
+#include "scanner.h"
+#include "sources.h"
+
+namespace toit {
+namespace compiler {
+
+struct FormatLine {
```
floitsch: Can (most of) the structs and classes move to the format_source.cc?

PR #3200, opened 2026-09-01
lsp: pull in latest pkg-host

girtsf:
This fixes `ALREADY_CLOSED` LSP log message spam
(https://github.com/toitlang/pkg-host/pull/100).


girtsf:
The `package.lock` content changes look a bit funny, but perhaps that's due to format changing between jag versions?

floitsch (approved):
LGTM!

floitsch:
The package.lock content changed because we are now using the Toit based package manager which has a different way of writing the lockfile. (still pretty much the same content, though).

PR #3202, opened 2026-09-02
LSP utilization improvements

girtsf:
I noticed that the LSP churns through many processes during normal editor use. I did some digging with help of robots, and came up with a few suggested improvements:
1) kill compiler processes when LSP requests are canceled
2) debounce `didChange` notifications so that we don't spawn a compiler process on every keystroke

Thoughts?

The diff looks bigger than it is, since large parts are just indentation changes to add `critical-do` guards.

### DIFF
--- a/tools/lsp/server/compiler-scheduler.toit
+++ b/tools/lsp/server/compiler-scheduler.toit
@@ -1,0 +1,151 @@
+// Copyright (C) 2026 Toit contributors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; version
+// 2.1 only.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// The license can be found in the file `LICENSE` in the top level
+// directory of this repository.
+
+import monitor
+
+import .verbose
+
+/**
+Decides when compiler actions run.
+
+All compiler work of the server goes through this class: $run executes an
+  action as soon as a compiler slot is free, and $schedule debounces and
+  coalesces actions that are triggered in bursts, like the analysis after
+  every keystroke. Heuristics about how often and how many compilers run
+  live here and not in the server.
+*/
+class CompilerScheduler:
+  slots_ /Slots_ ::= Slots_
+  /// Tasks that currently hold a slot. Nested $run calls reuse the slot.
+  slot-holders_ /List ::= []
+  /// Bursts of scheduled actions that are still being handled, keyed by id.
+  bursts_ /Map ::= {:}
+  /// Called when the last pending burst of scheduled actions has finished.
+  /// Runs under a cancelation guard and must not yield.
+  on-idle /Lambda? := null
+  /// How long a burst must be quiet before a scheduled action runs. A value <=
+  /// 0 disables debouncing.
+  debounce-ms /int := ?
+
+  constructor --max-concurrent/int --.debounce-ms/int:
+    slots_.limit = max-concurrent
+
+  /// Sets how many compiler actions may run at the same time. A value <= 0 means no limit.
+  max-concurrent= value/int -> none:
+    slots_.limit = value
+
+  /// Whether no scheduled action is pending or running.
+  is-idle -> bool:
+    return bursts_.is-empty
+
+  /**
+  Runs the compiler action $block as soon as a compiler slot is free.
+
+  The $id names the kind of action, for example "completion".
+  A nested call from within a running action reuses the slot of that action.
+
+  Returns the result of the $block.
+  */
+  run --id/string [block] -> any:
+    current := Task.current
+    if slot-holders_.contains current: return block.call
+    verbose: "Waiting for compiler slot: $id"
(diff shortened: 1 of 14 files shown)


tools/lsp/server/server.toit
```diff
@@ -114,34 +124,55 @@ class LspServer:
...
-        if id and (open-requests_.contains id):
-          open-requests_.remove id
-          connection_.reply id response
-        active-requests_--
-        if active-requests_ == 0 and not on-idle-callbacks_.is-empty:
-          callbacks := on-idle-callbacks_
-          on-idle-callbacks_ = []
-          callbacks.do: it.call
+      request := connection_.read
+      if request == null: return
```
floitsch:
```suggestion
      if not request: return
```

floitsch:
I'm in favor of a change like this.

I wonder if we want to debounce already the very first request.
I feel like getting an immediate response is pretty nice. After that we can rate-limit (by debouncing) the compilers.
We could also limit the number of compilers that are allowed to run in parallel.
Wdyt?

girtsf:
> I wonder if we want to debounce already the very first request. I feel like getting an immediate response is pretty nice. After that we can rate-limit (by debouncing) the compilers. We could also limit the number of compilers that are allowed to run in parallel. Wdyt?

Done! We now kick off analysis after the first request, then wait for the requests to stop before firing off another one.

Also added rate-limiting of compilers.

I'm not too fond of the timing-based tests, but I don't have great ideas on how to do it better. I'll share a separate draft PR with what the robots came up for it.

girtsf:
https://github.com/girtsf/toit/pull/1 - this is what robots came up with as an approach to synchronize with the mock compiler. (Pull request is on my fork, since stacked PRs only work if branches are in the same repo.)

If you think this approach has merit, I can resubmit as proper PR after the current one lands.

floitsch:
I haven't fully reviewed yet, but I wonder if it would be cleaner to have the CompilerLimiter be in charge of debouncing and running the actual compiler thing.

Basically, I would like the server to say:
```
   compiler-limiter --id="analysis" --arg=input-path:
      do compiler action (analysis, ...)
```

The limiter could then be smart about what compiler action should be run, how often, ...
All of the heuristics for when the compiler should run something would then be in a separate class/file.
The `id` and `arg` (feel free to rename to something more sensible) would just be to be able to coalesce multiple similar requests.

girtsf:
> I haven't fully reviewed yet, but I wonder if it would be cleaner to have the CompilerLimiter be in charge of debouncing and running the actual compiler thing.

Done!

tools/lsp/server/compiler-scheduler.toit
```diff
@@ -0,0 +1,147 @@
...
+          // while we were sleeping (or running). If nothing did, the burst is over.
+          if burst.args.is-empty: break
+          args := List.from burst.args
+          burst.args.clear
+          run --id=id: burst.action.call args
+      finally:
+        critical-do:
+          // Calls that arrive from here on start a new burst.
+          bursts_.remove id
+          if bursts_.is-empty and on-idle: on-idle.call
```
floitsch, line 122: That would put the `on-idle.call` in a `critical-do`. That feels wrong and/or dangerous.

tools/lsp/server/server.toit
```diff
@@ -264,12 +313,12 @@ class LspServer:
     //   taken into account.
     content-revision := next-analysis-revision_
     documents_.did-open --uri=uri document.text content-revision
-    analyze [uri]
+    scheduler_.run --id="analysis": analyze [uri]
```
floitsch, line 325: This would put different analysis requests (for different urls) into the same bucket. No?
Are we still analyzing them all if they arrive in bursts?
An analysis of one document doesn't guarantee that we will provide diagnostics for another.

We could, probably, keep track of which documents have been analyzed, but even that is difficult if they were analyzed from different "roots". (although we might already have enough knowledge for that).

floitsch (commented):
LGTM, but two comments. After that we should be good to go.

tools/lsp/server/compiler-scheduler.toit
girtsf: We could get rid of `critical-do` here, since nothing can cancel the burst task. But we have the same in https://github.com/girtsf/toit/blob/28b7967a548b25a0bdd5b40302232a95f5426c31/tools/lsp/server/server.toit#L172-L174 - there `work-finished_` -> `check-idle_` which calls the callbacks, still under the `critical-do` cancellation guard. And there the tasks can get cancelled.

I changed `server.toit` to use a separate task that runs callbacks when server is idle, and gets woken up by monitor signal to check. This still uses `critical-do` here, but the only user is `server.toit` and I added a comment that `on-idle` must not yield.

tools/lsp/server/server.toit
girtsf: `CompilerScheduler` has `run` and `schedule` methods. Only `schedule` method buckets things since `run` is used for a bunch of different things that can't easily be bucketed together by arg. For `run`, the `--id` is used only for log messages (e.g., `Waiting for compiler slot: analysis`). All `run` blocks run immediately, as long as there are free slots, otherwise, they wait for a slot.

floitsch (approved):
LGTM.
Thanks.

floitsch:
We could, in theory, also give priorities to some interactions (like completion, or goto-definition), and maybe even allow an extra compiler for those (to keep the latency low), but I would do that only as a follow-up PR (and only if we notice something).

PR #3208, opened 2026-09-08
Preserve Debian upgrade ordering with a version epoch

floitsch:
Changing prereleases from `-alpha.` to `~alpha.` makes the next Debian package sort below published versions such as `2.0.0-alpha.198-1`. Introduce epoch `1` so existing installations can upgrade while final releases still sort after prereleases. The epoch must remain in future Debian versions.

Restore historical changelog entries to their published spelling and add a package-version comparison against alpha.198 to the release workflow.

Validation: executed the workflow's version-generation commands, parsed generated changelogs with `dpkg-parsechangelog`, and verified ordering from the published alpha.198 through epoch-1 alpha.199, alpha.200, final 2.0.0, and 2.0.1. Shell syntax and whitespace checks passed. A full Debian package installation/upgrade was not run.

### DIFF
--- a/.github/workflows/tar-ball.yml
+++ b/.github/workflows/tar-ball.yml
@@ -181,7 +181,9 @@ jobs:
           # Remove the leading 'v' from the version if it exists.
           version_without_v=${version#v}
           # Debian uses '~' to sort prereleases before the final release.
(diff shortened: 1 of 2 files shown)


floitsch:
Not actually necessary.
Users will see a downgrade warning, but that's acceptable.

PR #3219, opened 2026-09-09
tests: drive the mock compiler over a control socket

girtsf:
The mock compiler used to fetch its canned answer through the LSP file server, by having the test open a "mock:" document for each command. That made every answer update trigger a phantom analysis run, and gave tests no way to tell when a compiler was running, so they guessed with delays and retries.

Now each mock compiler run connects back to the test over a localhost port, announces its request, and blocks until the test sends the output it should print. Holding the answer back is the rendezvous: tests can see which compilers are running and decide when they may finish, without relying on timing.

run-client-test --use-mock owns the MockCompiler: it publishes the port in the server's environment before spawning, hands the mock to the test block next to the client, and closes it afterwards.

The concurrency-limit, cancel and slow-compiler tests use the rendezvous instead of delays. The concurrency-limit test now also covers a limit of 2. Crash and timeout directives are behind --crash and --timeout flags, and the unused DUMP_FILE_NAMES command and SLOW directive are gone.


floitsch (approved):
LGTM.

floitsch:
Looks like the windows buildbot isn't happy about the changes. Could you have a look?

girtsf:
> Looks like the windows buildbot isn't happy about the changes. Could you have a look?

The robots think that this might be due to TCP socket eating buffered data on close on Windows, vs on POSIX it gets sent out even after the socket gets closed. I'll try to set up a Windows VM and replicate the issue, but in the meanwhile, the second commit might fix the issue if the theory is correct.

floitsch:
> > Looks like the windows buildbot isn't happy about the changes. Could you have a look?
> 
> The robots think that this might be due to TCP socket eating buffered data on close on Windows, vs on POSIX it gets sent out even after the socket gets closed. I'll try to set up a Windows VM and replicate the issue, but in the meanwhile, the second commit might fix the issue if the theory is correct.

Instead of setting up a VM you can also use the GitHub builders themselves. Let the AI shrink down the worflow to that particular use-case and let it push/run on the GitHub builders. Involves a lot of latency, but the AI doesn't care.

girtsf:
@floitsch the yak stack gets deeper. Looking into Windows failures uncovered some TCP socket handling peculiarities on Windows. Looking into those uncovered related issues with pipes (and probably UART).

Fix here: #3221 After that lands, this should be rebased and retested on Windows.

PR #3221, opened 2026-09-13
Fix abortive close and use-after-free on Windows resource close

girtsf:
Closing Windows TCP sockets with an overlapped receive armed can reset the peer and discard unread data. Closing TCP, UDP, pipe, and UART handles also allows pending completions to write into freed `OVERLAPPED` structures.

Add `WindowsOverlapped` to track successfully issued operations and cancel/reap them before destroying their handles, events, and resources. Synchronous failures are not waited on, because no operation was started.

TCP writes now use nonblocking `send` and `FD_WRITE` readiness. Each primitive call retries `send` directly; no cached write-readiness flag can overwrite an `FD_WRITE` notification that arrives as a blocked send returns. They report only bytes accepted by the transport, handle partial writes and backpressure, and leave no queued `WSASend` for close to cancel. This also corrects the address length passed to `connect` and closes the auxiliary socket event. Waiting for a pending send inside the shared event thread would stall unrelated I/O; nonblocking sends avoid that dependency. See [Microsoft's send semantics](https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsasend).

Add `pipe.write-result` to return the actual completed byte count, null while pending, or an asynchronous error. The existing write primitive retains its queued-count behavior for package compatibility. [pkg-host#101](https://github.com/toitlang/pkg-host/pull/101) uses the new primitive to suspend the writing task until completion, serialize concurrent writers, and propagate cancellation/errors. **Preventing pipe write-then-close data loss requires that companion package update.** It remains draft until an SDK release includes the new primitive and its minimum SDK requirement can be updated.

UART writes and older host packages still report queued writes; closing them may cancel pending data. The gzip test retains file input while the SDK tests use the released host package.

Validation:
- Cross-compiled the Windows runtime, compiler, CLI, and executable templates with MinGW.
- TCP write/close and four successive 16 MiB backpressure rounds on the same socket pass on Linux and Wine.
- A focused Winsock harness forces `FD_WRITE` handling before a would-block send returns: the original implementation fails to retry on a writable socket; the updated implementation accepts the next byte.
- Concurrent socket read/write and TCP cancellation/close regressions pass under Wine.
- Pipe close with a pending write and allocation reuse passes under Wine; the test now closes from a separate task so it also supports completion-aware writers.
- All four companion package scenarios pass under Wine: byte-array delivery, chunked io.Data delivery, broken-pipe errors, and concurrent close. With the original package, the delivery regression receives 0 of 262,144 bytes.
- POSIX pipe primitive passes a syntax check. Native Windows validation is delegated to CI; Wine is not a substitute for it.


girtsf:
**FULL DISCLOSURE**: this was written by LLMs, though with significant back and forth from me. I'm not a Windows dev, so I'm not super qualified to review the changes. It did pass the tests on a Windows VM. No UART to test on the VM though.

As the changed zlib-gzip-test shows, this can prevent the last written data from being delivered across the pipe if it gets closed. (Though arguably still better than a use-after-free ;)) If we want pipes on windows to act the same as they do on Linux, we would need some special handling for it. Let me know if you think this is important, and I can follow up with some options.

floitsch:
@coderabbit review

floitsch:
@coderabbit review

girtsf:
image

PR #3222, opened 2026-09-13
Don't redefine mkdtemp on mingw-w64 14+

girtsf:
Without this, newer compilers get compilation errors:

```
  C:/toit/tests/ctest/ar-read-test.cc:19:14: error: 'char* mkdtemp(char*)' was declared 'extern' and later
  'static' [-fpermissive]
```


floitsch (approved):
LGTM. Thanks.

PR #3223, opened 2026-09-13
Unify async I2C transfers and SPI controller lifecycle

floitsch:
Consolidate asynchronous I2C controller read, write, and write-read operations into one native transfer start/finish pair. Preserve address-only probing and explicit abort; reject empty public transfers consistently. An omitted clock-stretch timeout uses the platform default (100 ms on ESP32). EC618 can reject unsupported custom hardware timeouts while still supporting task deadlines.

Serialize SPI transfers and reservations in the common library so each backend has one active controller operation. Reservations belong to the calling task and device; conflicting operations from that task fail instead of deadlocking. Cancellation either aborts natively or drains under protection before releasing ownership, preserving ESP32's drain-before-release behavior.

Add regression coverage for cancellation, scheduler progress, reservation contention, immediate reuse, repeated I2C NACK recovery, and SPI target setup in clock-high modes. Based on master `71d24d83`, including the ESP-IDF fixes merged through #3224; this PR does not change the dependency pin.

This is the common prerequisite for the EC618 adaptation in #3225, which targets the platform branch in #3199.

Validation:
- ESP32 and ESP32-S3 rig preflight, flashing, and wiring health: 8/8 passed.
- Focused I2C/SPI hardware suites: 13/13 passed on each variant, including 100 consecutive missing-device write/read NACK pairs.
- Full ESP32/S3 hardware matrix: 120/121 passed across the final run and reruns after resolving host serial-port contention and temporary-file quota issues. The remaining classic ESP-NOW packet-loss failure also reproduces on the same boards after flashing unchanged master (`0ef62329`) and using its SDK. All I2C/SPI cases passed on both variants.
- Host suite: 1287/1288 passed in the host environment. The multicast test also failed with the pre-existing SDK and passed in an isolated network namespace with its expected multicast loopback route.
- Fuzzer library: 353/353 passed.
- Rebase changes no refactor source content; `git diff --check` passes.


lib/i2c.toit
```diff
@@ -719,6 +726,8 @@ class Device implements serial.Device:
   /**
   Writes the $bytes to the device.
 
+  Throws `INVALID_ARGUMENT` if $bytes is empty.
```
floitsch: We typically don't state what we throw, unless it's an "exception" (which the program should act upon").
Here "The $bytes data must not be empty." (or similar) is more typical of how we would write this.

floitsch:
@coderabbit review

PR #3227, opened 2026-09-13
Fix LSP rename range for nullable type annotations

girtsf:
Previously, if `Foo` was being renamed to `Bar`, then `f/Foo?` would end up as `f/FBar` instead of `f/Bar?`.


floitsch (approved):
LGTM!
thanks.

PR #3230, opened 2026-09-14
Find prefixed class instantiations in LSP rename

girtsf:
`prefix.Class args` call sites were never found: the IR call range only covers the prefix, and the resolver only mapped calls with named arguments into ir_to_ast. Map prefixed calls too, and match the no-argument `prefix.Class` form by the end of its range.


floitsch (approved):
LGTM.
It's a bit of a hack, but fixing it correctly would require a big rewrite...

PR #3231, opened 2026-09-14
Use MAP_PRIVATE for anonymous host flash registry.

girtsf:
Only the file-backed registry needs MAP_SHARED. Private anonymous reads map the zero page, so the boot scan of the 64 MB region no longer allocates ~16k shmem pages. VM start on a snapshot drops from ~27 ms to ~11 ms.


floitsch:
@coderabbit review

floitsch (approved):
LGTM.
Thanks.

PR #3234, opened 2026-09-14
Use a line table to compute source locations

girtsf:
`SourceManager::compute_location` rescanned file text from the start (or a forward-only cache) on every call. Code generation calls it once per expression, making it ~15% of compile time.

Now each source builds a line-start table on first use and locations are a binary search. `source_for_position` uses `std::upper_bound` instead of a hand-written search.

Compile time, median wall:
- hello world: 112 -> 95 ms
- tests/string-test.toit: 167 -> 112 ms

Also fixes a CRLF miscount in the old cache: after a query on the '\n' of a "\r\n", the next query later in the same file was one line too high. A position on that '\n' now reports the line it ends (previously next line, column -1).

Adds tests/ctest/sources-test.cc (LF, CRLF, lone CR, empty file, multiple sources, backwards queries).


floitsch (approved):
LGTM.

floitsch:
Thanks.

PR #3237, opened 2026-09-17
Avoid copying synthetic TLS handshake records

girtsf:
- Handshake messages are now built as external byte arrays, so `tls.set_incoming` takes the buffer instead of copying it. Saves 4 KB of peak memory during a TLS handshake.
- Fix `tls.set_incoming` for a non-zero `from`, which added the offset twice and freed an invalid pointer (no current caller passes one).

Both changes are covered by `tests/tls-set-incoming-test.toit`.


floitsch (approved):
LGTM!
thanks.

PR #3240, opened 2026-09-20
Add ESP32-H2 support

floitsch:
Add ESP32-H2 as a firmware target (`make esp32h2`) and test it with an ESP32 companion using the existing breadboard connections.

- Add toolchain configuration, envelope/CI builds, chip identification, GPIO/ADC support, image recognition and RTC handling. H2 exposes 3840 bytes of RTC user memory. Enable BLE without Wi-Fi, Ethernet or ESP-NOW.
- Read the six-byte base MAC through `esp_read_mac`, avoiding the eight-byte eFuse MAC overflowing a six-byte buffer on IEEE 802.15.4 targets.
- Support serial test setup without Wi-Fi credentials and a paired Toit wiring precheck. The tester initiates cases, checks measurements/data, enforces deadlines, and accepts completion. Both boards delay the ordinary completion marker; the host runner needs no special mode.
- Put selected reusable tests in `tests/hw/paired`, with explicit roles, local pins and SPI target DMA capabilities. H2 entrypoints provide fixture configuration. Keep independent copies of the BLE/I2S utilities in the paired suite; leave the original ESP32 tests untouched for a follow-up PR.
- Add PWM waveform/endpoints/lifecycle, UART continuity with deliberate-pause controls, duplex/framing and actual UART pixel waveforms, GPIO open-drain release/ownership/edge waits, pulse filtering, SPI boundaries/cancellation/reuse, I2C read-buffer guards, long RMT output, BLE descriptors/payloads, and portable formatting/multipage storage tests.
- Fix two PWM defects exposed by the tests: frequency changes compared hertz against the duty-counter maximum, and 100% duty left a recurring low notch. Validate against the configured frequency limit and use idle-high for full duty, preserving readback and resumption of PWM.
- Requires the merged repeated-start fix in https://github.com/toitware/esp-idf/pull/134. The ESP-IDF pin update is separate in #3245; merge that first. This PR does not change the submodule pin.

Validation on ESP32-H2-DevKitM-1 and ESP32-DevKitC-V4:

- ESP32-H2 and ESP32 firmware/envelope builds pass; Toit analysis passes with warnings treated as errors, including the original ESP32 BLE/I2S tests. Earlier C6 build and firmware image/map checks passed.
- All 21 default focused regression cases pass. The same PWM tests also pass with ESP32 as testee and H2 as observer.
- All 79 bus cases pass, plus the 38-case repeated-start boundary regression. Wiring, identity, basic GPIO/ADC, external deep-sleep wakeup, runtime/storage/formatting and both BLE roles pass.
- UART tests verify 32769 bytes in each direction through 921600 baud, 32 KiB TX continuity through 2.5 Mbaud, and LED-style consecutive writes. Deliberate pauses are detected. Seven-pixel UART color/pulse captures pass.
- Four Philips16 I2S direction/clock-role combinations pass with over 200 KB verified per case. RX/master recorded 22 mismatches under the existing allowance of 30; the other three recorded zero. MSB32 RX/slave recorded zero; MSB32 TX/master recorded 30 and is marginal. Writers stop before completion.
- Corrupt data, silence, cleanup without completion, skipped work and testee crash all fail without either completion marker.

Known limitations exposed by the added tests:

- The RMT backend of pixel-strip 1.4.0 requests 20 MHz, which H2's divider realizes as 16 MHz. Its waveform test fails; the default regression selection uses the passing UART backend. The RMT reproducer remains available.
- MSB32 I2S receive/master exceeds the existing mismatch allowance. PCM16 transmit in either clock role fails to synchronize on this mixed-chip fixture. These cases remain explicit failing reproducers; no allowance was increased.
- Unpaced 2.5 Mbaud duplex dropped bytes. Passing TX continuity tests does not establish lossless duplex at that rate. Gap measurements have documented detection thresholds.

Setup, selected cases and failure reproducers are documented in `tests/hw/esp32h2/README.md` and `tests/hw/paired/README.md`. Changes to the existing ESP32/S3 suites are deferred to a follow-up PR.


lib/esp32/esp32.toit
```diff
@@ -172,6 +173,7 @@ The following GPIO pins can be used:
 - ESP32: 0, 2, 4, 12-15, 25-27, 32-39
 - ESP32-S2: 0-21
 - ESP32-S3: 0-21
+- ESP32-H2: 7-14, subject to package and board availability
```
floitsch: Here and in the other places: "subject to package and board availability" is obvious and doesn't need to be mentioned.

tests/hw/esp32h2/wiring-fixture/main/fixture.c
```diff
@@ -0,0 +1,121 @@
...
+#include "driver/uart.h"
+#include "esp_chip_info.h"
+#include "esp_mac.h"
+#include "esp_timer.h"
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+
+#if CONFIG_IDF_TARGET_ESP32H2
+static const int pins[] = {0, 1, 2, 3, 4, 5, 10};
+#elif CONFIG_IDF_TARGET_ESP32
```
floitsch: Why do we need this #elif?

tests/hw/esp32h2/check-wiring.py
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env python3
```
floitsch: We do not want to have any Python code in our repository.
You might as well replace with a Toit program now. Don't forget to update the README.

tests/hw/esp32h2/i2c-repeated-start.toit
```diff
@@ -0,0 +1,25 @@
...
+      --rx=(h2 ? H2-RX : HELPER-RX)
+      --tx=(h2 ? H2-TX : HELPER-TX)
+      --baud-rate=115200
+  if h2: sleep --ms=1500
+  try:
+    [50_000, 100_000].do: | frequency |
+      // Exercise both sides of the controller's 32-byte FIFO boundaries.
+      [17, 19, 30, 31, 32, 33, 62, 63, 64, 65, 94, 95, 96, 127, 128, 129, 255, 256, 1024].do: | size |
+        print "Repeated-start write size $size frequency=$frequency"
+        buses.i2c-case port h2 frequency size
```
floitsch: Who is checking that things work?

tests/hw/esp32h2/README.md
```diff
@@ -50,50 +55,78 @@ TOIT="$PWD/build/host/sdk/bin/toit"
...
 ```
 
-`basics.toit` checks chip identity and the H2 RTC buffer size. `runtime.toit`
-reuses the EC618 suite's platform-independent GC and RAM/flash storage checks.
-`tests/hw/esp32/run-time-test.toit` checks runtime accounting and RAM storage
-retention across timer deep sleep with the same command.
-
-For paired tests, install the tester on the helper too, using its port and
-`build/esp32/firmware.envelope`. Then run:
+All H2 suites now run on both boards. The ESP32 is the trusted **tester**;
```
floitsch: Avoid historical comments like "now".
This is our first commit of the h2.

PR #3241, opened 2026-09-20
Add RP2350 support with USB OTA and hardware peripherals

floitsch:
Adds a Toit port for the WeAct RP2350B: the VM and GC, persistent containers and storage, wired peripherals, and USB firmware updates with validation and rollback.

## Runtime and peripherals

- Integrate the pinned Pico SDK fork and upstream FreeRTOS kernel. The compiler is not forked or pinned.
- Use one event task for GPIO, UART, I2C, SPI, and USB input. Peripheral APIs take integer GP numbers.
- Support GPIO/interrupts, ADC, PWM, UART with software RS485, and I2C/SPI controller and target operation. ADC shuts down after its last resource closes and reinitializes on reopen.
- Support timed deep sleep, native-failure recovery, and a shared `system.watchdog` API on ESP32, EC618, and RP2350. Existing platform watchdog APIs remain available; the watchdog package update is [toitware/toit-watchdog#17](https://github.com/toitware/toit-watchdog/pull/17).
- Register the TLS event source and support TLS session export through a small C adapter for mbedTLS's private C headers.
- Remove the obsolete `SweepingVisitor` declaration left behind by the fast-sweep implementation; it caused the ARM linker to reference the removed `add_free_list_region` method.

## Flashing and updates

`toit tool firmware -e firmware.envelope flash --port SERIAL_PORT` updates running firmware. `flash --bootloader [--serial BOARD_SERIAL]` performs initial installation or recovery through the USB ROM bootloader. Recovery images are generated from the current envelope, including installed containers, assets, and configuration.

The normal SDK bundles picotool, the native OTA uploader, and replaceable libusb with its source and notices. Flashing requires neither Python nor a mounted drive. The Debian package installs the PICOBOOT USB access rule automatically; archive installs document the one-time setup and the alternative of copying an extracted recovery UF2 onto the mounted ROM drive. Trial updates preserve the previous firmware until the application validates startup; recovery preserves slot B and the registry, with normal ROM version selection still applying.

Documentation covers supported usage, architecture, and reproducible tests. Bring-up transcripts and investigation diaries are not tracked. New first-party build helpers and hardware-test runners use Toit instead of Python.

Changing `RP2350_PROGRAM` now rebuilds the generated application snapshot, including when the selected source is older than the previous build.

## Validation

- Host image parser/boundary/fuzz, hash, picotool, envelope install/extract, retained-memory layout, and uploader PTY tests.
- CLI regressions for envelope contents/configuration, ROM selectors, failed load/reboot handling, option validation, and installed SDK discovery for both flashing modes.
- Linux SDK build/install/relocation and standalone flasher archive; MinGW build and runtime dependency inspection. Native watchdog implementations compile for ESP32, EC618, and RP2350; the watchdog package's six existing tests pass.
- Physical mapped-wire tests, GPIO/ADC/PWM/UART, I2C and SPI controller/target operation, cancellation, process-exit cleanup, and simultaneous peripheral traffic with forced GCs.
- Physical OTA validation/rollback, equal-version updates and downgrades, malformed/interrupted uploads, persistent container installation/uninstallation, native-failure recovery, and repeated deep-sleep wakeups.
- Physical TLS session export and encrypted echo assertions over UART through the ESP32 relay, plus existing host TLS regressions.
- Physical RS485 direction/drain, ADC last-close/reopen, shared watchdog rearm/feed/stop, and all four cases in the converted OTA rejection test.
- Installed SDK CLI flashed a production envelope containing a validating application, assets, and configuration through both the bundled serial uploader and PICOBOOT tool, with no tool-path overrides. ROM flashing verified the written image and rebooted successfully. The board is left on confirmed v66, partition 0, with the serial port released.
- Debian udev packaging checked in an isolated staging directory with debhelper 13.11.4; the installed rule matches the bundled upstream rule.
- Application-switch regression verifies that selecting another program changes the built snapshot.

## Limits

Physical coverage is for the WeAct RP2350B in Arm Secure mode. RP2040, RISC-V, external PSRAM, networking, and GPIO deep-sleep wakeup are not implemented. I2C scanning/address-only probes remain TODO; a GPIO implementation must reuse the bus's reserved pins and restore their peripheral function. SPI controllers support all modes; hardware targets support modes 1/3 and reject modes 0/2.

The classic ESP32 Jaguar I2C target can intermittently NACK long writes at 400 kHz. The merged ESP-IDF fixes are integrated separately by [#3245](https://github.com/toitlang/toit/pull/3245); this PR does not change the ESP-IDF submodule.

macOS/MSVC builds and Windows/macOS physical flashing remain for CI or platform testing.


docs/rp2350-envelope-bundle.md
```diff
@@ -0,0 +1,67 @@
...
+macOS have not been tested with physical RP2350 hardware; the macOS build
+remains unverified.
+
+`recovery.uf2` is for initial provisioning and recovery in BOOTSEL mode. Use
+the separately distributed picotool bundle for direct USB loading; the
+recovery image installs its own partition table, so the command must use
+absolute addressing:
+
+```sh
+picotool load -v --ignore-partitions recovery.uf2 --ser BOARD_SERIAL
```
floitsch: All of this sounds like upbringing.
1. if we need a recovery.uf2, then it should be in our envelope.
2. if we need a tool to something then we should bundle it, and use it from `toit firmware flash`.
I assume that `firmware flash` works fine with the pico, but if not, discuss with me how best we can do that.

docs/rp2350-ota.md
```diff
@@ -0,0 +1,531 @@
...
+
+An update replaces the complete firmware image: the native VM, system program,
+and bundled containers. Images can come from the bring-up build selected with
+`TOIT_RP2350_PROGRAM` or from a firmware envelope. Separately installed
+containers and storage live in the persistent registry and survive firmware
+updates. There is no network stack or network update transport on this port
+yet; the current uploader uses USB.
+
+The RP2350 TLS port also leaves `tls.get_internals` unimplemented because the
+Pico SDK's private mbedTLS headers are not C++ compatible. That primitive is
```
floitsch: Can/should we change that?

docs/rp2350-ota.md
```diff
@@ -0,0 +1,531 @@
...
+1. OTA supports upgrades, downgrades, and equal-version replacements. The
+version affects normal ROM selection after validation and recovery, as
+described below.
+
+```sh
+make rp2350 \
+  RP2350_VERSION=2 \
+  RP2350_PROGRAM="$PWD/tests/hw/rp2350/vm-smoke.toit"
+
+make rp2350-ota-upload
```
floitsch: why do we have this?
shouldn't `toit firmware...` just do that for us?

docs/rp2350-ota.md
```diff
@@ -0,0 +1,531 @@
...
+  distinguishes the explicit error reboot from a watchdog-only rollback.
+- [Application-controlled validation](../tests/hw/rp2350/results/2026-09-19-envelope-health-validation-upload.log):
+  the production envelope uploaded and validated successfully with validation
+  performed by its critical application after all startup checks. The system
+  boot program no longer automatically confirms production envelopes.
+
+Run the rejection suite against a confirmed board with Python and pyserial:
+
+```sh
+python tests/hw/rp2350/ota-negative-test.py \
```
floitsch: We absolutely don't want to have any Python in our repository.

If you think this test is useful, rewrite it in Toit.

docs/rp2350-rig-guide.md
```diff
@@ -0,0 +1,874 @@
...
+
+The test passed exact echoes of 1, 31, 256, and 4,096 bytes at an actual baud
+rate of 115,207, a 4,096-byte echo at 921,658 baud, and a 257-byte echo at
+9,600 baud. It also passed receive-buffer overflow accounting, retained-prefix
+behavior, post-overflow recovery, controller and pin reservation checks, and
+clean VM teardown. These test points cover actual baud rates from 9,600 through
+921,658. Hardware RTS/CTS pin-role validation passed with RTS GP19 and CTS
+GP18; that flow-control pair was not physically wired for this run.
+
+RS485 half-duplex mode, transmitted break, and 1.5 stop bits are unsupported.
```
floitsch: I believe (but I could be wrong) that the ec618 implements the R485 in software?

lib/rp2350/watchdog.toit
```diff
@@ -0,0 +1,71 @@
+// Copyright (C) 2026 Toit contributors.
```
floitsch: Could we share the watchdog lib and primitives with the ones from esp32 and ec618?

Also: does the watchdog package handle ec618 and pico?

lib/i2c.toit
```diff
@@ -544,6 +556,9 @@ class Bus:
 
   Waits at most $timeout-ms for a response on each address. If the bus is very
     slow, increase the timeout.
+
+  RP2350 cannot issue an address-only transfer, so this method throws
+    `UNIMPLEMENTED` there.
```
floitsch: that sucks.
we can't do a scan on the pico?
Can we change the SDK to make this work?

src/event_sources/event_rp2350.cc
```diff
@@ -0,0 +1,76 @@
+// Copyright (C) 2026 Toit contributors.
+// Use of this source code is governed by the LGPL-2.1 license in LICENSE.
+#include "../top.h"
+#ifdef TOIT_RP2350
+#include "event_rp2350.h"
+
+namespace toit {
+Rp2350EventDispatcher* Rp2350EventDispatcher::instance_ = null;
```
floitsch: How similar is this to the esp32 dispatcher.
Should we combine them?

src/event_sources/gpio_rp2350.cc
```diff
@@ -0,0 +1,111 @@
...
+#ifdef TOIT_RP2350
+
+#include "gpio_rp2350.h"
+
+#include "hardware/gpio.h"
+#include "hardware/irq.h"
+
+namespace toit {
+
+Rp2350GpioEventSource* Rp2350GpioEventSource::instance_ = null;
```
floitsch: How similar is this to the esp32 gpio event sources?
should we combine them?

src/event_sources/uart_rp2350.h
```diff
@@ -0,0 +1,86 @@
...
+
+// The event bits intentionally match lib/uart.toit's ResourceState bits.
+enum Rp2350UartState : uint32_t {
+  kRp2350UartReadState = 1 << 0,
+  kRp2350UartErrorState = 1 << 1,
+  kRp2350UartWriteState = 1 << 2,
+  kRp2350UartBreakState = 1 << 3,
+};
+
+class Rp2350UartEventResource : public Resource {
```
floitsch: For all the 'h' files: do we have to have the class in there?
Are they used outside the c++ files?

src/resources/adc_rp2350.cc
```diff
@@ -0,0 +1,152 @@
...
+
+  int pin() const { return pin_; }
+  int channel() const { return pin_ - static_cast<int>(ADC_BASE_PIN); }
+
+ private:
+  int pin_;
+};
+
+// Selects the resource's channel and discards conversions to settle the ADC's
+// sample-and-hold after a mux change. This matters for the rig's 10 kohm source
```
floitsch: Don't mention "the rig" in C++ code. Outside testing it shouldn't exist.

src/resources/adc_rp2350.cc
```diff
@@ -0,0 +1,152 @@
...
+
+  {
+    // RP2350 has one ADC shared by every input. Initialization, mux selection,
+    // and conversions are serialized on this lock.
+    Locker locker(OS::global_mutex());
+    if (!adc_initialized) {
+      adc_init();
+      adc_initialized = true;
+    }
+    adc_gpio_init(pin);
```
floitsch: Who is responsible for turning it off?

